PR #286, opened 2018-11-15
Fixes #177

### LINKED ISSUE #177: Console supports registering new empty service and delete empty service
It's possible to create an empty service before any of the instances is registered, so service properties can be saved without instances.

This feature is quite different from Eureka or Consul but useful since Nacos introduces service level properties which should not be eliminated even after all instances under serivce are deregistered.

loadchange:
FE fix issue #177


loadchange:
@nkorange Verify that the back-end code is incorporated into the develop branch

Issue #672, opened 2019-01-22
startup.cmd bug


ankeway:
There is  a  Bug in Nacos Server Folder ，Bin directory ，startup.cmd File,
if not ""%2"" == "cluster" (
)
if  you run startup.cmd -m cluster can’t startup with cluster mode，
if not "%2" == "cluster" (
)
run startup.cmd -m cluster  Ok，you can running in cluster mode

chinchillia:
CREATE TABLE users (
	username varchar(50) NOT NULL PRIMARY KEY,
	password varchar(500) NOT NULL,
	enabled boolean NOT NULL
);

CREATE TABLE roles (
	username varchar(50) NOT NULL,
	role varchar(50) NOT NULL
);

INSERT INTO users (username, password, enabled) VALUES ('nacos', '$2a$10$EuWPZHzz32dJN7jexM34MOeYirDdFAZm2kuWj7VEOJhhZkDrxfvUu', TRUE);

INSERT INTO roles (username, role) VALUES

PR #674, opened 2019-01-22
hotfix #672

ankeway:
There is a Bug in Nacos Server Folder ，Bin directory ，startup.cmd File,
if not ""%2"" == "cluster" (
)
if you run startup.cmd -m cluster can’t startup with cluster mode，
if not "%2" == "cluster" (
)
run startup.cmd -m cluster Ok，you can running in cluster mode


yanlinly:
Fix #673

PR #897, opened 2019-03-13
更新HomePage

fluffywar:
仓库里的显示的主页不对
https://mvnrepository.com/artifact/com.alibaba.nacos/nacos-client/0.9.1
改成了github的地址


nkorange:
@fluffywar You should always set your target branch to alibaba:develop.

PR #1014, opened 2019-04-07
#1013 fix

niget:
not add http prefix when it is already there


client/src/main/java/com/alibaba/nacos/client/config/http/ServerHttpAgent.java
```diff
@@ -66,7 +67,7 @@ public HttpResult httpGet(String path, List<String> headers, List<String> paramV
                               long readTimeoutMs) throws IOException {
         final long endTime = System.currentTimeMillis() + readTimeoutMs;
 
-        boolean isSSL = false;
+        boolean isSSL = Boolean.getBoolean("com.alibaba.nacos.client.naming.tls.enable");
```
nkorange: I think the property name should not carry a 'naming', since the current module is Config.

niget: copy from com.alibaba.nacos.client.config.http.ServerHttpAgent.httpGet:70

nkorange:
This feature is already supported in #1432

Issue #359, opened 2018-11-28
Support user extended health check type


nkorange:
Nacos currently supports TCP, HTTP and MYSQL health check types, respectively for these three type of services. We should let more types of service in Nacos such as REDIS, MONGODB, MQ etc. and provide a way for user to implement their health checkers.

XCXCXCXCX:
> Nacos currently supports TCP, HTTP and MYSQL health check types, respectively for these three type of services. We should let more types of service in Nacos such as REDIS, MONGODB, MQ etc. and provide a way for user to implement their health checkers.

hi, I think that in any case we need to remodel the nacos server (eg.SPI), it seems that this is not very elegant, do you think there is a better way to provide extensions?

PR #1126, opened 2019-04-27
[#359] provide a way for user to implement their health checkers.

XCXCXCXCX:

#359 
provide a way for user to implement their health checkers.

I found that the current health uses a lot of hard coding, I made some modifications to it, and allowed the user to construct their own health check object by inheriting AbstractHealthChecker (different ways need to use different json format to transfer HealthChecker), and by implementing HealthCheckProcessor To complete the processing logic of the heartbeat packet of the server.

Using the SPI method, the user can introduce the nacos-naming dependency and customize the implementation to create a jar package, which can be introduced when starting the project.

Users must pay attention to:

1. When implementing extensions, do not conflict with existing types (tcp/http/mysql/none), otherwise they will be overridden by the default implementation;

2. When inheriting the AbstractHealthChecker extension, you must implement the getType interface, otherwise the default type "unknown" will be used;

3. CustomHealthChecker and CustomHealthCheckProcessor must be implemented at the same time, and type must correspond;


naming/src/main/java/com/alibaba/nacos/naming/healthcheck/extend/HealthCheckExtendProvider.java
```diff
@@ -0,0 +1,92 @@
+package com.alibaba.nacos.naming.healthcheck.extend;
+
+import com.alibaba.nacos.api.naming.pojo.AbstractHealthChecker;
+/*
```
nkorange: licence info is alway on the top of the file.

XCXCXCXCX: OK

naming/src/main/java/com/alibaba/nacos/naming/healthcheck/extend/HealthCheckExtendProvider.java
```diff
@@ -0,0 +1,92 @@
...
+            healthCheckerType.add(checker.getType());
+            HealthCheckType.registerHealthChecker(checker.getType(), checker.getClass());
+        }
+        if(!processorType.equals(healthCheckerType)){
+            throw new RuntimeException("An unmatched processor and healthChecker are detected in the extension package.");
+        }
+    }
+
+    private String lowerFirstChar(String simpleName) {
+        if(simpleName == null || "".equals(simpleName)){
```
nkorange: StringUtils.isBlank() is more recommended.

naming/src/main/java/com/alibaba/nacos/naming/healthcheck/extend/HealthCheckExtendProvider.java
```diff
@@ -0,0 +1,92 @@
...
+            throw new IllegalArgumentException("can't find extend processor class name");
+        }
+        char[] chars = simpleName.toCharArray();
+        if(chars[0] >= LOWER_A && chars[0] <= LOWER_Z){
+            chars[0] = (char)(chars[0] + 32);
+        }
+        return String.valueOf(chars);
```
nkorange: Why not just use ```String.valueOf(simpleName.charAt(0)).toLowerCase() + simpleName.substring(1)``` ?

naming/src/main/java/com/alibaba/nacos/naming/healthcheck/extend/HealthCheckExtendProvider.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.ServiceLoader;
+import java.util.Set;
+
+/**
+ * @author XCXCXCXCX
+ */
+@Component
+public class HealthCheckExtendProvider implements BeanFactoryAware{
```
nkorange: The code needs to be formatted.

naming/src/main/java/com/alibaba/nacos/naming/healthcheck/HealthCheckType.java
```diff
@@ -15,24 +15,49 @@
...
      */
-    MYSQL,
+    MYSQL("mysql", AbstractHealthChecker.Mysql.class),
     /**
      * No check
      */
-    NONE
+    NONE("none", AbstractHealthChecker.None.class);
+
+    private String name;
```
nkorange: The name here is useless?

naming/src/main/java/com/alibaba/nacos/naming/healthcheck/extend/HealthCheckExtendProvider.java
nkorange: And why the class has to implement BeanFactoryAware?

nkorange:
@XCXCXCXCX The general idea of your code is fine. Besides the comments I left in the code, I have some questions:
1. Have you tested two or more jars? And what will happen if they use the same type?
2. How does user select the new health check type on console?
3. Does the code work correctly if no jar exists in the plugin directory?
4. Is this extend function able to be turned off?

naming/src/main/java/com/alibaba/nacos/naming/healthcheck/extend/HealthCheckExtendProvider.java
XCXCXCXCX: OK

naming/src/main/java/com/alibaba/nacos/naming/healthcheck/extend/HealthCheckExtendProvider.java
XCXCXCXCX: Good idea.

naming/src/main/java/com/alibaba/nacos/naming/healthcheck/extend/HealthCheckExtendProvider.java
XCXCXCXCX: OK

naming/src/main/java/com/alibaba/nacos/naming/healthcheck/HealthCheckType.java
XCXCXCXCX: Not currently, but I think it may be used in the future.

XCXCXCXCX:
> @XCXCXCXCX The general idea of your code is fine. Besides the comments I left in the code, I have some questions:
> 
> 1. Have you tested two or more jars? And what will happen if they use the same type?
> 2. How does user select the new health check type on console?
> 3. Does the code work correctly if no jar exists in the plugin directory?
> 4. Is this extend function able to be turned off?

1.In doing so, there will be coverage, but in which case do users need to use multiple health detection methods of the same type at the same time?

2.I have already added this part. The current method also requires front-end page cooperation, providing a custom json form for passing the new extended HealthChecker.

3.of course, we should not force users to use extensions.

4.no, it isn't. If necessary, we can add a configuration switch, but I don't think this is necessary.

nkorange:
> > @XCXCXCXCX The general idea of your code is fine. Besides the comments I left in the code, I have some questions:
> > 
> > 1. Have you tested two or more jars? And what will happen if they use the same type?
> > 2. How does user select the new health check type on console?
> > 3. Does the code work correctly if no jar exists in the plugin directory?
> > 4. Is this extend function able to be turned off?
> 
> 1.In doing so, there will be coverage, but in which case do users need to use multiple health detection methods of the same type at the same time?
> 
> 2.I have already added this part. The current method also requires front-end page cooperation, providing a custom json form for passing the new extended HealthChecker.
> 
> 3.of course, we should not force users to use extensions.
> 
> 4.no, it isn't. If necessary, we can add a configuration switch, but I don't think this is necessary.

1. User may mistakenly write two jars with the same type, I think it's better to determine this in the code and report an error when it occurs.

XCXCXCXCX:
> > > @XCXCXCXCX The general idea of your code is fine. Besides the comments I left in the code, I have some questions:
> > > 
> > > 1. Have you tested two or more jars? And what will happen if they use the same type?
> > > 2. How does user select the new health check type on console?
> > > 3. Does the code work correctly if no jar exists in the plugin directory?
> > > 4. Is this extend function able to be turned off?
> > 
> > 
> > 1.In doing so, there will be coverage, but in which case do users need to use multiple health detection methods of the same type at the same time?
> > 2.I have already added this part. The current method also requires front-end page cooperation, providing a custom json form for passing the new extended HealthChecker.
> > 3.of course, we should not force users to use extensions.
> > 4.no, it isn't. If necessary, we can add a configuration switch, but I don't think this is necessary.
> 
> 1. User may mistakenly write two jars with the same type, I think it's better to determine this in the code and report an error when it occurs.

I added a checksum and throw an exception when it find the same type (including the nacos native implementation).

naming/src/main/java/com/alibaba/nacos/naming/controllers/HealthController.java
```diff
@@ -110,4 +116,20 @@ public String update(HttpServletRequest request) throws Exception {
 
         return "ok";
     }
+
+    @ResponseBody
+    @RequestMapping(value = "checkers", method = RequestMethod.GET)
+    public ResponseEntity checkers(){
+        List<Class> classes = HealthCheckType.getLoadedHealthCheckerClasses();
+        Map<String, AbstractHealthChecker> demos = new HashMap<>();
```
nkorange: 为啥叫demos？

distribution/bin/startup.sh
```diff
@@ -102,10 +102,11 @@ else
 fi
 
 JAVA_OPT="${JAVA_OPT} -Dnacos.home=${BASE_DIR}"
-JAVA_OPT="${JAVA_OPT} -jar ${BASE_DIR}/target/nacos-server.jar"
+JAVA_OPT="${JAVA_OPT} -Dloader.path=${BASE_DIR}/plugins/health/nacos-health-plugin-example-1.0.jar -jar ${BASE_DIR}/target/nacos-server.jar"
```
nkorange: Script of linux should be updated too.

naming/src/main/java/com/alibaba/nacos/naming/controllers/HealthController.java
```diff
@@ -110,4 +116,20 @@ public String update(HttpServletRequest request) throws Exception {
 
         return "ok";
     }
+
+    @ResponseBody
+    @RequestMapping(value = "checkers", method = RequestMethod.GET)
+    public ResponseEntity checkers(){
```
nkorange: The code in not formatted.

PR #1161, opened 2019-04-30
[ISSUE #1118] 历史配置文件保留天数，可配置

rushsky518:

历史配置文件保留时间改成可配置

## Brief changelog
DumpService 增加 retentionDays 字段，对应配置项为 retention.days

### DIFF
--- a/config/src/main/java/com/alibaba/nacos/config/server/service/dump/DumpService.java
+++ b/config/src/main/java/com/alibaba/nacos/config/server/service/dump/DumpService.java
@@ -279,7 +279,7 @@ public class DumpService {
     }
 
     private int getRetentionDays() {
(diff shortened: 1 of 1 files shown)


config/src/main/java/com/alibaba/nacos/config/server/service/dump/DumpService.java
```diff
@@ -278,6 +278,25 @@ private Boolean isQuickStart() {
         return isQuickStart;
     }
 
+    private int getRetentionDays() {
+        String val = env.getProperty("retention.days");
```
nkorange: It's better to use property in application.properties. And the property name should be with prefix 'nacos.config'.

Issue #978, opened 2019-03-30
[BUG] Ipv6 Not perfect Support / IPv6 支持不完善


tao1991123:
## Issue Description

Type: *bug report* or *feature request*

### Describe what happened (or what feature you want)

- Use  API /nacos/v1/ns/instance to register a service with a valid ipv6, It respond 400 Bad Request. 
- 使用PUT /nacos/v1/ns/instance 注册服务的时候, 传入了一个合法的ipv6地址.得到400的响应.
在 GET /nacos/v1/ns/instance又能够获取到这个实例。

![](https://ws4.sinaimg.cn/large/006tKfTcly1g1kr65shlqj313d0u0dl0.jpg)

此外发现一个拼写错误 `malfomed ip config: fe80::3c:7d01:f92a:8dd0:7001:unknown_1.0_true_false_DEFAULT` malformed 畸形的, 少了一个r
### Describe what you expected to happen
When register a service, IPv6 should be allowed

#### Fully Support IPv6: 
- PUT /nacos/v1/ns/instance with a valid ipv6 , response success
- GET /nacos/v1/ns/instance should return ipv6 instance

#### 完全支持IPv6: 
- PUT /nacos/v1/ns/instance 传入IPv6不报错可以正常注册
- GET /nacos/v1/ns/instance 同时也能被获取到

### How to reproduce it (as minimally and precisely as possible)

1. download https://github.com/nacos-group/nacos-docker/
2. docker-compose -f example/standalone-derby.yaml up
3. curl -X PUT 'http://127.0.0.1:8848/nacos/v1/ns/instance?clusterName=DEFAULT&enable=true&healthy=true&ip=fe80::3c:7d01:f92a:8dd0&metadata={}&port=7001&serviceName=nodejs.test.domain&tenant=default&weight=1'
4. Response 400

### Tell us your environment
Docker latest image

nkorange:
@tao1991123 Will support in future, welcome to contribute to this feature.

neatlife:
@i will solve it@

PR #1164, opened 2019-05-01
fix ipv6 instance register code 400

neatlife:
支持IPv6注册 #978


neatlife:
regex from: http://vernon.mauery.com/content/projects/linux/ipv6_regex

nkorange:
@neatlife Should consider how filter IPs with the client IP type.

nkorange:
Close due to no feedback from @neatlife

Issue #1430, opened 2019-06-25
Too many error logs in naming-server.log in nacos-server


naturezhm:
## Issue Description

Type: *feature request*

### Describe what happened (or what feature you want)

Too many error logs in `naming-server.log`  in nacos-server ! （more than 10K lines per second of exception error stack）

Should change log level from "ERROR" to "WARN". because these NacosException in `com.alibaba.nacos.naming.controllers` is not **nacos system error** or **some error cause by nacos service**. Usually we can return error code when some application subscribe non-existent config. 

<img width="647" alt="WechatIMG5" src="https://user-images.githubusercontent.com/10045399/60076486-43b07780-975a-11e9-9191-0fb5ae69439b.png">

### Describe what you expected to happen

Expect print "WARN" log level for `com.alibaba.nacos.naming.controllers` throws NacosException as follows：

```java
com.alibaba.nacos.naming.exception.ResponseExceptionHandler#handleNacosException

before:  Loggers.SRV_LOG.error("got exception. {}", e.getErrorMsg(), e);
after:  Loggers.SRV_LOG.warn("got exception. {}", e.getErrorMsg(), e);
```

### How to reproduce it (as minimally and precisely as possible)

1. Start a demo under guild of https://nacos.io/en-us/docs/use-nacos-with-dubbo.html	
2. Watch `naming-server.log`  form nacos server 

   

### Tell us your environment

JDK1.8

dubbo 2.7.2

Nacos 1.0.0

PR #1431, opened 2019-06-25
NacosException调整为warn级别

naturezhm:
## Issue Description

Type: *feature request*

### Describe what happened (or what feature you want)

Too many error logs in `naming-server.log`  in nacos-server ! （more than 10K lines per second of exception error stack）

Should change log level from "ERROR" to "WARN". because these NacosException in `com.alibaba.nacos.naming.controllers` is not **nacos system error** or **some error cause by nacos service**. Usually we can return error code when some application subscribe non-existent config. 

WechatIMG5

### Describe what you expected to happen

Expect print "WARN" log level for `com.alibaba.nacos.naming.controllers` throws NacosException as follows：

```java
com.alibaba.nacos.naming.exception.ResponseExceptionHandler#handleNacosException

before:  Loggers.SRV_LOG.error("got exception. {}", e.getErrorMsg(), e);
after:  Loggers.SRV_LOG.warn("got exception. {}", e.getErrorMsg(), e);
```

### How to reproduce it (as minimally and precisely as possible)

1. Start a demo under guild of https://nacos.io/en-us/docs/use-nacos-with-dubbo.html	
2. Watch `naming-server.log`  form nacos server 

   

### Tell us your environment

JDK1.8

dubbo 2.7.2

Nacos 1.0.0

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/exception/ResponseExceptionHandler.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/exception/ResponseExceptionHandler.java
@@ -30,7 +30,7 @@ public class ResponseExceptionHandler {
 
     @ExceptionHandler(NacosException.class)
     private ResponseEntity<String> handleNacosException(NacosException e) {
-        Loggers.SRV_LOG.error("got exception. {}", e.getErrorMsg(), e);
+        Loggers.SRV_LOG.warn("got exception. {}", e.getErrorMsg(), e);
         return ResponseEntity.status(e.getErrorCode()).body(e.getMessage());
     }
 


naturezhm:
associate issue https://github.com/alibaba/nacos/issues/1430

naming/src/main/java/com/alibaba/nacos/naming/exception/ResponseExceptionHandler.java
```diff
@@ -30,7 +30,7 @@
 
     @ExceptionHandler(NacosException.class)
     private ResponseEntity<String> handleNacosException(NacosException e) {
-        Loggers.SRV_LOG.error("got exception. {}", e.getErrorMsg(), e);
+        Loggers.SRV_LOG.warn("got exception. {}", e.getErrorMsg(), e);
```
nkorange: This is the log triggered by all errors that throw NacoException, not just the 'serivce not found' error. So your modification is global and would influence other kinds of errors that throw NacoException.

Issue #1878, opened 2019-09-25
The subscribers function of Service  Management can't return the right subscriber list on cluster mode


yuchaozhou:
After looking at the code, I found that the return error was written when the synchronization loop got the subscriber.

see com.alibaba.nacos.naming.core.SubscribeManager.getSubscribers() method, at 92 line.

### Tell us your environment
Nacos 1.1.3 version

thank you.

yuchaozhou:
``

nkorange:
Can you paste the error detail?

yuchaozhou:
![图片](https://user-images.githubusercontent.com/1917747/66813822-10dbfa80-ef68-11e9-89db-de6f0b42d5fe.png)
This is a loop query information method，  but the **return statement** is written early

PR #1944, opened 2019-10-17
fix the getting subscribers error

Nicholas2015:

fix the return error when the synchronization loop got the subscriber.

## Brief changelog

fix the return error when the synchronization loop got the subscriber.


nkorange:
related issue is #1878

Issue #2237, opened 2019-12-29
com.alibaba.nacos.naming.misc.HttpClient#httpPutLarge is a little odd


rushsky518:
![image](https://user-images.githubusercontent.com/20531420/71556433-8b9ab900-2a73-11ea-8f3e-94c9c2c1fb16.png)

![image](https://user-images.githubusercontent.com/20531420/71556467-12e82c80-2a74-11ea-851e-23509a933068.png)

byte[] content -> String
then
String -> byte[]

i suggest ByteArrayEntity. Can i change it?

nkorange:
Welcome to contribute!

rushsky518:
@i will solve it@

PR #2287, opened 2020-01-10
HttpCLient uses ByteArrayEntity when sync data

rushsky518:
HttpCLient uses ByteArrayEntity when sync data

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/controllers/DistroController.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/controllers/DistroController.java
@@ -103,13 +103,13 @@ public class DistroController {
             datumMap.put(key, datum);
         }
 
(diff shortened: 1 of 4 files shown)


KomachiSion:
Can you create issue first to discuss why use ByteArrayEntity to replace StringEntity?

rushsky518:
@KomachiSion 
````java
public StringEntity(final String string, final ContentType contentType) throws UnsupportedCharsetException {
	super();
	Args.notNull(string, "Source string");
	Charset charset = contentType != null ? contentType.getCharset() : null;
	if (charset == null) {
		charset = HTTP.DEF_CONTENT_CHARSET;
	}
	this.content = string.getBytes(charset);
	if (contentType != null) {
		setContentType(contentType.toString());
	}
}
````
byte[] -> String(char[]) -> byte[], this may be unnecessary.
And issue is #2237

Issue #2401, opened 2020-03-03
timed sync task failed


lkxiaolou:
### Describe what happened (or what feature you want)

2020-03-03 04:06:47,613 ERROR timed sync task failed.

java.lang.ArrayIndexOutOfBoundsException: null
        at sun.security.provider.DigestBase.engineUpdate(DigestBase.java:115)
        at sun.security.provider.MD5.implDigest(MD5.java:109)
        at sun.security.provider.DigestBase.engineDigest(DigestBase.java:183)
        at sun.security.provider.DigestBase.engineDigest(DigestBase.java:162)
        at java.security.MessageDigest$Delegate.engineDigest(MessageDigest.java:592)
        at java.security.MessageDigest.digest(MessageDigest.java:365)
        at java.security.MessageDigest.digest(MessageDigest.java:411)
        at com.alibaba.nacos.naming.core.Instances.recalculateChecksum(Instances.java:88)
        at com.alibaba.nacos.naming.core.Instances.getChecksum(Instances.java:72)
        at com.alibaba.nacos.naming.consistency.ephemeral.distro.DataSyncer$TimedSync.run(DataSyncer.java:179)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)

### Tell us your environment
java8 +nacos 1.1.4

PR #2403, opened 2020-03-03
fix checksum bug #2401

lkxiaolou:

fix the checksum bug, refer to #2401 

MessageDigest is not safe in multi-thread 

## Brief changelog
Add a lock to ensure execute success

## Verifying this change

see InstancesTest


naming/src/main/java/com/alibaba/nacos/naming/core/Instances.java
```diff
@@ -84,8 +86,10 @@ private String recalculateChecksum() {
         }
 
         if (MESSAGE_DIGEST != null) {
-            checksum =
-                new BigInteger(1, MESSAGE_DIGEST.digest((sb.toString()).getBytes(Charset.forName("UTF-8")))).toString(16);
+            synchronized (LOCK) {
```
chuntaojun: Can ThreadLocal be used to avoid lock contention

lkxiaolou: okay

lkxiaolou: > Can ThreadLocal be used to avoid lock contention

done

naming/src/main/java/com/alibaba/nacos/naming/core/Instances.java
```diff
@@ -83,9 +84,10 @@ private String recalculateChecksum() {
             sb.append(",");
         }
 
-        if (MESSAGE_DIGEST != null) {
-            checksum =
-                new BigInteger(1, MESSAGE_DIGEST.digest((sb.toString()).getBytes(Charset.forName("UTF-8")))).toString(16);
+        if (MESSAGE_DIGEST_LOCAL.get() != null) {
```
chuntaojun: Here, you can consider holding the MessageDigest with a local variable to avoid getting again later

lkxiaolou: > Here, you can consider holding the MessageDigest with a local variable to avoid getting again later

done

lkxiaolou:
@chuntaojun I have unified the md5 calculation in thread-local way instead of using a lock

Issue #2618, opened 2020-04-13
[1.2.1]The config service sometimes generates MD5 with a length of 31


amenema:
This writing method in the **com.alibaba.nacos.common.utils.getMD5(byte[] bytes)**  tool class of the server will lead to the generation of MD5 with length of 31 under specific conditions.But the **com.alibaba.nacos.client.config.utils.getMD5String(String content)** tool of the client does not find the problem at present, so it will cause the MD5 generated by the client and the server to be inconsistent. So that the client always requests to update the corresponding data ID.    
__
客户端与服务端对config.content 的md5编写方式不一致，导致在特定参数下，服务端生成了31位的MD5，从而使客户端一直刷新数据。BigInteger.toString(16)会删除前导0，所以不应该用这种方式。
### case

```
bailing:
 http:
  protocol: http
```

wangweizZZ:
@i will solve it@

connnywilmeringBA:
https://www.youtube.com/watch?v=uFYQv9HDAXk

PR #2647, opened 2020-04-18
[ISSUE #2618]modify config service md5 generation method

wangweizZZ:

#2618 

## Brief changelog
modify  `com.alibaba.nacos.common.utils.Md5Utils`
use `org.apache.commons.codec`  generate md5 string

## Verifying this change

XXXX


chuntaojun:
Just unify the MD5 generation on both sides using the same utility class

chuntaojun:
Drop the MD5 utility classes into the common module

common/src/main/java/com/alibaba/nacos/common/utils/MD5.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.alibaba.nacos.client.config.utils;
```
chuntaojun, line 16: The utility class is already at the end of Utils

common/pom.xml
```diff
@@ -44,6 +44,11 @@
             <artifactId>commons-io</artifactId>
         </dependency>
 
+        <dependency>
```
chuntaojun, line 47: It looks like this dependency is not used, so it can be removed

wangweizZZ: yes, i agree

common/src/main/java/com/alibaba/nacos/common/utils/MD5.java
wangweizZZ: sorry,I don't understand what you mean. the utility class is move from client module to common module. do you mean  name  should like MD5Utils  end with Utils?

chuntaojun: yes

config/src/main/java/com/alibaba/nacos/config/server/aspect/RequestLogAspect.java
```diff
@@ -67,7 +66,7 @@
     public Object interfacePublishSingle(ProceedingJoinPoint pjp, HttpServletRequest request,
                                          HttpServletResponse response, String dataId, String group, String tenant,
                                          String content) throws Throwable {
-        final String md5 = content == null ? null : Md5Utils.getMD5(content, Constants.ENCODE);
+        final String md5 = content == null ? null : MD5.getInstance().getMD5String(content);
```
chuntaojun, line 69: I think the MD5 implementation will not be as good as the MD5Utils you delete because of the locking mechanism, you can only modify the string method of MD5Utils

old

```
return new BigInteger(1, messageDigest.digest(bytes)).toString(HEX_VALUE_COUNT);
```

new

```java
public String bytes2string(byte[] bt) {
        int l = bt.length;

        char[] out = new char[l << 1];

        for (int i = 0, j = 0; i < l; i++) {
            out[j++] = digits[(0xF0 & bt[i]) >>> 4];
            out[j++] = digits[0x0F & bt[i]];
        }

        return new String(out);
    }
```

wangweizZZ: Thanks for your patient.I choose this one ,consider MessageDigest classes  is instantiated only once.

PR #2716, opened 2020-04-30
change log level

hsoftxl:
change log level


KomachiSion:
Can we discuss which log need change level first?

yanlinly:
Most of log is important in runtime，If the specific log level has a problem, we will discuss it separately

PR #2745, opened 2020-05-06
[ISSUE #2739]: modify the way of get remote state when received beat

horizonzy:

modify the way of get remote state when received beat


KomachiSion:
I think we should not merge this PR, reason follow:
1. beat info update cache data from the remote data is ok.
2. We will replace this raft with SOFAJRaft.

horizonzy:
> I think we should not merge this PR, reason follow:
> 
> 1. beat info update cache data from the remote data is ok.
> 2. We will replace this raft with SOFAJRaft.

yep.i will close it

PR #2838, opened 2020-05-20
Solve the security problem of the old version of fastjson

wcc526:
Fastjson has a serious security problem in 1.2.58,which will cause RCE
https://www.anquanke.com/post/id/199527
https://www.cnblogs.com/tr1ple/p/12348886.html
https://github.com/jas502n/fastjson-1.2.58-rce

XXXXX

## Brief changelog

Solve the security problem of the old version of fastjson

## Verifying this change

Solve the security problem of the old version of fastjson

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -553,7 +553,7 @@
             <dependency>
                 <groupId>com.alibaba</groupId>
                 <artifactId>fastjson</artifactId>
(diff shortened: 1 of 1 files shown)


chuntaojun:
We will remove fastjson

Issue #2860, opened 2020-05-22
Maven dependency optimizations, where only one version of each dependency exists


chuntaojun:
## Issue Description

Type: *feature request*

### Describe what happened (or what feature you want)

background
1. At present, the internal dependencies of nacos are relatively disordered, and the unused and repeated dependencies need to be removed
2. For example, if you rely on netty, it is best to ensure that you only rely on one netty version, and do not have multiple netty versions.
Ideas
1. Clearly sort out the current maven dependencies
2. For version control, it is best to use the form of properties

### How to reproduce it (as minimally and precisely as possible)

1. 
2. 
3.

jintonghuoya:
@i will solve it@

PR #2921, opened 2020-05-29
[ISSUE #2860]Maven dependency optimizations, where only one version of each dependency exists 

jintonghuoya:

Unify maven dependency and plugin version

## Brief changelog

Mange the whole version of maven dependencies and plugins in root pom file.
Remove the version of  maven dependencies and plugins in children's pom file.

## Verifying this change

Cause no logical code change, so there is no need to write unit test to verify the change.


jintonghuoya:
有冲突，我需要再提一次PR

PR #2923, opened 2020-05-29
[ISSUE #2860]Maven dependency optimizations, where only one version of each dependency exists

jintonghuoya:

Unify maven dependency and plugin version

## Brief changelog

Mange the whole version of maven dependencies and plugins in root pom file.
Remove the version of maven dependencies and plugins in children's pom file.

## Verifying this change

Cause no logical code change, so there is no need to write unit test to verify the change.


cmdb/pom.xml
```diff
@@ -27,8 +27,7 @@
     <packaging>jar</packaging>
 
     <name>nacos-cmdb ${project.version}</name>
-    <!-- FIXME change it to the project's website -->
-    <url>http://www.example.com</url>
```
chuntaojun, line 31: It should be the nacos project address

jintonghuoya: So how about "https://www.nacos.io/"?

jintonghuoya: So how about "https://nacos.io/"?

Issue #2810, opened 2020-05-15
the jugde logic exist problem when send beat in raftcore


horizonzy:
## Issue Description

Type: *bug report*

### Describe what happened (or what feature you want)
```
 public void sendBeat() throws IOException, InterruptedException {
            RaftPeer local = peers.local();
            if (local.state != RaftPeer.State.LEADER && !STANDALONE_MODE) {
                return;
            }
            if (Loggers.RAFT.isDebugEnabled()) {
                Loggers.RAFT.debug("[RAFT] send beat with {} keys.", datums.size());
            }
    ....................
}

```
if STANDALONE_MODE is true,it shoule return directly.

PR #3047, opened 2020-06-12
[ISSUE #2810] don't send beat when standalone model

horizonzy:
fix-#2810


naming/src/main/java/com/alibaba/nacos/naming/consistency/persistent/raft/RaftCore.java
```diff
@@ -484,10 +484,9 @@ public void run() {
 
         public void sendBeat() throws IOException, InterruptedException {
             RaftPeer local = peers.local();
-            if (local.state != RaftPeer.State.LEADER && !ApplicationUtils.getStandaloneMode()) {
+            if (ApplicationUtils.getStandaloneMode() || local.state != RaftPeer.State.LEADER){
```
KomachiSion, line 487: There should be one blank between `)` and `{` please

Issue #3056, opened 2020-06-12
EmbeddedStoragePersistServiceImpl  and ExternalStoragePaginationHelperImpl use “+=”   to concatenated String, will cause performance  issue


showkawa:
## Issue Description
*bug report* - performance  issue

### Describe what happened (or what feature you want)
com.alibaba.nacos.config.server.service.repository.EmbeddedStoragePersistServiceImpl 
and
com.alibaba.nacos.config.server.service.repository.ExternalStoragePaginationHelperImpl
use “+=”   to concatenated String, will cause performance  issue
### Describe what you expected to happen 
 use StringBuilder will be better, and i will handle it 

### How to reproduce it (as minimally and precisely as possible)

1. replace "+=" to StringBuilder append  method
2. 
3.

showkawa:
@i will solve it@

chuntaojun:
Simplify the label of issue

PR #3057, opened 2020-06-12
#3056 performance issue

showkawa:
replace "+=" to StringBuilder append method


showkawa:
@chuntaojun please review -

PR #3236, opened 2020-07-03
optimize isExistTable function 

### LINKED ISSUE #3235: isExistTable function may be particularly slow
about class : `EmbeddedStoragePersistServiceImpl.java` and `ExternalStoragePersistServiceImpl.java`

If the table data volume is particularly large, count (*) will be particularly slow

```
    public boolean isExistTable(String tableName) {
        String sql = "SELECT COUNT(*) FROM " + tableName;
        try {
            databaseOperate.queryOne(sql, Integer.class);
            return true;
        } catch (Throwable e) {
            return false;
        }
    }
```

jiangcaijun:

fix #3235 

## Brief changelog

`String.format("select 1 from %s order by id limit 1", tableName);`

or

`String.format("select 1 from %s where id = 1", tableName);`

Both of these SQL statements are very fast, and can be viewed through `explain`

### DIFF
--- a/config/src/main/java/com/alibaba/nacos/config/server/service/repository/embedded/EmbeddedStoragePersistServiceImpl.java
+++ b/config/src/main/java/com/alibaba/nacos/config/server/service/repository/embedded/EmbeddedStoragePersistServiceImpl.java
@@ -2113,7 +2113,7 @@ public class EmbeddedStoragePersistServiceImpl implements PersistService {
     }
     
     public boolean isExistTable(String tableName) {
-        String sql = String.format("select 1 from %s order by id limit 1", tableName);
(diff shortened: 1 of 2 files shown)


KomachiSion:
@chuntaojun Help to check the change whether effect feature.

@jiangcaijun We will merge it after 1.3.1 release

chuntaojun:
select 1 from table limit 1。just fine，Do not specify a field order, because this field may not exist

jiangcaijun:
done ;  please check

atellwu:
这个方法好像不符合预期，即使表存在，里面没有记录的话，会抛出异常：Incorrect result size: expected 1, actual 0

image

Issue #3370, opened 2020-07-18
Issue with executing shell script /bin/sh^M on Unix/Linux


Veryfirefly:
## Issue Description
Type: Bug Report

### Describe what happened (or what feature you want)

1. After nacos is packaged, the /bin/sh^M problem will appear when executing the shell script on Linux. This is caused by script line breaks instead of LF. I checked the script in the source code and found that startup.sh and shutdown.sh are CRLF formatted line breaks
2. `ERROR OS=Windows and the assembly descriptor contains a *nix-specific root-relative-reference`. This error occurs during binary packaging in Maven, but it will not affect the packaging. I queried the documentation and found that the reason is: because the assembly task is executed on Windows, but it contains a path relative to Linux.

### Describe what you expected to happen

I submitted a PR and revised the above two questions

1. Format startup.sh and shutdown.sh. Use Unix/Linux standard line break LF to avoid converting line breaks during execution, and tested on Linux and Unix.
2. Solve the OS=Windows error when packaging Maven. Successfully packaged on Linux and Windows

### Tell us your environment

1. JDK 1.8
2. CentOS 6.9

Veryfirefly:
#3371

KomachiSion:
I have comment in your PR. Please check.

PR #3371, opened 2020-07-18
[ISSUE #3370] Issue with executing shell script /bin/sh^M on Unix/Linux

Veryfirefly:

1. Fixed the /bin/sh^M problem when executing startup.sh and shutdown.sh scripts on Linux.
2. Fixed the following problems when performing binary packaging: `ERROR OS = Windows, and the assembly descriptor contains *nix specific root-relative-reference`

## Brief changelog

<table style="border-collapse: collapse;">
    <tr>
	  <td>Path</td>
	  <td>Change</td>
	</tr>
    <tr>
	  <td>nacos/distribution/bin/startup.sh</td>
	  <td>CRLF -> LF</td>
	</tr>
    <tr>
	  <td>nacos/distribution/bin/shutdown.sh</td>
	  <td>CRLF -> LF</td>
	</tr>
    <tr>
	  <td>nacos/distribution/release-nacos.xml</td>
	  <td>nacos-server/outputDirectory</td>
	</tr>
</table>


KomachiSion:
1. I check the develop branch, the four bin files is `LF` already. I think we have no need to change.
2. the `nacos-server/outputDirectory`, It has been able to run well, I do not know why it needs to be modified.

Veryfirefly:
> 1. I check the develop branch, the four bin files is `LF` already. I think we have no need to change.
> 2. the `nacos-server/outputDirectory`, It has been able to run well, I do not know why it needs to be modified.

The `nacos-server/outputDirectory`,  The following error occurs when packaging: `[ERROR] OS=Windows and the assembly descriptor contains a *nix-specific root-relative-reference (starting with slash) /target/`. But this does not affect the packaging. This will give the developer an error message, similar to warning

 I queried the documentation and found that the reason is: because the assembly task is executed on Windows, but it contains a path relative to Linux. The document states that `outputDirectory` defines a `/` which would in consequence mean to extract all contens to always same location.

This is my image when packing
OS=Windows

KomachiSion:
My env is MacOs, has no error. 

Can someone in the community help to confirm whether this change can eliminate the error message and be packaged correctly?

Veryfirefly:
> My env is MacOs, has no error.
> 
> Can someone in the community help to confirm whether this change can eliminate the error message and be packaged correctly on windows OS?

Ok i am looking for help

zongtanghu:
image

zongtanghu:
I have already verified this in my local windows enviroment.The result is as above.

zongtanghu:
And If no anything modified, the result is as below:

image

Veryfirefly:
> And If no anything modified, the result is as below:
> 
> image

Thanks for help

Issue #3388, opened 2020-07-20
Deleted a role for an user, but this user staill has the permission about the deleted role. 


yanjunnf:
## Issue Description

Type: *bug report* or *feature request*

### Describe what happened (or what feature you want)
When deleted a role(role1) for an user, the user should loss the associated permission of role1, but in fact, the user still has the permission of role1.

### How to reproduce it (as minimally and precisely as possible)

1.  Create an user (user1)
2. Check the persmission for user1
2.  Create a role(role1) which links the user user1
3.  Delete the role1 
4. Check the permission for user1 again

### Tell us your environment
Nacos1.3.1
nacos.core.auth.enabled=true
nacos.core.auth.caching.enabled=true

yanjunnf:
@i will solve it@

PR #3419, opened 2020-07-23
[#3388]Remove the deleted roles and permissions info during reloading role info

yanjunnf:

#3388

## Brief changelog
We should remove these already deleted roles, permissions during reloading nacos role info.

## Verifying this change

XXXX

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/config/NacosConfigService.java
+++ b/client/src/main/java/com/alibaba/nacos/client/config/NacosConfigService.java
@@ -28,21 +28,20 @@ import com.alibaba.nacos.client.config.http.HttpAgent;
 import com.alibaba.nacos.client.config.http.MetricsHttpAgent;
 import com.alibaba.nacos.client.config.http.ServerHttpAgent;
 import com.alibaba.nacos.client.config.impl.ClientWorker;
-import com.alibaba.nacos.client.config.impl.HttpSimpleClient.HttpResult;
 import com.alibaba.nacos.client.config.impl.LocalConfigInfoProcessor;
 import com.alibaba.nacos.client.config.utils.ContentUtils;
 import com.alibaba.nacos.client.config.utils.ParamUtils;
 import com.alibaba.nacos.client.utils.LogUtils;
 import com.alibaba.nacos.client.utils.ParamUtil;
 import com.alibaba.nacos.client.utils.ValidatorUtils;
+import com.alibaba.nacos.common.http.HttpRestResult;
 import com.alibaba.nacos.common.utils.StringUtils;
 import org.slf4j.Logger;
 
-import java.io.IOException;
 import java.net.HttpURLConnection;
-import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.List;
+import java.util.HashMap;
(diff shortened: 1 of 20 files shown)


yanjunnf:
Why do not clear roleSet, roleInfoMap and permissionInfoMap before putAll/addAll?
The clear method cannot provide the whole logic as a atomic operation.
For example, roleSet.clear() finished to execute, but now this thread is suspended, then antoher thread try to get all roles, it will get a empty list since the roleSet object was cleared.

yanjunnf:
license/cla Pending — Contributor License Agreement is not signed yet.
I really signed several times, but this message still shows.

KomachiSion:
> license/cla Pending — Contributor License Agreement is not signed yet.
> I really signed several times, but this message still shows.

The reason is your git username and email is not your github id.
So the git commit log author is junyan29, but you signed CLA is for yanjunnf.

It's different.

Solution is to set the git username and email to your github id.

Search in search engine, you can get many answer.

KomachiSion:
And for the PR content, I describe in PR before you submit.

Why don't clear all map and then putAll again, but to remove one by one. I think the new put is all data for users and permissions.

yanjunnf:
> And for the PR content, I describe in PR before you submit.
> 
> Why don't clear all map and then putAll again, but to remove one by one. I think the new put is all data for users and permissions.
> 
> @chuntaojun Is my understanding correct?

The clear method cannot guarantee the whole logic as a atomic operation.
For example, roleSet.clear() finished to execute, but at this time, this thread is suspended, then antoher thread try to get all roles, it will get a empty list since the roleSet object was cleared, that may cuase the empty role/permission list showes on UI part.
Remove one by one, that can reduce impaction.

chuntaojun:
I think this is too complicated, but logically, there are only reads to the map, not writes. You can take advantage of the volatile property, replace the application directly, and then destroy the old Map

KomachiSion:
I agree with @chuntaojun, if you think about the thread safe, you can use a tmp map/set to save new data from persistence, and directly replace them.

KomachiSion:
Once more about the CLA, I check your commit log, it contain many `merge branch`, I recommend that you use rebase to squash them. And if there are still `junyan29`'s commit. I think you should revert them and re commit with `yanjunnf`

yanjunnf:
Thanks for your reply.
Yes， I agree with you in priciple about keeping the code clean and simple.
I though about to use tmp object and replace, If that we have to remove “final “ modifier for class members roleSet, roleInfoMap and permissionInfoMap.

KomachiSion:
If there map/set need to be reset, remove `final` will be ok. @chuntaojun What's your idea?

chuntaojun:
agree

发自我的iPhone

在 2020年7月24日，17:03，杨翊 SionYang <notifications@github.com> 写道：

﻿

If these map/set need to be reset, remove final will be ok. @chuntaojun<https://github.com/chuntaojun> What's your idea?

yanjunnf:
Closed it and recreate a new PR([#3388]Reset nacos roles, permissions during reloading role info from database) to fix issue 3388 since CLA problem.

PR #3427, opened 2020-07-23
Update issue templates

KomachiSion:
I want to update the issue template.
Split to `bug report` and `feature request`.

In `bug report`, users need to fill more information like nacos version, env information. and need them describe more detail for situation of bugs.

For `feature request`, is a recommend template of feature request.

The old template will also be saved. But it may delete in future.


Lorena-babygirl (commented):
update issue templates.

Issue #3509, opened 2020-08-04
The server address server mode cannot obtain the application.properties Variables in（ address.server.domain=xxx . xxx.com )


qm0123:
**Describe the bug**
中文:服务端 address-server 模式无法获取从application.properties中的变量(address.server.domain=xxx.xxx.com)
英文:The server address server mode cannot obtain the application.properties Variables in（ address.server.domain=xxx . xxx.com )

**Expected behavior**
中文:可以从配置文件中获取到系统变量, 当前的方式是 系统变量和启动jvm中获取, 也应该支持从配置文件中获取. 并且增加文档描述
    获取配置的优先级:系统环境变量 -> 启动jvm(-D) -> properties
英文: the system variables can be obtained from the configuration file. The current way is to obtain the system variables from the system variables and the startup JVM. It should also support obtaining system variables from the configuration files.And add document description
Get the priority of configuration: system environment variables > start JVM (- D) - > properties

**Acutally behavior**
A clear and concise description of what you actually to happen.

**How to Reproduce**
1. application.properties配置 address.server.domain=xxx.xxx.com
2. 启动 debug: address.server.domain= null

![4612C75B-D12A-466a-85B4-931B7ABD043B](https://user-images.githubusercontent.com/67568757/89245522-9043d300-d63b-11ea-9fea-86af997239fe.png)

**Desktop (please complete the following information):**
 - OS: windows
 - Version nacos-server 1.3.1

qm0123:
If this needs to be addressed,@i will solve it@

KomachiSion:
Welcome to contribute. Thanks

KomachiSion:
@chuntaojun What‘s your design? Can it be set by users?

PR #3516, opened 2020-08-04
[#3509] Fix address server mode cannot be obtained application.properties

qm0123:
#  What is the purpose of the change

see #3509 

## Brief changelog

1. Use ApplicationUtils.getProperty () get variable

2. Synchronize the parameters in properties

3. After the change, get the priority system environment variable of the variable, then start the JVM (- D) - > properties property

### DIFF
--- a/core/src/main/java/com/alibaba/nacos/core/cluster/lookup/AddressServerMemberLookup.java
+++ b/core/src/main/java/com/alibaba/nacos/core/cluster/lookup/AddressServerMemberLookup.java
@@ -43,30 +43,30 @@ import java.util.Map;
  * @author <a href="mailto:liaochuntao@live.com">liaochuntao</a>
  */
 public class AddressServerMemberLookup extends AbstractMemberLookup {
-
+    
     private final GenericType<RestResult<String>> genericType = new GenericType<RestResult<String>>() {
     };
-
+    
     public String domainName;
-
+    
     public String addressPort;
-
+    
     public String addressUrl;
-
+    
     public String envIdUrl;
-
+    
     public String addressServerUrl;
-
+    
     private volatile boolean isAddressServerHealth = true;
-
+    
     private int addressServerFailCount = 0;
-
+    
     private int maxFailCount = 12;
-
+    
     private NSyncHttpClient syncHttpClient = HttpClientManager.getSyncHttpClient();
-
+    
     private volatile boolean shutdown = false;
-
+    
     @Override
     public void start() throws NacosException {
         if (start.compareAndSet(false, true)) {
@@ -75,7 +75,7 @@ public class AddressServerMemberLookup extends AbstractMemberLookup {
             run();
         }
     }
-
+    
     private void initAddressSys() {
         String envDomainName = System.getenv("address_server_domain");
         if (StringUtils.isBlank(envDomainName)) {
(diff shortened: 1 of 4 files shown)


chuntaojun:
请按照贡献文档对你的代码进行格式化下

core/src/main/java/com/alibaba/nacos/core/cluster/lookup/AddressServerMemberLookup.java
```diff
@@ -43,30 +43,30 @@
  * @author <a href="mailto:liaochuntao@live.com">liaochuntao</a>
  */
 public class AddressServerMemberLookup extends AbstractMemberLookup {
-    
+
```
KomachiSion, line 46: Do not change indent.
Please use nacos code style to format.
see src/style/codeStyle.md

qm0123: ok,copy that.

distribution/conf/application.properties
```diff
@@ -37,8 +37,8 @@ server.port=8848
 
 ### Connect URL of DB:
 # db.url.0=jdbc:mysql://127.0.0.1:3306/nacos?characterEncoding=utf8&connectTimeout=1000&socketTimeout=3000&autoReconnect=true&useUnicode=true&useSSL=false&serverTimezone=UTC
-# db.user=nacos
-# db.password=nacos
+# db.user=root
```
KomachiSion, line 40: do not change the default value

distribution/conf/application.properties
```diff
@@ -37,8 +37,8 @@ server.port=8848
...
+# db.password=1017
```
KomachiSion, line 41: do not change the default value

distribution/conf/application.properties.example
```diff
@@ -29,16 +29,16 @@ server.port=8848
...
 # spring.datasource.platform=mysql
 
 ### Count of DB:
 # db.num=1
 
 ### Connect URL of DB:
 # db.url.0=jdbc:mysql://127.0.0.1:3306/nacos?characterEncoding=utf8&connectTimeout=1000&socketTimeout=3000&autoReconnect=true&useUnicode=true&useSSL=false&serverTimezone=UTC
-# db.user=nacos
-# db.password=nacos
+# db.user=root
```
KomachiSion, line 40: do not change the default value

distribution/conf/application.properties.example
```diff
@@ -29,16 +29,16 @@ server.port=8848
...
+# db.password=1017
```
KomachiSion, line 41: do not change the default value

distribution/conf/application.properties
qm0123: I think it should be combined with (console-src-main-resources-application.properties)The default value of is the same, but nacos:nacos Is the default value, I will change the default value to nacos:nacos

distribution/conf/application.properties
qm0123: I think it should be combined with (console-src-main-resources-application.properties)The default value of is the same, but nacos:nacos Is the default value, I will change the default value to nacos:nacos

distribution/conf/application.properties.example
qm0123: I think it should be combined with (console-src-main-resources-application.properties)The default value of is the same, but nacos:nacos Is the default value, I will change the default value to nacos:nacos

distribution/conf/application.properties.example
qm0123: I think it should be combined with (console-src-main-resources-application.properties)The default value of is the same, but nacos:nacos Is the default value, I will change the default value to nacos:nacos

PR #3553, opened 2020-08-08
ISSUE #3530 add button for user role page

zzq1314zll:
ISSUE #3530 add button for user role page


KomachiSion:
There are conflict in main.js please solve it. thanks.

Issue #3613, opened 2020-08-16
Unit test method  can not be static in `ConfigTest.java`


Xarrow:
**Describe the bug**
Unit test method  can not be static in `ConfigTest`

**Expected behavior**
Fix it and  update method of `test`

**Acutally behavior**

**How to Reproduce**
execute unit test in ` ConfigTest.java`

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version : Nacos dev branch
 - Module : nacos-client
 - SDK :None

**Additional context**
Nothing

PR #3614, opened 2020-08-16
[ISSUE #3613]  Fix `unit test method not be static` & update publish config listener  in `ConfigTest.java`

Xarrow:

Fix issue #3613 

## Brief changelog

1. remove `static` in method of `before`, `test`.

2.  update new  method of `test`.


client/src/test/java/com/alibaba/nacos/client/ConfigTest.java
```diff
@@ -29,52 +30,76 @@
 
 import java.util.Properties;
 import java.util.Scanner;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 @Ignore
 public class ConfigTest {
-    
+
```
KomachiSion, line 37: Please do not change the indent.

Please use nacos CodeStyle to reformat code

Xarrow: Already fixed it and reformatted code  using  nacos CodeStyle.

Issue #3628, opened 2020-08-18
[Enhancement] Use cache time set by server when subscribe non-exist service.


KomachiSion:
**Is your feature request related to a problem? Please describe.**

When users subscribe a service, the client will create a thread try to update the client cache. See `HostReactor#UpdateTask`. But when the service is not exist, server return an empty service without the cache time so that the interval time only use default `1 second`, which will make it cost a lot resource and can't be controlled.

**Describe the solution you'd like**

Return cache time set by server even the service is not exist so that the client update task interval can be controlled.
And if there are some exception in update task, the next update need to have more interval.

**Describe alternatives you've considered**

Add update task when service exist and subscribe success.

**Additional context**
Nothing

KomachiSion:
The relative issue #3119

horizonzy:
@KomachiSion I will solve it

PR #3637, opened 2020-08-19
[ISSUE #3628] set naming client updateTask interval more flexible

horizonzy:

fix-#3628

### DIFF
--- a/api/src/main/java/com/alibaba/nacos/api/naming/pojo/ServiceInfo.java
+++ b/api/src/main/java/com/alibaba/nacos/api/naming/pojo/ServiceInfo.java
@@ -48,8 +48,6 @@ public class ServiceInfo {
     
     private long cacheMillis = 1000L;
     
-    private long failCount = 0L;
-    
     private List<Instance> hosts = new ArrayList<Instance>();
     
     private long lastRefTime = 0L;
@@ -87,18 +85,6 @@ public class ServiceInfo {
     public ServiceInfo(String name, String clusters) {
         this.name = name;
         this.clusters = clusters;
-    }
-    
-    public void incFailCount() {
-        this.failCount++;
-    }
-    
-    public long getFailCount() {
-        return failCount;
-    }
-    
-    public void setFailCount(long failCount) {
-        this.failCount = failCount;
     }
     
     public int ipCount() {
--- a/api/src/main/java/com/alibaba/nacos/api/naming/pojo/ServiceInfo.java
+++ b/api/src/main/java/com/alibaba/nacos/api/naming/pojo/ServiceInfo.java
@@ -48,6 +48,8 @@ public class ServiceInfo {
     
     private long cacheMillis = 1000L;
     
+    private long failCount = 0L;
+    
     private List<Instance> hosts = new ArrayList<Instance>();
     
     private long lastRefTime = 0L;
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/alibaba/nacos/api/naming/pojo/ServiceInfo.java
```diff
@@ -48,6 +48,8 @@
     
     private long cacheMillis = 1000L;
     
+    private long failCount = 0L;
```
KomachiSion, line 51: fail count should not be attribute of ServiceInfo. It should be maintained in UpdateTask.

client/src/main/java/com/alibaba/nacos/client/naming/core/HostReactor.java
```diff
@@ -411,7 +412,7 @@ public void run() {
                     return;
                 }
                 
-                delayTime = serviceObj.getCacheMillis();
+                delayTime = serviceObj.getCacheMillis() + Math.min(serviceObj.getFailCount() * 1000L, 10000L);
```
KomachiSion, line 415: The exception delay time should be calculate in `catch`
And the algorithm should be enhance. Such as double time then last time. And the max delay time is not more than 60s.

naming/src/main/java/com/alibaba/nacos/naming/controllers/InstanceController.java
```diff
@@ -516,21 +516,21 @@ public ObjectNode doSrvIpxt(String namespaceId, String serviceName, String agent
         ClientInfo clientInfo = new ClientInfo(agent);
         ObjectNode result = JacksonUtils.createEmptyJsonNode();
         Service service = serviceManager.getService(namespaceId, serviceName);
-        
+        long cacheMillis = switchDomain.getDefaultCacheMillis();
```
KomachiSion, line 519: If updPort > 0 , it should be pushCacheMillis. 
See the logic in 541 line.

KomachiSion (changes requested):

client/src/main/java/com/alibaba/nacos/client/naming/core/HostReactor.java
```diff
@@ -412,12 +419,14 @@ public void run() {
...
                 delayTime = serviceObj.getCacheMillis();
-                
+                failCount = 0;
             } catch (Throwable e) {
+                failCount++;
                 NAMING_LOGGER.warn("[NA] failed to update serviceName: " + serviceName, e);
             } finally {
                 if (delayTime > 0) {
-                    executor.schedule(this, delayTime, TimeUnit.MILLISECONDS);
+                    executor.schedule(this, Math.min(delayTime << failCount, DEFAULT_DELAY * 60),
```
KomachiSion, line 428: The implementation may cause `delayTime << failCount` become negative number when failCount=54.

KomachiSion (changes requested):

client/src/main/java/com/alibaba/nacos/client/naming/core/HostReactor.java
horizonzy: now, the failCunt cat't limit 6.

KomachiSion (approved):

Issue #3695, opened 2020-08-26
Logging in with the wrong username or password will cause jackson serialization results to fail


gitdoit:
**Describe the bug**
Logging in with the wrong username or password will cause jackson serialization results to fail, because the response header contains null key,and after catching this exception, the username and password will be printed,i think that is dangerous.

**Expected behavior**
Do not print the password after the login fails, just print the message in the error log

**Acutally behavior**
Log in with the wrong username or password, jackson serialization login response  failed

**How to Reproduce**
1. Login with wrong username or password
2. You will see the following error log in the console

2020-08-26 15:52:40.709 ERROR 9140 --- [.naming.updater] c.a.nacos.client.security.SecurityProxy  : [SecurityProxy] login http request failed url: http://192.168.10.64:8848/nacos/v1/auth/users/login, params: {username=nacos}, bodyMap: {password=abc}, errorMsg: errCode: 100, errMsg: Nacos serialize for class [com.alibaba.nacos.common.http.HttpRestResult] failed. 

**Desktop (please complete the following information):**
 - OS: windows
 - Version nacos-server 1.3.2
 - Module client

Technoboy-:
Try to modify SecurityProxy#login at line 153, print stackTrace instead of message,  and paste it.

gitdoit:
![image](https://user-images.githubusercontent.com/28971980/91286186-b3881b00-e7c0-11ea-82d4-f30eedcfa106.png)

`com.alibaba.nacos.api.exception.runtime.NacosSerializationException: errCode: 100, errMsg: Nacos serialize for class [com.alibaba.nacos.common.http.HttpRestResult] failed.  
	at com.alibaba.nacos.common.utils.JacksonUtils.toJson(JacksonUtils.java:61)
	at com.alibaba.nacos.client.security.SecurityProxy.login(SecurityProxy.java:146)
	at com.alibaba.nacos.client.security.SecurityProxy.login(SecurityProxy.java:113)
	at com.alibaba.nacos.client.config.http.ServerHttpAgent.<init>(ServerHttpAgent.java:279)
	at com.alibaba.nacos.client.config.NacosConfigService.<init>(NacosConfigService.java:85)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
[...]

PR #3698, opened 2020-08-26
Fix:Login failure causes Jackson serialization exception

gitdoit:
fix: #3695

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/security/SecurityProxy.java
+++ b/client/src/main/java/com/alibaba/nacos/client/security/SecurityProxy.java
@@ -143,7 +143,7 @@ public class SecurityProxy {
                 HttpRestResult<String> restResult = nacosRestTemplate
                         .postForm(url, Header.EMPTY, Query.newInstance().initParams(params), bodyMap, String.class);
                 if (!restResult.ok()) {
-                    SECURITY_LOGGER.error("login failed: {}", JacksonUtils.toJson(restResult));
+                    SECURITY_LOGGER.error("login failed: {}", restResult.getMessage());
                     return false;
                 }
                 JsonNode obj = JacksonUtils.toObj(restResult.getData());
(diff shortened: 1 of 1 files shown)


KomachiSion:
I think the beat solution is to find the reason of null key and solve it.

gitdoit:
At the source of ` sun.net.www.MessageHeader` ,you can see

>An RFC 844 or MIME message header.  Includes methods
>for parsing headers from incoming streams, fetching
>values, setting values, and printing headers.
> **Key values of null are legal** : they indicate lines in
>the header that don't have a valid key, but do have
>a value (this isn't legal according to the standard,
>but lines like this are everywhere).

It can't be solved

client/src/main/java/com/alibaba/nacos/client/security/SecurityProxy.java
```diff
@@ -143,7 +143,7 @@ public boolean login(String server) throws UnsupportedEncodingException {
                 HttpRestResult<String> restResult = nacosRestTemplate
                         .postForm(url, Header.EMPTY, Query.newInstance().initParams(params), bodyMap, String.class);
                 if (!restResult.ok()) {
-                    SECURITY_LOGGER.error("login failed: {}", JacksonUtils.toJson(restResult));
+                    SECURITY_LOGGER.error("login failed: {}", restResult.getMessage());
```
KomachiSion, line 146: Should there show username and password ?

gitdoit: 登录失败后显示用户名密码有助于快速定位问题，但是会保存到日志里。或许，可以给一个更友好的提示。另外用户如果发现登录失败，首先肯定会去检查账号密码。

KomachiSion: 这个日志里的用户名密码本身也就是用户配置的，保存在日志里问题应该不大。
至于更明显的提示，对应用来说也还是日志吧。总不能把程序杀掉吧。

KomachiSion: 然后更友好的提示指的是 服务端的优化吗？ 比如用户不存在或者密码不正确？

gitdoit: 首先，感谢你这么耐心的回复。你说的没错，日志文件也是面对用户的，账号密码也同样是。期待nacos越来越完善

PR #3714, opened 2020-08-28
fix #3710

### LINKED ISSUE #3710: Metadata内无法使用特殊字符
使用`nacos-client`**1.3.2**版本时，向Server注册服务时携带的元数据内容：
```java
instance.setMetadata(new HashMap() {{
      put("pipeNames", "1+,2,3");
}});
```
拉取服务时，元数据的内容则把`+`替换成了空格

zzq1314zll:
fix #3710

### DIFF
--- a/common/src/main/java/com/alibaba/nacos/common/http/HttpUtils.java
+++ b/common/src/main/java/com/alibaba/nacos/common/http/HttpUtils.java
@@ -242,6 +242,7 @@ public final class HttpUtils {
             return pre;
         }
         pre = now;
+        now = URLEncoder.encode(now, encode);
         now = URLDecoder.decode(now, encode);
         return innerDecode(pre, now, encode);
     }


common/src/main/java/com/alibaba/nacos/common/http/HttpUtils.java
```diff
@@ -242,6 +242,7 @@ private static String innerDecode(String pre, String now, String encode) throws
             return pre;
         }
         pre = now;
+        now = URLEncoder.encode(now, encode);
```
KomachiSion, line 245: encode before decode? 
it it right?

KomachiSion (changes requested):

common/src/main/java/com/alibaba/nacos/common/http/HttpUtils.java
zzq1314zll: right

KomachiSion:
@zzq1314zll I think the implementation is not right. encode before decode will caused the result is error.

image

PR #3721, opened 2020-08-31
Fix Logging in with the wrong username or password will cause jackson serialization results to fail

### LINKED ISSUE #3695: Logging in with the wrong username or password will cause jackson serialization results to fail

Technoboy-:
Fix #3695

### DIFF
--- a/common/src/main/java/com/alibaba/nacos/common/http/param/Header.java
+++ b/common/src/main/java/com/alibaba/nacos/common/http/param/Header.java
@@ -51,9 +51,18 @@ public class Header {
     public static Header newInstance() {
         return new Header();
     }
-    
+
+    /**
+     * Add the key and value to the header.
+     *
+     * @param key the key
+     * @param value the value
+     * @return header
+     */
     public Header addParam(String key, String value) {
-        header.put(key, value);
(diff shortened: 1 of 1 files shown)


Technoboy-:
cc @KomachiSion

KomachiSion (approved):

PR #3729, opened 2020-09-01
Added: when JAVA_HOME environment variable is not found,trigger

lnsane:

A friendly prompt when the JAVA_HOME environment variable cannot be found

## Brief changelog

XX

## Verifying this change

XXXX

### DIFF
--- a/distribution/bin/startup.sh
+++ b/distribution/bin/startup.sh
@@ -21,9 +21,16 @@ CYGWIN*) cygwin=true;;
 Darwin*) darwin=true;;
 OS400*) os400=true;;
 esac
-error_exit ()
-{
-    echo "ERROR: $1 !!"
+error_exit() {
+    echo "ERROR: $1!!"
+    echo "You can also use this command: "
+    lens=$2
+    printf '%0.s#' $(seq 1 ${#lens})
+    printf "####\n"
+    echo "# $2 #"
+    printf '%0.s#' $(seq 1 ${#lens})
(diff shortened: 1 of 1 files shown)


distribution/bin/startup.sh
```diff
@@ -47,7 +54,12 @@ if [ -z "$JAVA_HOME" ]; then
     fi
   fi
   if [ -z "$JAVA_HOME" ]; then
-        error_exit "Please set the JAVA_HOME variable in your environment, We need java(x64)! jdk8 or later is better!"
+    if [ ! -d "$(dirname $(readlink -f $0))/jdk8u262-b10" ]; then
```
KomachiSion, line 57: hint command locks the version of jdk `jdk8u262-b10` may be not good.

KomachiSion:
@lnsane Please sign the SLA

distribution/bin/startup.sh
lnsane: > hint command locks the version of jdk `jdk8u262-b10` may be not good.

It does not look good, I just have this idea, from my point of view: the download address will not change, so the file name will not change.

KomachiSion: I think the hint is hard to do. 

Download the official JDK8 need oracle account. 
Use github or openJDK need bind the version is not very good.

So I prefer to keep it as now. Tell users nacos config need JAVA_HOME, let users choose there own version and install to users' path.

PR #3872, opened 2020-09-20
[ISSUE #3867] replace the way which get version

horizonzy:
fix-#3867


KomachiSion:
There are other application.properties in `nacos-api` module.

KomachiSion (approved):

Issue #3850, opened 2020-09-17
项目停止是，不要输出java.net.SocketException: socket closed这样的错误


jasonpyf:
![11](https://user-images.githubusercontent.com/16953632/93411553-fd4eb780-f8cd-11ea-9af3-de980f04af1a.png)

KomachiSion:
@jasonpyf which version you used?

yueyemisi:
nacos server version:1.3.2
nacos client version:1.3.2

2020-09-18 11:50:22.156  INFO c.a.n.c.naming            149-[]: com.alibaba.nacos.client.naming.beat.BeatReactor do shutdown stop
2020-09-18 11:50:22.157  INFO c.a.n.c.naming            149-[]: com.alibaba.nacos.client.naming.core.EventDispatcher do shutdown begin
2020-09-18 11:50:25.160  INFO c.a.n.c.naming            152-[]: com.alibaba.nacos.client.naming.core.EventDispatcher do shutdown stop
2020-09-18 11:50:25.162  INFO c.a.n.c.naming            370-[]: com.alibaba.nacos.client.naming.core.HostReactor do shutdown begin
2020-09-18 11:50:27.550  INFO c.a.n.c.naming            118-[]: com.alibaba.nacos.client.naming.core.PushReceiver do shutdown begin
2020-09-18 11:50:30.553  INFO c.a.n.c.naming            122-[]: com.alibaba.nacos.client.naming.core.PushReceiver do shutdown stop
2020-09-18 11:50:30.555  INFO c.a.n.c.naming            132-[]: com.alibaba.nacos.client.naming.backups.FailoverReactor do shutdown begin
2020-09-18 11:50:30.558  INFO c.a.n.c.naming            134-[]: com.alibaba.nacos.client.naming.backups.FailoverReactor do shutdown stop
2020-09-18 11:50:30.558  INFO c.a.n.c.naming            374-[]: com.alibaba.nacos.client.naming.core.HostReactor do shutdown stop
2020-09-18 11:50:30.558  INFO c.a.n.c.naming            715-[]: com.alibaba.nacos.client.naming.net.NamingProxy do shutdown begin
2020-09-18 11:50:30.559  WARN c.a.n.c.naming            72-[]: [NamingHttpClientManager] Start destroying NacosRestTemplate
2020-09-18 11:50:30.560  WARN c.a.n.c.naming            79-[]: [NamingHttpClientManager] Destruction of the end
[...]

PR #3906, opened 2020-09-24
[ISSUE #3850] ignore socket exception when client destroy already

horizonzy:
fix-#3850

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/naming/core/PushReceiver.java
+++ b/client/src/main/java/com/alibaba/nacos/client/naming/core/PushReceiver.java
@@ -25,6 +25,7 @@ import com.alibaba.nacos.common.utils.ThreadUtils;
 
 import java.net.DatagramPacket;
 import java.net.DatagramSocket;
+import java.net.SocketException;
 import java.nio.charset.Charset;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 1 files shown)


KomachiSion (changes requested):
I think check `updSocket.isClose` before receive or send will be better.

If `updSocket.isClose` is true, directly return;

also check `close` in pushReceiver is ok.

horizonzy:
> I think check `updSocket.isClose` before receive or send will be better.
> 
> If `updSocket.isClose` is true, directly return;
> 
> also check `close` in pushReceiver is ok.

check updSocket.isClose before receive is unuseful, because it blocked. the judge will not work. before send check is necessary

KomachiSion:
> > I think check `updSocket.isClose` before receive or send will be better.
> > If `updSocket.isClose` is true, directly return;
> > also check `close` in pushReceiver is ok.
> 
> check `updSocket.isClose` before receive is unuseful most of the time, if `udpSocket.receive(packet);` already execute, the exception will occur still, so the judge logic in catch code block is necessary. if this code exist, it will cover the check situation

You are right, from this point, I think there are is no need to judgement the exception type. Because if closed means the PushReceiver has been shutdown, Any exception can be ignored.

KomachiSion (approved):

PR #2662, opened 2020-04-21
[ISSUE #2660]  Add the type parameter when call publishConfig(), and support Aliyun ACM

flyer5200:
## What is the purpose of the change

Fixes #2660 

## Brief changelog

See ISSUE #2660 

1. Add parammeter check and support Aliyun ACM
2. Add the type parameter when the [com.alibaba.nacos.client.config.NacosConfigService#publishConfig](https://github.com/alibaba/nacos/blob/4e52ff050abb65a7cb14a09a2d32641ef3505aa0/client/src/main/java/com/alibaba/nacos/client/config/NacosConfigService.java#L110) method is executed so that the naco or Aliyun ACM dashboard can be displayed correctly

## Verifying this change

Follow this checklist to help us incorporate your contribution quickly and easily:

* [ ] Make sure there is a Github issue filed for the change (usually before you start working on it). Trivial changes like typos do not require a Github issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.
* [ ] Format the pull request title like `[ISSUE #123] Fix UnknownException when host config not exist`. Each commit in the pull request should have a meaningful subject line and body.
* [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
* [ ] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/alibaba/nacos/tree/master/test).
* [ ] Run `mvn -B clean package apache-rat:check findbugs:findbugs -Dmaven.test.skip=true` to make sure basic checks pass. Run `mvn clean install -DskipITs` to make sure unit-test pass. Run `mvn clean test-compile failsafe:integration-test`  to make sure integration-test pass.

PR #3915, opened 2020-09-25
[ISSUE #3907] when publishConfig, if type is absent, set default value 'text'

horizonzy:
fix-#3907


KomachiSion:
I think it can't fix the issue, it will cause the type be replaces with text.

horizonzy:
> I think it can't fix the issue, it will cause the type be replaces with text.

It just replaced when type not setted. when type not setted, we set default value as 'text' or throw exception?

KomachiSion:
I think the solution should be to modifying config-client, get the type by dataId suffix, if not suffix or unsupported suffix, use text as default.

horizonzy:
@KomachiSion @chuntaojun 
#2662 is no response, use this pr to fix it.

KomachiSion (approved):

Issue #3976, opened 2020-10-12
Beatinfo in client will not update when client didn't call com.alibaba.nacos.client.naming.core.HostReactor#getServiceInfo


horizonzy:
**Describe the bug**
in #2800. we will update beatinfo when instance changed. but if we not invoke `com.alibaba.nacos.client.naming.core.HostReactor#getServiceInfo`. it can't be updated. the problem in #2800 will exist still.

**Expected behavior**
when register instance, call `hostReactor
                        .getServiceInfo(NamingUtils.getGroupedName(serviceName, groupName), StringUtils.join(clusters, ",")`

KomachiSion:
We need to discuss this feature, From the original design, the information of instance is based on client register and beat report. When app restarted, the information will be reset.

It seems same as eureka and consul.

And I think add this auto-fresh will cost much resources for client.

chuntaojun:
> We need to discuss this feature, From the original design, the information of instance is based on client register and beat report. When app restarted, the information will be reset.
> 
> It seems same as eureka and consul.
> 
> And I think add this auto-fresh will cost much resources for client.

As far as I am concerned, the heartbeat is just a live operation, and the metadata information and weight should not be modified. Service: Group: Cluster: IP: port should only be synchronized

Metadata information, weights, and so on should be handled by additional interfaces

horizonzy:
> > We need to discuss this feature, From the original design, the information of instance is based on client register and beat report. When app restarted, the information will be reset.
> > It seems same as eureka and consul.
> > And I think add this auto-fresh will cost much resources for client.
> 
> As far as I am concerned, the heartbeat is just a live operation, and the metadata information and weight should not be modified. Service: Group: Cluster: IP: port should only be synchronized
> 
> Metadata information, weights, and so on should be handled by additional interfaces

Now the beatinfo will updated if call com.alibaba.nacos.client.naming.core.HostReactor#getServiceInfo. com.alibaba.nacos.client.naming.core.HostReactor#getServiceInfo will call in com.alibaba.nacos.client.naming.NacosNamingService#getAllInstances or com.alibaba.nacos.client.naming.NacosNamingService#subscribe。if user not call one of both. the beatInfo will not update.

PR #3977, opened 2020-10-12
[ISSUE-#3976] Add instance changed listener when register ephemeral instance

horizonzy:
fix-#3976


KomachiSion:
Close PR beacuse of closing issue.

PR #4010, opened 2020-10-16
For #3710, Fix special characters cannot be used in Metadata problem

zilongTong:
**purpose of the change** 
For #3710 

**Brief changelog**
 Fix special characters cannot be used in Metadata problem

**Verifying this change**

- nacos-client uses special characters to register nacos verified

- openAPI uses special characters to register nacos verified

- Verified in the official environment


KomachiSion:
@zilongTong two question
1. your submit is not your github id, so CLA may not pass. Please bind your github id to your git and submit them again.
2. your changes remove decoding totally, I think it might cause so error. We need do more discussion.

Issue #4003, opened 2020-10-16
[Discuss] module dependency relation. Should common-module should relay on api-module ?


horizonzy:
### Describe what happened (or what feature you want)
Now common-module relay on api-module it relay on api-module class such as (`com.alibaba.nacos.api.exception.NacosException`, `com.alibaba.nacos.api.common.Constants`),if other module relay on common-module. it will relay on common-module tohether. 
In my opinion, common-module should not relay on other module. Now api-module some class such as (`com.alibaba.nacos.api.utils.StringUtils`, `com.alibaba.nacos.api.common.Constants`). I think these class should be defined in common-module. api-module should relay on common-module.

KomachiSion:
Yes, I also think modules dependency need to be enhance.

horizonzy:
I will have a try to enhance it.

KomachiSion:
I suggest to discuss it first before do it.

horizonzy:
that's fine

Issue #3102, opened 2020-06-18
Remove apache commons-lang3 dependency in Server.


KomachiSion:
###  Issue Description

Type: *Remove dependency*

### Describe what happened (or what feature you want)

Nacos wants to reduce the dependency to reduce size of server. And the community volunteers wants to remove commons-lang3 first. So I create an issue and split tasks for community.

### Describe what you expected to happen

Replace commons-lang3 dependency module by module. And the behavior of server do not change.

### How to reproduce it (as minimally and precisely as possible)

- [x] Create utils class to nacos-common or nacos-api module.
- [x] Replace commons-lang3 in nacos-common module.
- [x] Replace commons-lang3 in nacos-core module.
- [x] Replace commons-lang3 in nacos-config module.
- [x] Replace commons-lang3 in nacos-naming module.
- [x] Replace commons-lang3 in nacos-console module.
- [x] Replace commons-lang3 in other modules(if necessary, will be spiltted again).
- [x] remove commons-lang3 pom dependency.

Welcome volunteer join us.

zongtanghu:
[] Create utils class to nacos-common or nacos-core module.
[] Replace commons-lang3 in nacos-common module.
[] Replace commons-lang3 in nacos-core module.
[] Replace commons-lang3 in nacos-config module.
[] Replace commons-lang3 in nacos-naming module.
[] Replace commons-lang3 in nacos-console module.
[] Replace commons-lang3 in other modules(if necessary, will be spiltted again).
[] remove commons-lang3 pom dependency.

Then, we claim this feature issue? @KomachiSion

KomachiSion:
@zongtanghu you can create an issue according these unfinished task. and refer to this issue.

Of course, you can also directly open PR to finished task one by one.

wangweizZZ:
@lin-mt @zongtanghu  hi，is this part of work still in progress? Maybe @corey89757  can help solve this problem together :-)

shizhengxing:
I‘d like to work this out together

corey89757:
I am working on it.

corey89757:
I'll split the pr and optimise some details.

corey89757:
Since now, there isn't any lang dependency in common/api/client module.

li-xiao-shuang:
Has the problem been dealt with yet

PR #4016, opened 2020-10-18
[ISSUE-#4003] Refactor: change module dependency relationship. 

horizonzy:
#4003 
#3102 

## Brief changelog
1.common-module relay on api-module --> api-module relay on common-mo
2.remove common-lang3 dependency in api-module. 
ps1:`org.apache.commons.lang3.BooleanUtils#toBoolean(java.lang.String)` not transfer to `com.alibaba.nacos.common.utils.BooleanUtils`, just use Boolean.parse(String xxx) in project)
ps2: `org.apache.commons.lang3.ArrayUtils#toString(java.lang.Object)` not transfer to `com.alibaba.nacos.common.utils.ArrayUtils`, the use position just one place, so just replace it with this code `            String key = service.getName() + ":" + (service.getOwners() == null ? "[]" : service.getOwners().toString());
`
3.move some class from api-module to common-module


horizonzy:
This is the first commit. The Constants will be categorized in two days.

KomachiSion:
Changing module dependency relationship is a big issue. I prefer pending this PR first, and do it in 1.4.1.

horizonzy:
> Changing module dependency relationship is a big issue. I prefer pending this PR first, and do it in 1.4.1.

I think so.

Issue #4056, opened 2020-10-25
NamingModule can't supply service normally with standalone mode


horizonzy:
**Describe the bug**
At the newest code of develop branch and 1.4-BETA, if run nacos server by standalone mode, in `com.alibaba.nacos.naming.web.TrafficReviseFilter`, `serverStatusManager.getServerStatus()` will always Down, it lead namingModule can't supply services.

horizonzy:
Finnally, I found the reson. The jraftConsistencyServer will save log, I have start nacos server when my pc IP is `10.12.217.94`. When I occur this issue, my pc IP is `10.12.217.120`. In the progress of jraftServer starting, it will load log from embedded rocksDB. the log save IP is `10.12.217.94`. So the jraftServer can't elect a leader, naming module can't supply services.

horizonzy:
When I check the reason, I found that `com.alibaba.nacos.core.distributed.distro.task.load.DistroLoadDataTask` exist dead circle when standalone mode, it should be enhenced.

PR #4057, opened 2020-10-25
[ISSUE-#4056] Fix naming module can't supply services when standalone mode

horizonzy:

fix-#4056

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/consistency/persistent/PersistentConsistencyServiceDelegateImpl.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/consistency/persistent/PersistentConsistencyServiceDelegateImpl.java
@@ -22,6 +22,7 @@ import com.alibaba.nacos.naming.consistency.RecordListener;
 import com.alibaba.nacos.naming.consistency.persistent.impl.PersistentServiceProcessor;
 import com.alibaba.nacos.naming.consistency.persistent.raft.RaftConsistencyServiceImpl;
 import com.alibaba.nacos.naming.pojo.Record;
+import com.alibaba.nacos.sys.utils.ApplicationUtils;
 import org.springframework.stereotype.Component;
 
 /**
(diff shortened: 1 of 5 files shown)


naming/src/main/java/com/alibaba/nacos/naming/consistency/persistent/PersistentConsistencyServiceDelegateImpl.java
```diff
@@ -86,6 +87,9 @@ public boolean isAvailable() {
     }
     
     private PersistentConsistencyService switchOne() {
+        if (ApplicationUtils.getStandaloneMode()) {
+            return oldPersistentConsistencyService;
```
KomachiSion, line 91: If standalone, should we use new consistency service?
And can we make switchNewPersistentService = true rather than directly return oldPersistentConsistencyService

naming/src/main/java/com/alibaba/nacos/naming/consistency/persistent/impl/PersistentServiceProcessor.java
```diff
@@ -147,6 +147,10 @@ public PersistentServiceProcessor(final ProtocolManager protocolManager,
     private void init() {
         NotifyCenter.registerToPublisher(ValueChangeEvent.class, 16384);
         this.protocol.addLogProcessors(Collections.singletonList(this));
+        if (ApplicationUtils.getStandaloneMode()) {
```
KomachiSion, line 150: the two raft impl behavior in standalone need to be discuss.

KomachiSion (changes requested):

naming/src/main/java/com/alibaba/nacos/naming/consistency/persistent/PersistentConsistencyServiceDelegateImpl.java
chuntaojun: If it is in this form, then the purpose of replacing the old raft before does not exist, this should be optimized in the new PersistentService

PR #4067, opened 2020-10-26
[ISSUE-#4066] Let the service's order followed the order which it be added.

horizonzy:

Fix-#4066

### DIFF
--- a/istio/src/main/java/com/alibaba/nacos/istio/mcp/NacosMcpService.java
+++ b/istio/src/main/java/com/alibaba/nacos/istio/mcp/NacosMcpService.java
@@ -37,6 +37,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 
 import javax.annotation.PostConstruct;
+import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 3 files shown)


KomachiSion:
Will `Collections.synchronizedMap` reduce performance than the `ConcurrentHashMap`?

horizonzy:
> Will `Collections.synchronizedMap` reduce performance than the `ConcurrentHashMap`?

It will, it's exclusive lock. I will try another way to realize it, just pending it

horizonzy:
@chuntaojun @KomachiSion

istio/src/main/java/com/alibaba/nacos/istio/mcp/NacosMcpService.java
```diff
@@ -94,14 +95,14 @@ public void run() {
             Set<String> namespaces = serviceManager.getAllNamespaces();
             
             for (String namespace : namespaces) {
-                
-                Map<String, Service> services = serviceManager.chooseServiceMap(namespace);
-                
-                if (services.isEmpty()) {
+    
+                Collection<Service> values = serviceManager.chooseServiceMap(namespace).values();
```
chuntaojun, line 99: Here directly use forEach to traverse, don't use values()

naming/src/main/java/com/alibaba/nacos/naming/core/ServiceManager.java
```diff
@@ -1135,6 +1140,49 @@ public void run() {
         }
     }
     
+    private static class ServiceNameHelper implements Comparable<ServiceNameHelper> {
```
chuntaojun, line 1143: The meaning of this class? I feel that with this design, I am very puzzled. Is there a Map inside ServiceNameHelper?

horizonzy:
This pr can't ensure the global order in cluster.

PR #4103, opened 2020-10-30
when publishconfig, get the type through the dataid suffix, if there is no suffix or unsupported suffix, use the text as the default value

### LINKED ISSUE #3907: 调用 publishConfig 会覆盖数据库 type 字段
版本：Spring Boot 2.3.3，Nacos Client 1.3.0，Nacos Server 1.3.2

代码 NacosConfigService 类 publishConfigInner 第 252 行 result = agent.httpPost(url, headers, params, encode, POST_TIMEOUT);

params 参数如下：

image

缺少 type 参数，请求后会覆盖数据库原本的 type，导致 type 为 null。

Nacos 控制台：

image

以下是官方接口参数：

image

直接调用接口，忽略 type 参数和上面的结果相同。

image

ChasingInTheWind:
Fix #3907

XXXXX

## Brief changelog

XX

## Verifying this change

XXXX

### DIFF
--- a/config/src/main/java/com/alibaba/nacos/config/server/controller/ConfigController.java
+++ b/config/src/main/java/com/alibaba/nacos/config/server/controller/ConfigController.java
@@ -16,6 +16,7 @@
 
 package com.alibaba.nacos.config.server.controller;
 
+import com.alibaba.nacos.api.config.ConfigType;
 import com.alibaba.nacos.api.exception.NacosException;
 import com.alibaba.nacos.auth.annotation.Secured;
 import com.alibaba.nacos.auth.common.ActionTypes;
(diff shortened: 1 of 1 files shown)


horizonzy:
I think this logic which handle suffix is the client logic. In server, if you want to judge the text is valid, just check the type is 'Text' or other type which announce in server.

KomachiSion:
> I think this logic which handle suffix is the client logic. In server, if you want to judge the text is valid, just check the type is 'Text' or other type which announce in server.

If we use old client but new server, it will also is null or always text? I think client need contain parse type logic and send type to server, and server also need this logic if no type receivce.

BTW, the parsing logic can be extract to a util to common, client and server both use it.

What do you think?

horizonzy:
> > I think this logic which handle suffix is the client logic. In server, if you want to judge the text is valid, just check the type is 'Text' or other type which announce in server.
> 
> If we use old client but new server, it will also is null or always text? I think client need contain parse type logic and send type to server, and server also need this logic if no type receivce.
> 
> BTW, the parsing logic can be extract to a util to common, client and server both use it.
> 
> What do you think?

I prefer think suffix is business, it should be realized by third integrate framework, such as spring-nacos-context, nacos just save the type which is send from client. In server, we just check the type is valid (blank or is not announced in server). if invalid, set a default value 'text' to the type.

horizonzy:
Now in nacosConfigService, we cat't set type when publishConfig, maybe we can supply a way let user set type when publicConfig. I am not prefer parse type by suffix, If we do, almost dataId will contains suffiix, even the content isn't match suffix. And some user still don't set data with suffix, in server, we still to handle that case.

ChasingInTheWind:
> > I think this logic which handle suffix is the client logic. In server, if you want to judge the text is valid, just check the type is 'Text' or other type which announce in server.
> 
> If we use old client but new server, it will also is null or always text? I think client need contain parse type logic and send type to server, and server also need this logic if no type receivce.
> 
> BTW, the parsing logic can be extract to a util to common, client and server both use it.
> 
> What do you think?

I agree with horizonzy view that the service should only verify the validity of the data and the default value

KomachiSion:
If server only verify type, this PR should be close, because this implementation is parsing suffix and set it as type.

KomachiSion:
Refer to #3915

ChasingInTheWind:
> Refer to #3915

ok

Issue #4134, opened 2020-11-04
Check namespaceId is exist when some api need namespace to handle


horizonzy:
**Describe the bug**
Now I register a instance with the not exist namespaceId. In server, `com.alibaba.nacos.naming.core.ServiceManager` will create a new namesapce, and put service in it.
For this situation, we should check the namespaceId is exist. If the situation is not exist, throw an exception.

KomachiSion:
If need to check whether namespace is exist, it will depend persistency module first.

horizonzy:
Maybe we can add a filter in console module to fix it

KomachiSion:
And need to cache namespace in memory, and need to maintain the cache.

CaiJiHuzhuo:
I noticed in the nacos documentation that the namespace module will be moved to the public module. which module? common?

horizonzy:
I think move it to core module will be better, like authFilter.

JackSun-Developer:
public class NamespaceFilter implements Filter {
    Set<String> existingNamespaces;
    ...
    @Override
    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) {
        String namespace = request.getNamespace();
        if (!existingNamespaces.contains(namespace)) {
             existingNamespaces.replace(namespace, "public");
       }
   }
}

JackSun-Developer:
Can this ISSUE be solved by simply adding a NamespaceFilter? What other problems should be take cared?

PR #4161, opened 2020-11-06
[ISSUE-#4134] check namespace is exist when request naming api or config api

horizonzy:
issue-#4134


console/src/main/java/com/alibaba/nacos/console/controller/NamespaceController.java
```diff
@@ -137,6 +141,7 @@ public Boolean createNamespace(HttpServletRequest request, HttpServletResponse r
         }
         persistService.insertTenantInfoAtomic("1", namespaceId, namespaceName, namespaceDesc, "nacos",
                 System.currentTimeMillis());
+        namespaceService.addTenantId(namespaceId);
```
KomachiSion, line 144: replace method name by addNamespaceId?

KomachiSion (changes requested):
We may replace all tenantId with namespaceId in future. So I think use namespaceId as method name is better.

CaiJiHuzhuo:
move console/NamespaceController.java to naming module, then config module adds two api(add,delete). when namespace change，use Http send namespaceId update data to config module.
The namespaceId is stored in both the name and config modules.
check namespaceId exists in memory.

horizonzy:
> move console/NamespaceController.java to naming module, then config module adds two api(add,delete). when namespace change，use Http send namespaceId update data to config module.
> The namespaceId is stored in both the name and config modules.
> check namespaceId exists in memory.

Why move NamespaceController to naming module and config module? Run one module alonely?

horizonzy:
This feature will be moved to 2.x, so close it.

Issue #4166, opened 2020-11-07
[Code Quality] NotifyCenter code optimization


horizonzy:
**Describe the target**
optimize notityCenter code

PR #4167, opened 2020-11-07
[ISSUE-#4166] Optimize notifyCenter code

horizonzy:
For-#4166

### DIFF
--- a/common/src/main/java/com/alibaba/nacos/common/notify/NotifyCenter.java
+++ b/common/src/main/java/com/alibaba/nacos/common/notify/NotifyCenter.java
@@ -18,8 +18,8 @@ package com.alibaba.nacos.common.notify;
 
 import com.alibaba.nacos.api.exception.runtime.NacosRuntimeException;
 import com.alibaba.nacos.common.JustForTest;
-import com.alibaba.nacos.common.notify.listener.Subscriber;
 import com.alibaba.nacos.common.notify.listener.SmartSubscriber;
+import com.alibaba.nacos.common.notify.listener.Subscriber;
 import com.alibaba.nacos.common.utils.BiFunction;
 import com.alibaba.nacos.common.utils.ClassUtils;
 import com.alibaba.nacos.common.utils.MapUtils;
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/alibaba/nacos/common/notify/NotifyCenter.java
```diff
@@ -250,13 +250,12 @@ private static void addSubscriber(final Subscriber consumer, Class<? extends Eve
     private static boolean removeSubscriber(final Subscriber consumer, Class<? extends Event> subscribeType) {
         
         final String topic = ClassUtils.getCanonicalName(subscribeType);
-        if (INSTANCE.publisherMap.containsKey(topic)) {
-            EventPublisher publisher = INSTANCE.publisherMap.get(topic);
-            publisher.removeSubscriber(consumer);
-            return true;
+        EventPublisher eventPublisher = INSTANCE.publisherMap.get(topic);
```
KomachiSion, line 253: I think the old implementation is better for readablilty

common/src/main/java/com/alibaba/nacos/common/notify/NotifyCenter.java
```diff
@@ -281,18 +280,19 @@ public static boolean publishEvent(final Event event) {
...
             return INSTANCE.sharePublisher.publish(event);
         }
         
-        if (INSTANCE.publisherMap.containsKey(topic)) {
-            EventPublisher publisher = INSTANCE.publisherMap.get(topic);
-            return publisher.publish(event);
+        final String topic = ClassUtils.getCanonicalName(eventType);
+        
+        EventPublisher publisher = INSTANCE.publisherMap.get(topic);
+        if (publisher == null) {
```
KomachiSion, line 290: Same as above

KomachiSion (changes requested):

common/src/main/java/com/alibaba/nacos/common/notify/NotifyCenter.java
horizonzy: I think containsKey() and get() is duplicated. get() can cover containsKey(); Maybe should judge eventPublish != null?

KomachiSion: OK

KomachiSion (approved):

Issue #4136, opened 2020-11-04
there is a bug in file ServiceInfo


shizhengxing:
java class name:com.alibaba.nacos.api.naming.pojo.ServiceInfo  
method:

    public ServiceInfo(String key) {  
        int maxIndex = 2;  
        int clusterIndex = 1;
        int serviceNameIndex = 0;

        String[] keys = key.split(Constants.SERVICE_INFO_SPLITER);
        if (keys.length >= maxIndex) {
            this.name = keys[serviceNameIndex];
            this.clusters = keys[clusterIndex];
        }

        this.name = keys[0];
    }`
When read cache file from disk throuth method `DiskCache.read(String cacheDir)`, the method use constructor `public ServiceInfo(String key)` to instantiate `ServiceInfo`,you can check the code in file DiskCache of line 111.
I give a example of the input param for method `public ServiceInfo(String key)` like `group@@service_name@@clusters`.

You can see when spilt with `@@` ,the `clusterIndex` should be 2 not 1, and the serviceNameIndex should be 1 not 0,and the `groupName` can be fill with data of index=0.

shizhengxing:
if it confirmed ,i will fix it so quickly!

KomachiSion:
OK.

When you try to fix it. Please do some test for compatibility test with old version nacos server from 1.1.3 to 1.3.2

zrlw:
> java class name:com.alibaba.nacos.api.naming.pojo.ServiceInfo
> method:
> 
> ```
> public ServiceInfo(String key) {  
>     int maxIndex = 2;  
>     int clusterIndex = 1;
>     int serviceNameIndex = 0;
> 
>     String[] keys = key.split(Constants.SERVICE_INFO_SPLITER);
>     if (keys.length >= maxIndex) {
>         this.name = keys[serviceNameIndex];
>         this.clusters = keys[clusterIndex];
>     }
> 
>     this.name = keys[0];
> }`
> ```
> 
> When read cache file from disk throuth method `DiskCache.read(String cacheDir)`, the method use constructor `public ServiceInfo(String key)` to instantiate `ServiceInfo`,you can check the code in file DiskCache of line 111.
> I give a example of the input param for method `public ServiceInfo(String key)` like `group@@service_name@@clusters`.
> 
> You can see when spilt with `@@` ,the `clusterIndex` should be 2 not 1, and the serviceNameIndex should be 1 not 0,and the `groupName` can be fill with data of index=0.

[...]

PR #4183, opened 2020-11-09
[ISSUE #4136] Fix the bug that cann't correctlly instantiate ServiceInfo from cache file

shizhengxing:
#4136 

[ISSUE #4136] Fix the bug that cann't correctlly instantiate ServiceInfo from cache file


api/src/main/java/com/alibaba/nacos/api/naming/pojo/ServiceInfo.java
```diff
@@ -67,20 +68,24 @@ public void setAllIPs(boolean allIPs) {
...
+				int clusterIndex = 2;
+				int serviceNameIndex = 1;
+				int groupIndex = 0;
+			
+				String[] keys = key.split(Constants.SERVICE_INFO_SPLITER);
+				if (keys.length >= maxIndex + 1) {
+					this.groupName = keys[groupIndex];
+					this.name = keys[serviceNameIndex];
+					this.clusters = keys[clusterIndex];
+				} else if (keys.length == maxIndex) {
```
KomachiSion, line 82: If key is `serviceName@@cluster`, whether the condition is error?

From your `getKey`, if groupName is null or '' and serviceName without group, getKey will become `serviceName@@cluster`

shizhengxing: the `groupName` will not be real empty! when it be null，it must has be concatted to `serviceName` like `DEFAULT_GROUP@@serviceName`. And you must know that the fact is `groupName` never be null in server side in naming model.

shizhengxing: And you can check the code of method `ServiceInfo.fromKey`  
>`	public static ServiceInfo fromKey(String key) {
		ServiceInfo serviceInfo = new ServiceInfo();
		int maxSegCount = 3;
		String[] segs = key.split(Constants.SERVICE_INFO_SPLITER);
		if (segs.length == maxSegCount - 1) {
			serviceInfo.setGroupName(segs[0]);
			serviceInfo.setName(segs[1]);
		} else if (segs.length == maxSegCount) {
			serviceInfo.setGroupName(segs[0]);
			serviceInfo.setName(segs[1]);
			serviceInfo.setClusters(segs[2]);
		}
		return serviceInfo;
	}`     
if the `groupName` can be real null，this method also be a big bug，the user API of `NamingService.getSubscribeServices()` depends on it!

KomachiSion: The SeriviceInfo can be created by three parts:
1. constructor
2. factory method `fromKey`
3. Jackson deserialization

If old version server push an ServiceInfo without group. And it use `getKey` method, the key will be serviceName@@cluster.

shizhengxing: 1)constructor：I hava checked the three constructors's use place，it compatible with my changing.
2)factory method `fromKey`: Also compatibly，itself will be a big bug if the key can be like `serviceName@@clusters`，the user API of NamingService.getSubscribeServices() depends on it.
3)Jackson deserialization: Come from server side push , I have checked naming module from version 1.1.3 to newest version，the `name` is directly return as it send by client， and if the `groupName` be null in a abnormal situation , the `com.alibaba.nacos.naming.web.DistroFilter` will fill with `DEFAULT_GROUP`. You can check the code in file `DistroFilter`.

In short, I think that my changes is compatible enough , and it works well in my company,if you has better solution, I will learn it。

shizhengxing: I hope you can tell me which version of in the server side or client can real lead to an incompatible situation

KomachiSion: I just put forward this possible problem in this PR, and the mainly usage version for now is 1.1.3 to 1.3.2, I want to discuss your changes and make sure it can be compatible with these old version.

What's more, you need to check the CI first, I found it failed all.

api/src/main/java/com/alibaba/nacos/api/naming/pojo/ServiceInfo.java
```diff
@@ -197,11 +206,16 @@ public static String getKey(String name, String clusters) {
...
     public String getKeyEncoded() {
+        String serviceName;
         try {
-            return getKey(URLEncoder.encode(name, "UTF-8"), clusters);
+            serviceName = URLEncoder.encode(this.name, "UTF-8");
         } catch (UnsupportedEncodingException e) {
-            return getKey();
+            serviceName = this.name;
+        }
+        if (!StringUtils.isBlank(groupName) && this.name.indexOf(Constants.SERVICE_INFO_SPLITER) == -1) {
```
KomachiSion, line 215: this part codes is duplicate, it can be extract as an method in this class.

KomachiSion (changes requested):

api/src/main/java/com/alibaba/nacos/api/naming/pojo/ServiceInfo.java
chuntaojun: Agree, defensive programming is needed here, and don't dismiss it just because the current code doesn't exist

api/src/main/java/com/alibaba/nacos/api/naming/pojo/ServiceInfo.java
```diff
@@ -68,18 +69,22 @@ public void setAllIPs(boolean allIPs) {
     }
     
     public ServiceInfo(String key) {
```
chuntaojun, line 71: I think it's better to comment on each case here

chuntaojun:
please add some unit test to verify that your changes are correct

api/src/main/java/com/alibaba/nacos/api/naming/pojo/ServiceInfo.java
shizhengxing: > I think it's better to comment on each case here

OK, I wll add it

api/src/main/java/com/alibaba/nacos/api/naming/pojo/ServiceInfo.java
shizhengxing: > this part codes is duplicate, it can be extract as an method in this class.

OK,I will change it

shizhengxing:
> please add some unit test to verify that your changes are correct

OK,I will add some

api/src/main/java/com/alibaba/nacos/api/naming/pojo/ServiceInfo.java
shizhengxing: > Agree, defensive programming is needed here, and don't dismiss it just because the current code doesn't exist

What I can do is to set some clear rules when and how to generate the key used by `ServiceInfo`.

1. rule 1: groupName must not be null , if you donn't have a `groupName`,you must specify as `DEFAULT_GROUP`.
2. rule 2: only `clusters` can be null,so when there is a string like `XXX@@YYY`,the `XXX` must be `groupName` and the `YYY` must be `name`.Because if there is a string like `name@@clusters`, you have no idea to recognize it.

Then hope everybody observe the ruls, and donn't makes mistake here.And  I think the `defensive programming` here is throw an exception when somebody want to generate a `ServiceInfo` key or instantiate a `ServiceInfo` instance with null `groupName`.If you have more idea to share,I will be thankful ! 

Than hope anybody donn't makes mistake here,

shizhengxing:
修改提交后发现变动了几百个文件，关闭此PR，重新提交PR

PR #4186, opened 2020-11-10
[ISSUE-#4179] Add index for db.user and db.password

horizonzy:
For-#4179, add index for db.user and db.password, to avoid some situation that it contains symbol `,`.


KomachiSion (approved):

chuntaojun:
if db.user and db.password not index info, it mean is global user info, so I think you need to describe the two cases separately

KomachiSion:
Should we describe this issue in document?

If add all item in default application.properties, it looks too confused.

chuntaojun:
> Should we describe this issue in document?
> 
> If add all item in default application.properties, it looks too confused.

agree，wo need to describe this issue in document

horizonzy:
Agreen, I will add some description to doc later.

horizonzy:
Default config with index will be better I think.

KomachiSion (approved):

PR #4240, opened 2020-11-16
[ISSUES 4241] fix:  maybe NPE & remove useless code

fattyca1:
1. fix maybe NPE
2. remove useless code
issue-https://github.com/alibaba/nacos/issues/4241


chuntaojun:
you need to format code style and create issue to describe why you need to create this pr to fix problem

KomachiSion:
This part codes can't be removed for now, The reason I have commented in issue.

We will remove the produce support in client and make it plugin in future.

KomachiSion:
@fattyca1 Thanks again for your contribution. But this Pr and issue can't be accept for now. So sorry.

Issue #4264, opened 2020-11-19
Com.alibaba.nacos.sys.env.EnvUtil#getContextPath will always load property when the contextPath is '/'


horizonzy:
**Describe the bug**
com.alibaba.nacos.sys.env.EnvUtil#getContextPath will always load property when the contextPath is '/'

**Expected behavior**
just load property once

PR #4265, opened 2020-11-19
[ISSUE-#4264] Fix always load property when the contextPath is '/'

horizonzy:

For-#4264


chuntaojun (approved):

wangweizZZ:
This pr will cause the first call to return "" and then return "/" ?

horizonzy:
> This pr will cause the first call to return "" and then return "/"

yep, next commit.

PR #4271, opened 2020-11-19
Fix twice invoke and the value is not same.

horizonzy:
For-#4265


horizonzy:
@chuntaojun @wangweizZZ @sanxun0325 
Fix @wangweizZZ memtion in #4265

KomachiSion (approved):

wangweizZZ (approved):

Issue #4262, opened 2020-11-19
Nacos permission Bug


309283175:
turn on auth system, create new role ROLE_TEST,then add public resource to ROLE_TEST with rw,the data in table permissions  is:

```
role	       resource	action
ROLE_TEST	:*:*	rw
```

com.alibaba.nacos.console.security.nacos.roles.NacosRoleServiceImpl.hasPermission()
```
 String permissionResource = permissionInfo.getResource().replaceAll("\\*", ".*");
 String permissionAction = permissionInfo.getAction();
  if (permissionAction.contains(permission.getAction()) &&
       Pattern.matches(permissionResource, permission.getResource())) {
        return true;
 }
```

if i use spring.cloud.nacos.username=test register my service to nacos with public namespace,my resource is "public:DEFAULT_GROUP:naming/my-service",but the parttern is ```:.*:.*``` , **Pattern.matches(permissionResource, permission.getResource())** shoud be **true** but **false**

com.alibaba.nacos.api.exception.NacosException: failed to req API:/nacos/v1/ns/instance after all servers([x.x.x.x:8848]) tried: ErrCode:403, ErrMsg: authorization failed!

horizonzy:
Hi, did you bind user `test` with role `ROLE_TEST`.

309283175:
@horizonzy  of course i bind user `test` with role `ROLE_TEST`.

309283175:
@horizonzy ,As i mentioned above ,this bug only appear with user not have `ROLE_ADMIN` and want register to public namespace.

horizonzy:
I will have a try to reproduce it.

sanxun0325:
when public resource permissions are set, the permissionResource should be `public:.*:.*`

309283175:
@sanxun0325  I have tried to modify the resource record of the database `:*:*` to `public:*:*` directly. The service registration is successful. but it has caused another problem when I log in to the console with the test user and select the public namespace, a warning of no permission will appear。

sanxun0325:
> @sanxun0325 I have tried to modify the resource record of the database `:*:*` to `public:*:*` directly. The service registration is successful. but it has caused another problem when I log in to the console with the test user and select the public namespace, a warning of no permission will appear。

nacos server version can be tried with 1.4.0, which fixes the console problem

309283175:
[...]

PR #4273, opened 2020-11-19
[ISSUE-#4262] Fix public namespace permission problem

horizonzy:
For-#4262, #3524

### DIFF
--- a/common/src/main/java/com/alibaba/nacos/common/utils/NamespaceUtil.java
+++ b/common/src/main/java/com/alibaba/nacos/common/utils/NamespaceUtil.java
@@ -14,9 +14,7 @@
  * limitations under the License.
  */
 
-package com.alibaba.nacos.config.server.utils;
-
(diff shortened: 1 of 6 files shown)


Maijh97 (approved):

sanxun0325 (approved):

KomachiSion:
What will happen if users upgrade from old version? 
Whether old client can use after this PR merged?

horizonzy:
> What will happen if users upgrade from old version?
> Whether old client can use after this PR merged?

this is server side logic, whatever user set namespace with value `public` or not set, server all handle it to `:*:*`, it match our permission design.

KomachiSion:
So only affect the namespace during authorization verification, and still use the user's input for actual operations?

horizonzy:
yeah, this handle just affect auth process, if some operation need handle namespace, just use namespaceUtils by itsself.

KomachiSion (approved):

Issue #4232, opened 2020-11-15
Tool class replacement for environment variable reading


chuntaojun:
**Is your feature request related to a problem? Please describe.**
Tool class replacement for environment variable reading

ApplicationUtil => EnvUtil

**Describe the solution you'd like**

All Environment operations are handled by EnvUtil, Some methods in EnvUtil, ApplicationUtil no longer takes over

**Describe alternatives you've considered**

**Additional context**

horizonzy:
I will solve it.

Maijh97:
@horizonzy Have finished?

horizonzy:
yes.

horizonzy:
close it

PR #4281, opened 2020-11-20
[ISSUE-#4232] Use EnvUtil to replace ApplicationUtils about env operation

horizonzy:
For-#4232


KomachiSion (dismissed):

sys/src/main/java/com/alibaba/nacos/sys/env/EnvUtil.java
```diff
@@ -91,6 +92,26 @@ public static void setEnvironment(ConfigurableEnvironment environment) {
         EnvUtil.environment = environment;
     }
     
+    public static String[] getActiveProfiles() {
```
chuntaojun, line 95: spring这个profile的使用不要加了

horizonzy: ok

chuntaojun (approved):

Issue #4323, opened 2020-11-24
Remove lombok dependency


horizonzy:
**Describe the bug**

Remove lombok dependency

paderlol:
Why need to remove this dependency？

PR #4324, opened 2020-11-24
[ISSUE-#4323] Remove lombok dependency

horizonzy:

For-#4323

### DIFF
--- a/config/pom.xml
+++ b/config/pom.xml
@@ -54,12 +54,6 @@
         </dependency>
 
         <dependency>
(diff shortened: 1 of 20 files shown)


paderlol:
Why need to remove lombok?

horizonzy:
> Why need to remove lombok?


I think it is not nesessary for developer, shouldn't let user to use it forcely.

paderlol:
This is only on the server-side, not the client-side, and Lombok is currently better, isn’t it? So I didn’t find a reason to get rid of him.

Issue #4344, opened 2020-11-26
cacheMap property optimization


christophe-liwei:
I wonder why `cacheMap` properties use the AtomicReference wrapper in `ClientWorker`, but cacheMap writes with  lock. What's the point of using AtomicReference? I think `ConcurrentHashMap` could be used to replace it with a lock-free operation to improve concurrency.

![image](https://user-images.githubusercontent.com/20626345/100333749-ec2cd000-300d-11eb-8cf5-bc291227fc53.png)

paderlol:
Because this is not an operation, concurrenthashmap can only guarantee its own operation, and there is a reference replacement operation here

christophe-liwei:
> Because this is not an operation, concurrenthashmap can only guarantee its own operation, and there is a reference replacement operation here

What should replace the reference?

christophe-liwei:
> Because this is not an operation, concurrenthashmap can only guarantee its own operation, and there is a reference replacement operation here

What's the meaning of using a set method to replace a reference after synchronized so why not use CAS?

paderlol:
Maybe what I said is not clear enough. In many cases, not only do operations on the cachemap, it is a complete operation unit, and then why not use concurrent operations, because essentially the contents of the concurrenthashmap are also synchronized for locking

christophe-liwei:
> Maybe what I said is not clear enough. In many cases, not only do operations on the cachemap, it is a complete operation unit, and then why not use concurrent operations, because essentially the contents of the concurrenthashmap are also synchronized for locking

I think this code can be modified in the following way：

```
 CacheData cache = getCache(dataId, group, tenant);
        if (null != cache) {
            return cache;
        }
        String key = GroupKey.getKeyTenant(dataId, group, tenant);
        synchronized (cacheMap) {
            CacheData cacheFromMap = getCache(dataId, group, tenant);
            // multiple listeners on the same dataid+group and race condition,so
            // double check again
            // other listener thread beat me to set to cacheMap
            if (null != cacheFromMap) {
                cache = cacheFromMap;
                // reset so that server not hang this check
                cache.setInitializing(true);
            } else {
                cache = new CacheData(configFilterChainManager, agent.getName(), dataId, group, tenant);
                // fix issue # 1317
                if (enableRemoteSyncConfig) {
                    String[] ct = getServerConfig(dataId, group, tenant, 3000L);
                    cache.setContent(ct[0]);
                }
            }

[...]

PR #4352, opened 2020-11-27
issues #4344

christophe-liwei:
[issues 4344](https://github.com/alibaba/nacos/issues/4344)
The `cacheMap` attribute in `ClientWorker` now uese an instance lock, which can be reduced by `putIfAbsent` from `ConcurrentHashMap`.


christophe-liwei:
br error

Issue #4403, opened 2020-12-03
Didn't clear datums in com.alibaba.nacos.naming.consistency.persistent.raft.RaftCore#shutdown.


horizonzy:
**Describe the bug**
Now in `com.alibaba.nacos.naming.consistency.persistent.raft.RaftConsistencyServiceImpl#RaftConsistencyServiceImpl`'s constructor, it will subcribe isAllNewVersion. If it's true, will shutdown old raft core. 

In the shutdown logic, it will the clear old datums, the old datums load from old version raft file when application start, the data already act on serviceManager and service. If all server are version, we can stop the old raft work, but shouldn't clear datums.

Particularly, when I start nacos with standalone mode, `isAllNewVersion` will direct return true, there will be strange phenomenon in this case. If serviceManager listen raftCore before raftCore shutdown, it will read instances from old datum. If serviceManager listen raftCore after raftCore shuwdown, it can't read instances from old datum.

So, I think we just stop the raft core work, needn't clear old datums.

KomachiSion:
If start standalone, it should old all raft and clear all raft data to release memory. And read data from disk directly.

horizonzy:
> If start standalone, it should not start any raft and clear all raft data to release memory. And read data from disk directly.
> 
> But your problem might happen in one node cluster. So we need to discuss a solution to release this memory with a safety way.

I'm sure I start with standalone mode, `com.alibaba.nacos.naming.consistency.persistent.raft.RaftCore#init`, it will load data from disk, and then `com.alibaba.nacos.naming.core.ServiceManager#init`, it will listen raftCore datum, load service. In service init, it also listen raftCore datum, load instances. 

The process rely on raftCore datums, so if the process before or after datums clear, the result will not be same.

KomachiSion:
> > If start standalone, it should not start any raft and clear all raft data to release memory. And read data from disk directly.
> > But your problem might happen in one node cluster. So we need to discuss a solution to release this memory with a safety way.
> 
> I'm sure I start with standalone mode, `com.alibaba.nacos.naming.consistency.persistent.raft.RaftCore#init`, it will load data from disk, and then `com.alibaba.nacos.naming.core.ServiceManager#init`, it will listen raftCore datum, load service. In service init, it also listen raftCore datum, load instances.
> 
> The process rely on raftCore datums, so if the process before or after datums clear, the result will not be same.

Yes, beacuse current version has no completely implement what's I said.

horizonzy:
[...]

PR #4404, opened 2020-12-03
[ISSUE-#4403] Not clear datums when raftCore shutdown.

horizonzy:
For-#4403


naming/src/main/java/com/alibaba/nacos/naming/consistency/persistent/raft/RaftCore.java
```diff
@@ -464,8 +464,6 @@ public void shutdown() throws NacosException {
         masterTask.cancel(true);
         Loggers.RAFT.warn("stop old raft protocol task for heartbeat task");
         heartbeatTask.cancel(true);
-        Loggers.RAFT.warn("clean old cache datum for old raft");
-        datums.clear();
```
chuntaojun, line 468: This is cached data, not really persistent data，so clear this map it‘s no problem

Issue #4467, opened 2020-12-12
The instanceId(use snowflake generator ) is different when register same instance twice  


horizonzy:
**Describe the bug**
Now the user register instance with metadata {"preserved.instance.id.generator", "snowflake"}. It will generate instanceId with auto increment number. 
If we register the instance again or update is't metadata in console, it's id will change 0 -> 1 and 1 -> 0 again and again.

**Expected behavior**
If the origin id is snowflake id, just remains it.

KomachiSion:
Same instance register twice, should different instanceId.

horizonzy:
> If same instance register twice, should different instanceId.
> 
> If update instance, the instanceId should be same.

We can't differ update or twice register. we modify the instance matadata in console, it's update. But in controller, it's will register again.

KomachiSion:
So I think just judge whether old instance exist, if exist, use old instanceId, instanceId should not be update except deregister then register.

Or we need to discuss whether push down instance generating to client part. I think id generate in server part is not reasonable.

horizonzy:
> So I think just judge whether old instance exist, if exist, use old instanceId, instanceId should not be update except deregister then register.
> 
> Or we need to discuss whether push down instance generating to client part. I think id generate in server part is not reasonable.

If generating in client side, we can't gan guarantee that the instanceId will repeat, so also need judge in server side. 
It's not a good idea.

KomachiSion:
Users to guarantee. the instance definition is controlled by users. We only received and apply it.

The snowflake need add an workId, which need user input. It can make sure instance not duplicate.

horizonzy:
> Users to guarantee. the instance definition is controlled by users. We only received and apply it.
> 
> The snowflake need add an workId, which need user input. It can make sure instance not duplicate.

what's the concept of workId?

KomachiSion:
this is a concept from snowflake.

horizonzy:
> this is a concept from snowflake.

good idea.

KomachiSion:
Even add it in server part, user also need to learn workid and snowflake, otherwise they can't use well because `User` includes `developer` and `maintainer`.

BTW, the instanceId means the `identify of instance`, not nacos identify. So I think generate by client with app is more reasonable.

KomachiSion:
But now we has implemented in server part, so we need to adapt it first in a few time.

horizonzy:
[...]

PR #4468, opened 2020-12-12
[ISSUE-#4467] Fix snowflake id problem

horizonzy:

Fix-#4467

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/core/ServiceManager.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/core/ServiceManager.java
@@ -963,8 +963,7 @@ public class ServiceManager implements RecordListener<Service> {
                 List<Instance> instances = service.allIPs();
                 for (Instance instance : instances) {
                     if (IPUtil.containsPort(containedInstance)) {
-                        if (StringUtils.equals(instance.getIp() + IPUtil.IP_PORT_SPLITER + instance.getPort(),
-                                containedInstance)) {
+                        if (StringUtils.equals(instance.getIp() + IPUtil.IP_PORT_SPLITER + instance.getPort(), containedInstance)) {
                             contained = true;
                             break;
                         }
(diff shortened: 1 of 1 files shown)


naming/src/main/java/com/alibaba/nacos/naming/core/ServiceManager.java
```diff
@@ -811,7 +811,15 @@ public Instance getInstance(String namespaceId, String serviceName, String clust
...
-                instance.setInstanceId(instance.generateInstanceId(currentInstanceIds));
+                String instanceIdGenerator = instance.getInstanceIdGenerator();
+                if (Constants.SNOWFLAKE_INSTANCE_ID_GENERATOR.equalsIgnoreCase(instanceIdGenerator)) {
+                    Instance oldInstance = instanceMap.get(instance.getDatumKey());
+                    if (oldInstance != null && oldInstance.isSnowflakeId()) {
+                        instance.setInstanceId(oldInstance.getInstanceId());
+                    } else {
+                        instance.setInstanceId(generateSnowflakeInstanceId(currentInstanceIds));
+                    }
+                }
```
KomachiSion, line 822: Bad implementation I think. If oldInstance exist, only use instance id of old instance, can't change instanceId.

The original implementation also has some problem

horizonzy: some situation.
1. old instance didn't use snowflake generator, new instance use snowflake generator. New instance id should be snowflake Id.
2.old instance use snowflake generator, new instance isn't use  snowflake generator. New instance id shouldn't remains old instance Id. 
3.old instance and new instance all use snowflake generator, new instance should remains old instance Id.

KomachiSion: 1. Must deregister instance and register again.

2. Same as the first

3. yes it is.

KomachiSion: Instance id is a binder like ip and port (so default id generate from ip and port), If user want to change, they **must** deregister the older and register new. Can't update it directly.

horizonzy: > Instance id is a binder like ip and port (so default id generate from ip and port), If user want to change, they **must** deregister the older and register new. Can't update it directly.

agree

KomachiSion (approved):

PR #4504, opened 2020-12-16
Fix response desc always null

horizonzy:
the param is `decs `, and `this.desc = desc;`
so response desc always null


KomachiSion (approved):

KomachiSion (approved):

KomachiSion:
I think we should use lombok to avoid this problem. or use IDEA tool to help us create the constructor and setter getter.

PR #4505, opened 2020-12-16
Optimize PayloadRegistry#init

horizonzy:
split request and response package handle. reduce unnecessary scan operation.


horizonzy:
In client-sdk, `com.alibaba.nacos.naming.cluster.remote.request` and `com.alibaba.nacos.naming.cluster.remote.response` isn't in the classpath, It's also needn't to scan, need to differ server side and client side?

KomachiSion:
>In client-sdk, com.alibaba.nacos.naming.cluster.remote.request and com.alibaba.nacos.naming.cluster.remote.response isn't in the classpath. It's also needn't to scan, need to differ server side and client side?

We need to refactor this part, but it not affect usage.  Can keep scanning now.

KomachiSion (approved):

PR #4516, opened 2020-12-18
Append java doc info for feature that get all services

horizonzy:

in #4133, add feature that get all services without groupName filter. 
But forget append doc info in sdk.


KomachiSion (changes requested):
Sorry, we don't want to client query all group, client should know which group it wants. query all group may cause performance problem in server part.

Fuzzy query is design for ops part. we may define in ops sdk. Not user client.

Issue #4522, opened 2020-12-18
Fix spring.config.location not right in startup script.


horizonzy:
**Describe the bug**
Start nacos server in startup. The startup param spring.config.location as follow:
```
spring.config.location=file:/Users/apple/Downloads/nacos/conf/,classpath:/,classpath:/config/,file:./,file:./config/ 
```
The spring.config.location from startup script:
```
set DEFAULT_SEARCH_LOCATIONS="classpath:/,classpath:/config/,file:./,file:./config/"
set CUSTOM_SEARCH_LOCATIONS=file:%BASE_DIR%/conf/,%DEFAULT_SEARCH_LOCATIONS%
...
JAVA_OPT="${JAVA_OPT} --spring.config.location=${CUSTOM_SEARCH_LOCATIONS}"

```

The logic in com.alibaba.nacos.sys.env.EnvUtil#getCustomFileResource will use ./config to load properties, it's not right

The param DEFAULT_SEARCH_LOCATIONS can be removed.

KomachiSion:
Maybe Refer to #4541

PR #4523, opened 2020-12-18
[ISSUE-#4522] Remove CUSTOM_SEARCH_LOCATIONS in startup script

horizonzy:
For-#4522


KomachiSion:
I think we should discuss more about the start script and Env properties. We modified and fixed it too many times!!

Issue #3880, opened 2020-09-21
update service status bug


YyyQq-C:
### code bug

- com.alibaba.nacos.api.naming.pojo.ServiceInfo#validate function's return always true. 
```java
    public boolean validate() {
        if (isAllIPs()) {
            return true;
        }

        List<Instance> validHosts = new ArrayList<Instance>();
        for (Instance host : hosts) {
            if (!host.isHealthy()) {
                continue;
            }

            for (int i = 0; i < host.getWeight(); i++) {
                validHosts.add(host);
            }
        }

        return true;
    }
```
- com.alibaba.nacos.api.naming.pojo.ServiceInfo#getHosts function's return always not null
```java
    public List<Instance> getHosts() {
        return new ArrayList<Instance>(hosts);
    }
```

- com.alibaba.nacos.client.naming.core.HostReactor#processServiceJSON 
```java

100. public ServiceInfo processServiceJSON(String json) {
101.         ServiceInfo serviceInfo = JSON.parseObject(json, ServiceInfo.class);
102.         ServiceInfo oldService = serviceInfoMap.get(serviceInfo.getKey());
103.         if (serviceInfo.getHosts() == null || !serviceInfo.validate()) {
104.             //empty or error push, just ignore
105.             return oldService;
106.         }
107. 
108.         if (oldService != null) {
109.             if (oldService.getLastRefTime() > serviceInfo.getLastRefTime()) {
110.                 NAMING_LOGGER.warn("out of date data received, old-t: " + oldService.getLastRefTime()
111.                     + ", new-t: " + serviceInfo.getLastRefTime());
112.             }
113. 
114.             serviceInfoMap.put(serviceInfo.getKey(), serviceInfo);
115.             //.....
116.          }
117.         //....
118. }
``` 
line 103 judge always false.

KomachiSion:
I think the author want to implement `protection for push empty`, but don't implement completely.

I'm not  sure whether is a bug. But it is not useful for now.

horizonzy:
> I think the author want to implement `protection for push empty`, but don't implement completely.
> 
> I'm not sure whether is a bug. But it is not useful for now.
agree, the origin logic is not suitable, client's serviceInfo should be same as server, even though hosts is empty.  the code should be clean maybe

KomachiSion:
I think we should add a properties to open `push empty protection`. And fix this part logic.

PR #4665, opened 2021-01-10
[ISSUE-#3880] NamingService Client support pushEmptyProtection.

horizonzy:
For-#3880

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/naming/core/HostReactor.java
+++ b/client/src/main/java/com/alibaba/nacos/client/naming/core/HostReactor.java
@@ -78,16 +78,18 @@ public class HostReactor implements Closeable {
     
     private final String cacheDir;
     
+    private final boolean pushEmptyProtection;
+    
     private final ScheduledExecutorService executor;
     
     private final InstancesChangeNotifier notifier;
     
     public HostReactor(NamingProxy serverProxy, BeatReactor beatReactor, String cacheDir) {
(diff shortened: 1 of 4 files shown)


client/src/main/java/com/alibaba/nacos/client/naming/core/HostReactor.java
```diff
@@ -161,7 +163,8 @@ public void unSubscribe(String serviceName, String clusters, EventListener event
     public ServiceInfo processServiceJson(String json) {
         ServiceInfo serviceInfo = JacksonUtils.toObj(json, ServiceInfo.class);
         ServiceInfo oldService = serviceInfoMap.get(serviceInfo.getKey());
-        if (serviceInfo.getHosts() == null || !serviceInfo.validate()) {
+        
+        if (pushEmptyProtection && !serviceInfo.validate()) {
```
KomachiSion, line 167: should we judge null for serviceInfo.getHosts?
It come from json, and might is null, if hosts is null. the validate method may throw NPE.

horizonzy: the hosts cann't be null, it's instance variable.

KomachiSion: hosts is not final. And the set method can directly set it.
If the json return "{hosts: null}", it will be null

horizonzy: right, should be check. Can't guarantee the server side response value.

KomachiSion (approved):

PR #4683, opened 2021-01-13
[ISSUE #4593] Add server identity to replace user-agent white list.

KomachiSion:

For #4593 , Add server identity to replace user-agent white list.

Templately Fix security problem. User can shutdown user-agent white list and defined their own identity key and value for request between servers.

## Brief changelog

- Add server identity to replace user-agent white list.

## Verifying this change

XXXX

### DIFF
--- a/core/src/main/java/com/alibaba/nacos/core/auth/AuthFilter.java
+++ b/core/src/main/java/com/alibaba/nacos/core/auth/AuthFilter.java
@@ -24,9 +24,9 @@ import com.alibaba.nacos.auth.model.Permission;
 import com.alibaba.nacos.auth.parser.ResourceParser;
 import com.alibaba.nacos.common.utils.ExceptionUtil;
 import com.alibaba.nacos.core.code.ControllerMethodsCache;
-import com.alibaba.nacos.sys.env.Constants;
 import com.alibaba.nacos.core.utils.Loggers;
 import com.alibaba.nacos.core.utils.WebUtils;
+import com.alibaba.nacos.sys.env.Constants;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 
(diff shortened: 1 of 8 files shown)


KeRan213539 (approved):

core/src/main/java/com/alibaba/nacos/core/auth/AuthFilter.java
```diff
@@ -73,10 +73,25 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
+        if (authConfigs.isEnableUserAgentAuthWhite()) {
+            String userAgent = WebUtils.getUserAgent(req);
+            if (StringUtils.startsWith(userAgent, Constants.NACOS_SERVER_HEADER)) {
+                chain.doFilter(request, response);
+                return;
+            }
+        } else if (StringUtils.isNotBlank(authConfigs.getServerIdentityKey()) && StringUtils
+                .isNotBlank(authConfigs.getServerIdentityValue())) {
+            String serverIdentity = req.getHeader(authConfigs.getServerIdentityKey());
+            if (authConfigs.getServerIdentityValue().equals(serverIdentity)) {
```
QuantumGhost, line 85: Please note that comparing credential with `equals` makes your application vulnerable to [timing attack](https://en.wikipedia.org/wiki/Timing_attack)

KomachiSion: How to solve it? Can you provide some advise for this?

Slahser: https://github.com/playframework/play1/blob/master/framework/src/play/mvc/CookieDataCodec.java#L82

KomachiSion: In my opinion,  there is enough for equals. reason follow:

1. Nacos is Internal IDC application, we recommend users use Nacos with internal network not public network, If hacker want to timing attack, the network has been controlled by hacker. And Cost much time to analyze package between nacos server.
2. The key and value is defined by configuration file, as the first item said, if the internal network is in controlled by hacker. they can find the key and value by configuration file directly.

I'm not sure whether the timing attack is send a long string to hold equals resource. If so, I think we has no ability to solve all `public application security problem` for an internal network application.

Zheaoli: @KomachiSion FYI maybe you should use [`MessageDigest.isEquals` ](https://docs.oracle.com/javase/7/docs/api/java/security/MessageDigest.html)

Issue #4452, opened 2020-12-11
Can nacos support the environment configuration comparison function similar to appolo?


lidaoyang:
Can nacos support the environment configuration comparison function similar to appolo?

nacos 可以支持类似appolo的环境配置对比功能吗？
![image](https://user-images.githubusercontent.com/26211950/101854576-ca4e5400-3b9c-11eb-96e2-8247b0bc0aa4.png)

jzhishu:
> Can nacos support the environment configuration comparison function similar to appolo?
> 
> nacos 可以支持类似appolo的环境配置对比功能吗？
> ![image](https://user-images.githubusercontent.com/26211950/101854576-ca4e5400-3b9c-11eb-96e2-8247b0bc0aa4.png)

Could you please add more pictures to describe the function？

lidaoyang:
> > Can nacos support the environment configuration comparison function similar to appolo?
> > nacos 可以支持类似appolo的环境配置对比功能吗？
> > ![image](https://user-images.githubusercontent.com/26211950/101854576-ca4e5400-3b9c-11eb-96e2-8247b0bc0aa4.png)
> 
> Could you please add more pictures to describe the function？

apollo is a comparison configuration item. nacos cannot do this. You can compare the differences between the two namespaces.
apollo 那边的是比较配置项，nacos这边应该做不到，可以比较2个命名空间下的配置内容的差异。

apollo 演示地址：http://106.54.227.205/config/diff.html?#/appid=liasmy&env=DEV&clusterName=default&namespaceName=development.tom-public
![image](https://user-images.githubusercontent.com/26211950/102735280-f2983880-437c-11eb-80e5-9fba9f7d4377.png)
![image](https://user-images.githubusercontent.com/26211950/102735265-e8763a00-437c-11eb-8b01-1f56058fa6ab.png)

BlackBAKA:
@i will solve it@

BlackBAKA:
> > > Can nacos support the environment configuration comparison function similar to appolo?
> > > nacos 可以支持类似appolo的环境配置对比功能吗？
> > > ![image](https://user-images.githubusercontent.com/26211950/101854576-ca4e5400-3b9c-11eb-96e2-8247b0bc0aa4.png)
> > 
> > 
> > Could you please add more pictures to describe the function？
> 
> apollo is a comparison configuration item. nacos cannot do this. You can compare the differences between the two namespaces.
[...]

PR #4720, opened 2021-01-16
[ISSUE #4452]Can nacos support the environment configuration comparison function similar to appolo?

BlackBAKA:
# What is the purpose of the change

issue- #4452

### DIFF
--- a/config/pom.xml
+++ b/config/pom.xml
@@ -107,6 +107,15 @@
             <artifactId>jackson-databind</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.fasterxml.jackson.dataformat</groupId>
+            <artifactId>jackson-dataformat-properties</artifactId>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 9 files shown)


KomachiSion:
I think your implementation is complicated. I suggest you refer to the comparing for release configuration with the historical version and use that way to implement this features.

BlackBAKA:
> I think your implementation is complicated. I suggest you refer to the comparing for release configuration with the historical version and use that way to implement this features.

Do you means compare config to their history, not compare to any config?

KomachiSion:
> > I think your implementation is complicated. I suggest you refer to the comparing for release configuration with the historical version and use that way to implement this features.
> 
> Do you means compare config to their history, not compare to any config?

No, I mean you can imitate the features of comparing history to realize the feature of comparing different environment configurations.

KomachiSion:
Like this
image

KomachiSion:
This feature has finished.

PR #4778, opened 2021-01-22
make SecurityProxy.accessToken threadsafe in single writer  multi reader

shalk:

- add volatile keywork, make SecurityProxy.accessToken  threadsafe in single writer multi reader.
- adjust initRefreshTask, make SecurityProxy.login not running concurrently

## Brief changelog

XX

## Verifying this change

XXXX

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/naming/net/NamingProxy.java
+++ b/client/src/main/java/com/alibaba/nacos/client/naming/net/NamingProxy.java
@@ -142,6 +142,9 @@ public class NamingProxy implements Closeable {
             }
         });
         
+        refreshSrvIfNeed();
+        this.securityProxy.login(getServerList());
+        
         this.executorService.scheduleWithFixedDelay(new Runnable() {
             @Override
             public void run() {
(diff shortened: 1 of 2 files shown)


KomachiSion (changes requested):
In login method, there is judgement for token refresh windows.

And both of them is in init method, no multiple reader I think.

shalk:
> In login method, there is judgement for token refresh windows.
> 
> And both of them is in init method, no multiple reader I think.

`refresh windows` can not lock the login(Server) at first time,  because initDely is zero,  and `securityProxy.login(getServerList())` will run in different  thread.

the problem is **two writers work on share var without `volatile` or `lock`.**

If we change the order , `securityProxy.login(getServerList())` will running in order like single writer.  ref [ServerHttpAgent#L278](https://github.com/alibaba/nacos/blob/a4e3b7cbc75678bf8d813ac0a4c2ac1772878f8f/client/src/main/java/com/alibaba/nacos/client/config/http/ServerHttpAgent.java#L278)

May be this is not big deal, but it is good convention for concurrent program.

KomachiSion (approved):

PR #4838, opened 2021-01-30
ISSUE #4839 Add the interface of loading NacosNamingService with the spec…

65487123:
…ified class loader

When NamingService is created, NacosNamingService.class is loaded through reflection. When NacosNamingService and NamingFactory (current class) are in the same classpath, they can be successfully loaded. If NamingFactory and NacosNamingService are not in the same classpath (such as OSGI, different bundles will be different Class loader loaded), an exception will be thrown. So it should provide a method to load NacosNamingService with the specified classoader


KomachiSion:
Why add this? Please submit issue to discuss first.

PR #4948, opened 2021-02-24
[ISSUE #4925] correct the member's state when the cluster information…

MajorHe1:

try to fix issue #4925

## Brief changelog

1. set the member's state to DOWN when the member joins the server list for the first time in case of the corresponding nacos-server has not been started.
2.  Keep the state of member consistent with the value in the server list， otherwise there will be problems like issue #4925
## Verifying this change

XXXX

### DIFF
--- a/core/src/main/java/com/alibaba/nacos/core/cluster/ServerMemberManager.java
+++ b/core/src/main/java/com/alibaba/nacos/core/cluster/ServerMemberManager.java
@@ -307,6 +307,14 @@ public class ServerMemberManager implements ApplicationListener<WebServerInitial
             
             if (!serverList.containsKey(address)) {
                 hasChange = true;
+                // If the cluster information in cluster.conf or address-server has been changed,
+                // while the corresponding nacos-server has not been started yet, the member's state
+                // should be set to DOWN. If the corresponding nacos-server has been started, the
+                // member's state will be set to UP after detection in a few seconds.
+                member.setState(NodeState.DOWN);
+            } else {
+                //fix issue # 4925
+                member.setState(serverList.get(address).getState());
             }
             
             // Ensure that the node is created only once


core/src/main/java/com/alibaba/nacos/core/cluster/ServerMemberManager.java
```diff
@@ -307,6 +307,14 @@ synchronized boolean memberChange(Collection<Member> members) {
...
             if (!serverList.containsKey(address)) {
                 hasChange = true;
+                // If the cluster information in cluster.conf or address-server has been changed,
+                // while the corresponding nacos-server has not been started yet, the member's state
+                // should be set to DOWN. If the corresponding nacos-server has been started, the
+                // member's state will be set to UP after detection in a few seconds.
+                member.setState(NodeState.DOWN);
+            } else {
+                //fix issue # 4925
+                member.setState(serverList.get(address).getState());
```
KomachiSion, line 317: I don't think it's a right fix.

this method is update members, so we should use input member state I think.

KomachiSion (changes requested):

core/src/main/java/com/alibaba/nacos/core/cluster/ServerMemberManager.java
MajorHe1: I don't think we could use input member state in this method 
`synchronized boolean memberChange(Collection<Member> members) {}`

This method is only called in three places：
1.    
image

In this case, because **the member state is set to UP by default**,  whether it's from cluster.conf  or address-server.
image

If a member has been judged as down,  it will be reset to UP after this method. **So this issue 4925 occured**.

2. 
image

In this case, a joined member is not in the serverList. We are **not sure** if the joined member's process has started, so it should be set to DOWN at first, and set to UP after MemberInfoReportTask completed.

3.
image

This has no effect on the method memberLeave()

KomachiSion: I'm not sure whether it will affect other report or update states workflow. I think we need do some discussion.

@chuntaojun @shiyiyue1102 How do you think about?

MajorHe1: Not only the state value of the member, I think all the properties of the member should be based on the value in the serverlist. 
Otherwise, the data returned by the interface /nacos/v1/ns/operator/servers is always changing (be reset)，for the same reason as issue 4925.

KomachiSion: But some of other Member attributes should modified by other members‘ report.
So I think the final value should be updated from the inputs.

chuntaojun: this is temporary plan, finally we should figure out the remote data update and local data update to solve this problem

chuntaojun (approved):

KomachiSion (approved):

core/src/main/java/com/alibaba/nacos/core/cluster/ServerMemberManager.java
MajorHe1: > But some of other Member attributes should modified by other members‘ report.
> So I think the final value should be updated from the inputs.

There is no problem that  final value should be updated from the inputs, because this method 
memberChange(Collection<Member> members)
is not called when nacos-server received other members' report.
It seems that this method is only called when the cluster information is updated，I think the attribute value of member should be based on the value in nacos-server memory.

zrlw:
KomachiSion 的担心是对的，我们就碰上#4948修订导致的问题：
某个运维需要重新启动的nacos服务节点启动20多秒都不把心跳转发到负责的节点，后来发现因为4948把其他节点的状态从默认的UP改成了DOWN，从DOWN变UP要等到当前节点收到所有节点发来的健康检查报文，而其他节点发这个报文首先要等到重启的节点先发健康报文通知给他们。
这些问题的根源在于MemberInfoReportTask这个任务不仅是节点启动后延时5秒后开始执行，而且执行过程还是串行的，一次只发一个报文给目标节点，每次发送之间还要间隔2秒，如果有10个节点，等全部通知完了，25秒都过去了，期间这个重启的节点会越俎代庖收下本该由其他节点负责接收的心跳，后果就是其他节点会将心跳超期的实例下线。
我们改了MemberInfoReportTask，每次执行时都全量通知其他节点，免得节点多了通知太慢导致状态不一致。

PR #5139, opened 2021-03-19
Fix thread name generate strategy.

horizonzy:
String threadName = name + id.getAndDecrement(); -> String threadName = name + id.getAndIncrement();


KomachiSion:
I think the old author want use the negative as the split between name and id number.

horizonzy:
> I think the old author want use the negative as the split between name and id number.

hhh...

horizonzy:
> I think the old author want use the negative as the split between name and id number.

https://github.com/alibaba/nacos/blob/6df6097b63c73390892f5c79576de837bb34f795/common/src/main/java/com/alibaba/nacos/common/executor/NameThreadFactory.java#L35-L40
Maybe not, the split is DOT.

KomachiSion:
all right

KomachiSion (approved):

PR #5164, opened 2021-03-22
缓解nacos-client不停地清空服务列表的问题，客户端获取本地IP过滤loopback、anylocal和linklocal地址

zrlw:
将推空保护(称之为合法性校验可能更合适)默认值改为true，客户端获取本地IP过滤loopback、anylocal和linklocal地址。

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/naming/utils/NetUtils.java
+++ b/client/src/main/java/com/alibaba/nacos/client/naming/utils/NetUtils.java
@@ -44,7 +44,47 @@ public class NetUtils {
             String ip = System.getProperty("com.alibaba.nacos.client.naming.local.ip",
                     InetAddress.getLocalHost().getHostAddress());
             
-            return localIp = ip;
+            String ip = System.getProperty("com.alibaba.nacos.client.naming.local.ip");
+            if (null != ip) {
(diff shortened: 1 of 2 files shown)


KomachiSion (changes requested):
I explained in issue, the default value is false, because the old version has no push empty protection, for the same behavior, the default value is false.

and if you change the default value is true, validate method will check the host. So there's no need check isEmpty for hosts.

zrlw:
> I explained in issue, the default value is false, because the old version has no push empty protection, for the same behavior, the default value is false.
> 
> and if you change the default value is true, validate method will check the host. So there's no need check isEmpty for hosts.

Under Dubbo running environment, nacos-client old version 1.2.1 has some push empty protection while serviceInfo.getHosts() == null, but the new version 1.4.1 don't have any.
I don't think  the new added parameter 'pushEmptyProtection' has any practical meaning to our apps and just let  Dubbo clients lost Dubbo server list while turning provider off and on.

zrlw:
maybe  just remove the parameter 'namingPushEmptyProtection' and everything will be ok.

zrlw:
the loopback, anylocal, linklocal addresses should be filtered out as much as possible while getting local ip address.

PR #5221, opened 2021-03-28
filter out client loopback and local addresses, decrease delay between client beat check tasks

zrlw:
#5216


KomachiSion:
Please discuss in issue first.

PR #5242, opened 2021-03-30
Defer serverlist initialization, filter client loopback and local addresses, schedule beat task immediately, decrease client beat check task delay

zrlw:
#5240
#5216


KomachiSion:
Please one PR for one change and one issue

PR #5244, opened 2021-03-31
remove serverList initialization as self address is always added to serverList at memberChange()

### LINKED ISSUE #5240: com.alibaba.nacos.core.cluster.MemberUtilTest的setUp里的await建议加个超时设置
这个项目自动构建目前大概率会失败，总是在com.alibaba.nacos.core.cluster.MemberUtilTest测试方法hang掉。原因我没去查，因为本机测试没有出现这个问题。

latch.await();这行代码建议改一下，加个超时控制，免得一直等。

zrlw:
fix #5240

### DIFF
--- a/core/src/main/java/com/alibaba/nacos/core/cluster/ServerMemberManager.java
+++ b/core/src/main/java/com/alibaba/nacos/core/cluster/ServerMemberManager.java
@@ -134,14 +134,14 @@ public class ServerMemberManager implements ApplicationListener<WebServerInitial
         this.localAddress = InetUtils.getSelfIP() + ":" + port;
         this.self = MemberUtil.singleParse(this.localAddress);
         this.self.setExtendVal(MemberMetaDataConstants.VERSION, VersionUtils.version);
-        serverList.put(self.getAddress(), self);
         
         // register NodeChangeEvent publisher to NotifyManager
         registerClusterEvent();
         
         // Initializes the lookup mode
         initAndStartLookup();
-        
+
+        serverList.put(self.getAddress(), self);
         if (serverList.isEmpty()) {
             throw new NacosException(NacosException.SERVER_ERROR, "cannot get serverlist, so exit.");
         }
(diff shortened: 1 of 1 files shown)


KomachiSion (changes requested):
Maybe publish a member change event is better. Otherwise, the add self will not be notified.

zrlw:
> Maybe publish a member change event is better. Otherwise, the add self will not be notified.

keep serverlist empty before calling initAndStartLookup will publish a member change event.

KomachiSion (changes requested):
self should be add to server list always to ensure that after usage can get self.

zrlw:
> self should be add to server list always to ensure that after usage can get self.

the function memberChange is always called in initialization, self is always  added to server list even you don't add it explicitly.

KomachiSion:
I read the issue, and I found the problem is not in add self to list, the reason is that the Test case error.

In my opinion, when init, add self to list is a necessary operation, and the updateMembers add self is a protection operation. So no problem for this two codes.

In some case, for example, standalone or cluster but only one node, which means the server list after add self is same as read from loopUp, so it is no need to notified changed event, which also is right.

So I think solve problem should to fix the Test case.

zrlw:
> I read the issue, and I found the problem is not in add self to list, the reason is that the Test case error.
> 
> In my opinion, when init, add self to list is a necessary operation, and the updateMembers add self is a protection operation. So no problem for this two codes.
> 
> In some case, for example, standalone or cluster but only one node, which means the server list after add self is same as read from loopUp, so it is no need to notified changed event, which also is right.
> 
> So I think solve problem should to fix the Test case.

restarting standalone or one node cluster might also mean a change event, as least server startup time changed.

KomachiSion:
So maybe we need focus two question:

Whether member add self need send an event when init. 
And how to enhance the test case.

Moskalnazar (approved):

Moskalnazar (commented):
Thank you!

Moskalnazar (commented):
good

Issue #5236, opened 2021-03-30
After changing the account password, other logged-in browsers can still access 


myoss:
**Describe the bug**
1. Open 2 browsers(chrome, edge) and login in 
2. Modify password in chrome 
3. Safari browser can be accessed without logging in again 

**Expected behavior**
1. Safari browser can’t access  and must log in again 
2. Other users can’t access and must log in again 

**Acutally behavior**
1. Safari browser can still access
2. Other users can still access

**Desktop (please complete the following information):**
 - OS: Centos 7
 - Version nacos-server 1.3.2
 - Module console
 - SDK [spring-cloud-alibaba-nacos

JackSun-Developer:
I think this can be added as a feature.

Jayice-zjw:
i can solve it

PR #5254, opened 2021-03-31
Fix issue5236

Jayice-zjw:

fix issue https://github.com/alibaba/nacos/issues/5236


Jayice-zjw:
> > ## What is the purpose of the change
> > fix issue #5102
> 
> Hello, Jayice, The link is wrong. It should point at the ISSUE you want fix

fixed

PR #5255, opened 2021-03-31
[ISSUE #5236] token generation adds password factor

Jayice-zjw:

fix issue https://github.com/alibaba/nacos/issues/5236


chuntaojun:
In my opinion, the password is a sensitive information that is not suitable to be put in the token, and other information can be used to replace the perception of password update

KomachiSion:
> In my opinion, the password is a sensitive information that is not suitable to be put in the token, and other information can be used to replace the perception of password update

Agree

KomachiSion:
We need close PR first, this design need more discuss.

PR #5514, opened 2021-04-27
IPv6 and IPv4 check rules are optimized. splitIPPortStr method adds IPv4 check, and the optimized method does not affect the startup of container environment

pig-stdv:
The IPv6 judgment logic in this method will obtain IP address according to the host. In the container environment, K8S has not allocated IP to this pod, and then the judgment will report an error. After finding this problem, I gave feedback to remove the IPv4 judgment, but this problem still exists in IPv6. Suggested to remove the verification logic can be in other ways to optimize.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -30,3 +30,25 @@ under the License.
 Also, please refer to each LICENSE.<component>.txt file, which is located in
 the 'license' directory of the distribution file, for the license terms of the
 components that this product depends on.
+
+
+------
+com.alibaba.nacos.common.utils.InetAddressValidator.java in this product is
+copied from com.dynatrace.openkit.core.util.InetAddressValidator.java  of openkit-java project.
+https://github.com/Dynatrace/openkit-java
+                          openkit-java
+                    ======================
+
+     Copyright 2018-2021 Dynatrace LLC
+
+     Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


pig-stdv:
@KomachiSion

KomachiSion:
@KeRan213539 Please have some review and discuss.

Preparing release 1.4.2 and pending PR temparately.

pig-stdv:
IPv4 and IPv6 validation rule optimization (using regular), in order to prevent container deployment in the case of domain name resolution to add domain name rule validation, IPv6 rule validation before adding remove the slogan logic.

KeRan213539 (changes requested):
com.alibaba.nacos.sys.utils.InetUtils 中有个 isDomain 方法, 建议把那个方法里的实现也修改为不解析域名的(IPUtil不要去调用InetUtils#isDomain  ,因为两个类不在同一个模块, 仅修改 isDomain  的实现)
另外 IPUtil 中有个 isIP 方法, 这个方法也改为不解析域名的方式

pig-stdv:
>  IPUtil 中的 isIP 方法已经底层都是采用一套正则校验
> com.alibaba.nacos.sys.utils.InetUtils 中的 isDomain 方法已优化
>  NOTICE 添加许可

KomachiSion (approved):

KeRan213539 (approved):

Issue #5538, opened 2021-04-29
There are some failed test cases


SunJiFengPlus:
**Describe the bug**
![图片](https://user-images.githubusercontent.com/39667700/116557071-afecfb00-a930-11eb-8004-56c793f9de7b.png)

![图片](https://user-images.githubusercontent.com/39667700/116557325-eb87c500-a930-11eb-9225-e334ea9781c9.png)

If it passes, I can do this.

JackSun-Developer:
@ i will solve it@

KomachiSion:
@KeRan213539 Please check it

KomachiSion:
There are two PR for this issue? @KeRan213539 @SunJiFengPlus We need discuss first how to fix it.

PR #5620, opened 2021-05-09
[ISSUE #5538] fix test case fail

SunJiFengPlus:
fix test case fail

For #5538 

## Brief changelog
- fix test case fail

### DIFF
--- a/common/src/main/java/com/alibaba/nacos/common/utils/IPUtil.java
+++ b/common/src/main/java/com/alibaba/nacos/common/utils/IPUtil.java
@@ -53,6 +53,11 @@ public class IPUtil {
     
     private static Pattern domainPattern = Pattern.compile("[a-zA-Z0-9][-a-zA-Z0-9]{0,62}(\\.[a-zA-Z0-9][-a-zA-Z0-9]{0,62})+\\.?");
     
+    private static final String IPV4_TUPLE = "(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9][0-9]|[0-9])";
+    
+    private static final Pattern IPV4_PATTERN = Pattern
+            .compile("(?<!\\d)" + IPV4_TUPLE + "\\." + IPV4_TUPLE + "\\." + IPV4_TUPLE + "\\." + IPV4_TUPLE + "(?!\\d)");
+    
     /**
      * get localhost ip.
      * @return java.lang.String
@@ -156,7 +161,7 @@ public class IPUtil {
                 result = "";
             }
         } else {
-            Matcher m = InetAddressValidator.getIpv4Pattern().matcher(str);
+            Matcher m = IPV4_PATTERN.matcher(str);
             if (m.find()) {
                 result = m.group();
             }
--- a/common/src/test/java/com/alibaba/nacos/common/utils/IPUtilTest.java
+++ b/common/src/test/java/com/alibaba/nacos/common/utils/IPUtilTest.java
@@ -71,6 +71,18 @@ public class IPUtilTest {
         Assert.assertEquals("", IPUtil.getIPFromString(
                 "jdbc:mysql://666.288.333.444:3306/nacos_config_test?characterEncoding=utf8&connectTimeout=1000"
                         + "&socketTimeout=3000&autoReconnect=true&useUnicode=true&useSSL=false&serverTimezone=UTC"));
+        Assert.assertEquals("", IPUtil.getIPFromString(
+                "jdbc:mysql://292.168.1.1:3306/nacos_config_test?characterEncoding=utf8&connectTimeout=1000"
+                        + "&socketTimeout=3000&autoReconnect=true&useUnicode=true&useSSL=false&serverTimezone=UTC"));
+        Assert.assertEquals("", IPUtil.getIPFromString(
+                "jdbc:mysql://29.168.1.288:3306/nacos_config_test?characterEncoding=utf8&connectTimeout=1000"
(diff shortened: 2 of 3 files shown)


common/src/main/java/com/alibaba/nacos/common/utils/InetAddressValidator.java
```diff
@@ -37,8 +37,10 @@
     
     private static final int FIVE = 5;
     
+    private static final String IPV4_TUPLE = "(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9][0-9]|[0-9])";
+    
     private static final Pattern IPV4_PATTERN = Pattern
```
KeRan213539, line 42: Please do not modify this regular expression, IPUtil related please refer to https://github.com/alibaba/nacos/pull/5564/files modification

test/src/test/java/com/alibaba/nacos/test/core/JacksonUtilsTest.java
```diff
@@ -0,0 +1,43 @@
...
+	public void testToJson() {
+		Map<String, Map<String, Object>> map = new HashMap<>();
+		Map<String, Object> data = new HashMap<>();
+		data.put("test-1", LocalDateTime.of(2021, 10, 24, 10, 24));
+		data.put("test_2", LocalDateTime.of(2021, 6, 1, 6, 1));
+		map.put("global", data);
+
+		String json = com.alibaba.nacos.common.utils.JacksonUtils.toJson(map);
+		Assert.assertEquals("{\"global\":{\"test-1\":[2021,10,24,10,24],\"test_2\":[2021,6,1,6,1]}}", json);
+	}
```
KeRan213539, line 41: Please keep the original test and add your test

KeRan213539 (changes requested):

common/src/main/java/com/alibaba/nacos/common/utils/InetAddressValidator.java
SunJiFengPlus: in https://github.com/alibaba/nacos/pull/5564/files

This PR will not pass the following unit test case (in #5538)
```java
Assert.assertEquals("", IPUtil.getIPFromString(
                "jdbc:mysql://666.288.333.444:3306/nacos_config_test?characterEncoding=utf8&connectTimeout=1000"
                        + "&socketTimeout=3000&autoReconnect=true&useUnicode=true&useSSL=false&serverTimezone=UTC"));
```

test/src/test/java/com/alibaba/nacos/test/core/JacksonUtilsTest.java
SunJiFengPlus: I can do this, but I still find test cases without assertions meaningless

common/src/main/java/com/alibaba/nacos/common/utils/InetAddressValidator.java
SunJiFengPlus: I want to know why we need to nest and maintain two kinds of IP verification logic?

common/pom.xml
```diff
@@ -65,6 +65,10 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jsr310</artifactId>
+        </dependency>
```
KomachiSion, line 71: Why add this dependency? 
If just for test. I think we should change the test case not add dependency.

test/src/test/java/com/alibaba/nacos/test/core/JacksonUtilsTest.java
KomachiSion: Add the assert is right. But I have some problem for this case.

For this test, I think we should test the object to json and json to object by the Util. But this case only do some test for serialization of LocalDateTime which I remember nacos do not serialize.

I has no idea the original author add this cases. But I think if you want to fix this case and make it meaningful, you can add case for every line of test class without any more dependency.

common/src/main/java/com/alibaba/nacos/common/utils/InetAddressValidator.java
KomachiSion: @SunJiFengPlus This class is an validator for IP (including V6 and V4), which come from other community `Dynatrace LLC`. The reason why we include it is we need a useful validator to check the IPv6 address.

So there are two kinds of Ip validator temp. Maybe we can refactor `IpUtil` to use this validator only.

KomachiSion (changes requested):
@SunJiFengPlus Thanks for your contribution.

I think one PR for one think. You PR is linked to fix a test case issue. So I think you should focus on fixing them.

1. Check whether the test case is reasonable and right. If so, fix the source codes, otherwise fix the case.
2. Make cases pass first in this PR, If you think there are something need to enhance, you can  submit a new issue to discuss how to enhance them.

If you have any doubt, please comment and let us know.

common/src/main/java/com/alibaba/nacos/common/utils/InetAddressValidator.java
KeRan213539: This is using IPV4_TUPLE matching result of tuple.
image
The original expression does not have this problem, but it has another problem: it will match the whole string as IP. It cannot match the string containing IP.
So, first take the IP out of the string containing the IP, and then verify the IP
Or do you have a better way to solve this problem?

KeRan213539 (approved):

KomachiSion (approved):

Issue #5765, opened 2021-05-18
`IPUtil` can't match localhost


haoyann:
**Describe the bug**
`serverAddr` use `localhost:8848` then `IPUtil` can't match.
![image](https://user-images.githubusercontent.com/43994656/118587007-5b6dca80-b7ce-11eb-9cf8-7ac529f5b7eb.png)

**Desktop (please complete the following information):**
 - Version develop
 - Module [naming/config]
 - SDK [java]

**Additional context**
```
Caused by: java.lang.reflect.InvocationTargetException: null
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) ~[na:1.8.0_271]
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62) ~[na:1.8.0_271]
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) ~[na:1.8.0_271]
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423) ~[na:1.8.0_271]
	at com.alibaba.nacos.api.config.ConfigFactory.createConfigService(ConfigFactory.java:43) ~[nacos-api-2.0.2-SNAPSHOT.jar:na]
	... 45 common frames omitted
Caused by: java.lang.IllegalArgumentException: The IPv4 or Domain address("localhost") is incorrect.
	at com.alibaba.nacos.common.utils.IPUtil.splitIPPortStr(IPUtil.java:138) ~[nacos-common-2.0.2-SNAPSHOT.jar:na]
	at com.alibaba.nacos.client.config.impl.ServerListManager.<init>(ServerListManager.java:162) ~[nacos-client-2.0.2-SNAPSHOT.jar:na]
	at com.alibaba.nacos.client.config.NacosConfigService.<init>(NacosConfigService.java:74) ~[nacos-client-2.0.2-SNAPSHOT.jar:na]
	... 50 common frames omitted
```

JackSun-Developer:
This problem can be reproduced stably.

MentosL:
@JackSun-Developer   @haoyann   The Develop branch has been changed and should work properly。

![image](https://user-images.githubusercontent.com/33775732/118622757-31ca9880-b7fa-11eb-9490-f3b42b041299.png)

KomachiSion:
https://github.com/alibaba/nacos/pull/5514 This PR caused

[...]

PR #5787, opened 2021-05-19
[ISSUE #5765] fix localhost not match ip

SunJiFengPlus:
For #5765 
fix localhost not match ip

## Brief changelog

- fix problem
- add unit test


KomachiSion (changes requested):
Should localhost be in isDomain？

SunJiFengPlus:
> Should localhost be in isDomain？

My personal understanding is that domain can be resolved to IP, for example ```github.com``` can be resolved to ```140.82.113.3```. In the current context, ```localhost``` can be resolved to ```127.0.0.1```, so ```localhost``` is a domain, ```127.0.0.1``` is an IP

KomachiSion (approved):

PR #5852, opened 2021-05-25
[ISSUE #5765] fix localhost not match ip

SunJiFengPlus:
For #5765 
fix ```localhost``` in cluster.conf not available

## Brief changelog

- rename IPUitl to InternetAddressUtil
- move isDomain to InternetAddressUtil
- fix localhost not match ip

## Verifying this change

[unit test case](https://github.com/alibaba/nacos/commit/ad541f664c99c8dde3ec76e7b0dc4a9e87132a68#diff-e3feb653a6109bcc8e7bbc7974c2d6fa93b17cc25783e9a26f31d86f36f2d471R103) pass

### DIFF
--- a/common/src/main/java/com/alibaba/nacos/common/utils/InternetAddressUtil.java
+++ b/common/src/main/java/com/alibaba/nacos/common/utils/InternetAddressUtil.java
@@ -137,7 +137,7 @@ public class InternetAddressUtil {
                 throw new IllegalArgumentException("The IP address(\"" + str
                         + "\") is incorrect. If it is an IPv6 address, please use [] to enclose the IP part!");
             }
-            if (!isIPv4(serverAddrArr[0]) && !DOMAIN_PATTERN.matcher(serverAddrArr[0]).matches()) {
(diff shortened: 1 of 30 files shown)


common/src/main/java/com/alibaba/nacos/common/utils/InternetAddressUtil.java
```diff
@@ -215,4 +218,20 @@ public static String removeBrackets(String str) {
         return str.replaceAll("[\\[\\]]", "");
     }
     
+    /**
+     * juege str is right domain.（Check only rule）
```
KomachiSion, line 222: juede -- > judge

KomachiSion (changes requested):

common/src/main/java/com/alibaba/nacos/common/utils/InternetAddressUtil.java
SunJiFengPlus: I've fix it

KomachiSion (approved):

PR #5889, opened 2021-05-28
Feat resource parser group 

CherishCai:
feat #5888


KomachiSion (changes requested):
Can you add some unit test for this changes？

KomachiSion (approved):

Issue #5899, opened 2021-05-30
Double negation in code


7will10:
Double negation in code
代码中if判断使用双重否定，已经很长时间了还是看着不舒服~~

7will10:
@ready to close@

KomachiSion:
Reason has comment in pr

PR #5900, opened 2021-05-30
双重否定改为肯定

7will10:
#5899

solve #5899

format code：double negation in code；
原有代码中if判断使用双重否定改为肯定


KomachiSion (changes requested):
This class is auto-generate from istio proto buf file. We has no need to fix it.

Issue #5941, opened 2021-06-03
Feature searchConfig and fuzzySearchConfig do not return content, reduce network transmission.


CherishCai:
**Is your feature request related to a problem? Please describe.**

Now search the configuration list on the control console, and the configuration content will be returned. If there are 100 pages, there is actually a lot of data. There is no need to return content on the configuration list, and it should be removed to reduce network transmission.

--- 

现在管控台上搜索配置列表，会返回配置内容，若分页100条，其实数据量不少。在配置列表上也无需内容的返回，应该去除减少网络传输。

**Describe the solution you'd like**

You can remove the content field query in the following two database query methods SQL

可以在以下两个数据库查询的方法 SQL 里，去除 content 字段查询
```
persistService.findConfigInfo4Page
persistService.findConfigInfoLike4Page
```

shiyiyue1102:
This is an optimization point that can be considered，but should not change in sql level without changing api. If  we really want to optimize it ,we should create a new clearly declared api without content . In addition, these two api is not frequently invoked ，so we think there is no need to change it now

CherishCai:
Ok, I just submit a feasible suggestion.

PR #6064, opened 2021-06-15
Feat find config exclude content

CherishCai:
feature #5941


KomachiSion:
Please fix CI problem first

KomachiSion:
I see the issue, no discussion for now. So I think PR should be close first. And submit a new one after we discuss how to do or whether need to do.

PR #6118, opened 2021-06-21
[ISSUE#6116] Refactor ClientManager and make register client and schedule checkTask is atomic.

### LINKED ISSUE #6116: Unable to check client beats when register same ip and port in old clients with high concurrent
**Describe the bug**
When users register different services which has same ip and port in old clients and 2.0.x servers with high concurrent, The server may has some problem and cause the instance can't check the beats correctly.

The reason is in `com.alibaba.nacos.naming.core.InstanceOperatorClientImpl#createIpPortClientIfAbsent`. 

containClient, clientConnected and instance IpPortBasedClient is not atomic operation and make the beatCheckTask which async to do is mismatch the right client.

situation like under:
```

TA  containClient --> false --> instance ClientA -----> schedule BeatCheckTaskA -------------------------> clientConnected

TB  containClient --> false -------> instance ClientB ------> schedule BeatCheckTaskB ------> clientConnected

```  

So that clientManager cache the ClientA but schedule BeatCheckTaskB.

**Expected behavior**
expired instance can be removed.

**Acutally behavior**
expired instance can't be removed and set unhealthy.

**How to Reproduce**
1. start any mode server 2.0.x.
2. register same ip and port in old clients with high concurrent.
3. shutdown all instance.
4. check whether instance will be removed. 

 - Version nacos-server 2.0.x + nacos-client 1.X
 - Module naming

**Additional context**

Fix way:

make the instance Client(schedule BeatCheckTask) and register client to manager is atomic operation will be ok.

KomachiSion:

Fixes #6116 .

## Brief changelog

- Refactor ClientManager and make register client and schedule checkTask is atomic.

## Verifying this change

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/core/v2/service/impl/PersistentClientOperationServiceImpl.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/core/v2/service/impl/PersistentClientOperationServiceImpl.java
@@ -38,6 +38,7 @@ import com.alibaba.nacos.core.distributed.ProtocolManager;
 import com.alibaba.nacos.naming.consistency.persistent.impl.AbstractSnapshotOperation;
 import com.alibaba.nacos.naming.core.v2.ServiceManager;
 import com.alibaba.nacos.naming.core.v2.client.Client;
+import com.alibaba.nacos.naming.core.v2.client.ClientAttributes;
 import com.alibaba.nacos.naming.core.v2.client.ClientSyncData;
 import com.alibaba.nacos.naming.core.v2.client.impl.IpPortBasedClient;
 import com.alibaba.nacos.naming.core.v2.client.manager.impl.PersistentIpPortClientManager;
@@ -170,7 +171,10 @@ public class PersistentClientOperationServiceImpl extends RequestProcessor4CP im
     
     private void onInstanceRegister(Service service, Instance instance, String clientId) {
         Service singleton = ServiceManager.getInstance().getSingleton(service);
-        Client client = clientManager.computeIfAbsent(clientId, () -> new IpPortBasedClient(clientId, false));
+        if (!clientManager.contains(clientId)) {
+            clientManager.clientConnected(clientId, new ClientAttributes());
+        }
+        Client client = clientManager.getClient(clientId);
         InstancePublishInfo instancePublishInfo = getPublishInfo(instance);
         client.addServiceInstance(singleton, instancePublishInfo);
         client.setLastUpdatedTime();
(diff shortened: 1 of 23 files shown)


naming/src/main/java/com/alibaba/nacos/naming/core/v2/service/impl/PersistentClientOperationServiceImpl.java
```diff
@@ -170,7 +171,10 @@ public Response onApply(WriteRequest request) {
     
     private void onInstanceRegister(Service service, Instance instance, String clientId) {
         Service singleton = ServiceManager.getInstance().getSingleton(service);
-        Client client = clientManager.computeIfAbsent(clientId, () -> new IpPortBasedClient(clientId, false));
+        if (!clientManager.contains(clientId)) {
```
pixystone, line 174: This `contains` is unnecessary.

naming/src/main/java/com/alibaba/nacos/naming/core/v2/service/impl/PersistentClientOperationServiceImpl.java
```diff
@@ -170,7 +171,10 @@ public Response onApply(WriteRequest request) {
...
+            clientManager.clientConnected(clientId, new ClientAttributes());
```
pixystone, line 175: `ephemeral` is removed but never set at another place.

pixystone (changes requested):

naming/src/main/java/com/alibaba/nacos/naming/core/v2/service/impl/PersistentClientOperationServiceImpl.java
```diff
@@ -289,6 +293,7 @@ protected void loadSnapshot(byte[] snapshotBytes) {
             ConcurrentHashMap<String, IpPortBasedClient> snapshot = new ConcurrentHashMap<>(newData.size());
             for (Map.Entry<String, ClientSyncData> entry : newData.entrySet()) {
                 IpPortBasedClient snapshotClient = new IpPortBasedClient(entry.getKey(), false);
```
pixystone, line 295: Consider using factory ?

pixystone (changes requested):

naming/src/main/java/com/alibaba/nacos/naming/core/v2/service/impl/PersistentClientOperationServiceImpl.java
KomachiSion: Using factory will add multiple dependencies. So I think no neccessary.

naming/src/main/java/com/alibaba/nacos/naming/core/v2/service/impl/PersistentClientOperationServiceImpl.java
KomachiSion: To reduce invalid object create. When the node handle the register persistent instance, the client has been create and register. There is no need to create and try create again.

horizonzy (approved):

PR #6175, opened 2021-06-27
If the client already disconnect, ignore this request to avoid NPE.

horizonzy:
The client may be null after disconnect, fix it.

### DIFF
--- a/common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java
+++ b/common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java
@@ -89,7 +89,7 @@ public abstract class RpcClient implements Closeable {
     protected ClientAbilities clientAbilities;
     
     /**
-     * default keep alive time 10s.
+     * default keep alive time 5s.
      */
     private long keepAliveTime = 5000L;
     
@@ -176,7 +176,6 @@ public abstract class RpcClient implements Closeable {
         this.keepAliveTime = keepAliveTime * timeUnit.toMillis(keepAliveTime);
         LoggerUtils.printIfInfoEnabled(LOGGER, "[{}]RpcClient init keepalive time, keepAliveTimeMillis={}", name,
                 keepAliveTime);
-        
         return this;
     }
     
@@ -706,7 +705,7 @@ public abstract class RpcClient implements Closeable {
             throw (exceptionThrow instanceof NacosException) ? (NacosException) exceptionThrow
                     : new NacosException(SERVER_ERROR, exceptionThrow);
         } else {
-            throw new NacosException(SERVER_ERROR, "Request fail,Unknown Error");
+            throw new NacosException(SERVER_ERROR, "Request fail, unknown Error");
         }
     }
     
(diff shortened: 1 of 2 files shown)


KomachiSion (approved):

naming/src/main/java/com/alibaba/nacos/naming/core/v2/service/impl/EphemeralClientOperationServiceImpl.java
```diff
@@ -49,6 +49,9 @@ public EphemeralClientOperationServiceImpl(ClientManagerDelegate clientManager)
     public void registerInstance(Service service, Instance instance, String clientId) {
         Service singleton = ServiceManager.getInstance().getSingleton(service);
         Client client = clientManager.getClient(clientId);
+        if (null == client || !client.isEphemeral()) {
```
KomachiSion, line 52: abstract a method named `validateClient`?

KomachiSion (changes requested):

naming/src/main/java/com/alibaba/nacos/naming/core/v2/service/impl/EphemeralClientOperationServiceImpl.java
horizonzy: print an warn log, ok. 
abstract method validateClient I think is not necessary, If do it, it will be take over by util class, because need check client is null firstly. So I think it not necessary.

KomachiSion: Why take over by util class? just add one private method in EphemeralClientOperationServiceImpl is ok. it will means that check the client is valid for EphemeralClientOperationServiceImpl.

horizonzy: Because this judgement not at EphemeralClientOperationServiceImpl, other place also be at same case.
https://github.com/alibaba/nacos/blob/e0efe71b67ba60f955fddc462d6b7e1d5783b091/naming/src/main/java/com/alibaba/nacos/naming/consistency/ephemeral/distro/v2/DistroClientDataProcessor.java#L222-L235

KomachiSion: before you add in EphemeralClientOperationServiceImpl, only little place need this logic.
After you add, EphemeralClientOperationServiceImpl has at least four place use this logic and both need to print logs. So I think it's time to abstract a method in EphemeralClientOperationServiceImpl. If there are many place in other class need this method. We should abstract to a util class.

naming/src/main/java/com/alibaba/nacos/naming/core/v2/service/impl/EphemeralClientOperationServiceImpl.java
```diff
@@ -49,6 +49,10 @@ public EphemeralClientOperationServiceImpl(ClientManagerDelegate clientManager)
...
+            Loggers.SRV_LOG.warn("Client connection {} already disconnect", clientId);
```
KomachiSion, line 53: The log maybe dismiss maintainer.
It can't include `!client.isEphemeral()`

KomachiSion (changes requested):

naming/src/main/java/com/alibaba/nacos/naming/core/v2/service/impl/EphemeralClientOperationServiceImpl.java
horizonzy: yep.

Moskalnazar (commented):
wow

common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java
```diff
@@ -89,7 +89,7 @@
     protected ClientAbilities clientAbilities;
     
     /**
-     * default keep alive time 10s.
```
Moskalnazar, line 92: 5

KomachiSion (approved):

PR #6176, opened 2021-06-27
Use same request instance at stateless situation.

horizonzy:
Some request is stateless, so we can reuse it to avoid create new instance.


KomachiSion (changes requested):
There is no much difference between changes.

And use singleton may cause thread safe problem.

Issue #6216, opened 2021-06-30
SQL keyword naming is not uniform


mask616:
I found that in some places the SQL keyword is lowercase, but in some places the SQL keyword is uppercase. Can I unify them?
For example: 
![image](https://user-images.githubusercontent.com/33280931/123964186-c4f41200-d9e5-11eb-9b4a-41897538a5c7.png)

![image](https://user-images.githubusercontent.com/33280931/123964283-d806e200-d9e5-11eb-932a-b5b807fb881e.png)

realJackSun:
Welcome to unify them.

PR #6233, opened 2021-07-01
[ISSUE #6216]All SQL keywords are modified to uppercase

zzq1314zll:
for #6216

PR #6239, opened 2021-07-01
[ISSUE #6216] Unify SQL naming style.

mask616:

For [#6216]

Unify SQL naming style.
### Naming rules:
**SQL keyword : uppercase(eg.INSERT INTO, SELECT, UPDATE, DELETE, FROM, WHERE, LEFT JOIN, AND, LIKE... )**
**SQL function name : lowercase(eg. count, max...)**
**SQL field name: lowercase**
**Table name : lowercase**

### For  example: 
`SELECT  a,b,c FROM table;`
`SELECT count(*) FROM table;`


KomachiSion:
Is this PR duplicated with #6233？

mask616:
> #6233？

No, #6233 just modified the naming of some java source files SQL. This PR unifies the naming of all Java source files SQL in the config module. In addition, I also provide naming rules.But there may be conflicts.

KomachiSion (changes requested):
I found that the `count` is modified by other PR to upper case.

And all of them are modified to lower case. 

Should we unify al of them?

mask616:
> I found that the `count` is modified by other PR to upper case.
> 
> And all of them are modified to lower case.
> 
> Should we unify al of them?

Now, all the `count` is lower case.

KomachiSion (approved):

PR #6265, opened 2021-07-05
[ISSUE #6256] [config module] replace "*commons.lang3*.*Utils" to "co…

zzq1314zll:
for #6256


config/src/main/java/com/alibaba/nacos/config/server/utils/RegexParser.java
```diff
@@ -44,7 +42,7 @@ public static String regexFormat(String regex) {
         result.append("^");
         for (int i = 0; i < regex.length(); i++) {
             char ch = regex.charAt(i);
-            if (CharUtils.isAsciiAlphanumeric(ch) || CharUtils.isAsciiNumeric(ch)) {
+            if (isAsciiAlphanumeric(ch)) {
```
KomachiSion, line 45: Why not move them to CharUtils？

KomachiSion (changes requested):

config/src/main/java/com/alibaba/nacos/config/server/utils/RegexParser.java
zzq1314zll: Only used in this one place,So when I was writing the tool class, I had already planned to deal with it this way.

KomachiSion (approved):

config/src/main/java/com/alibaba/nacos/config/server/utils/RegexParser.java
KomachiSion: OK

Issue #6215, opened 2021-06-30
ServerHttpAgent never retried if long polling failed and it will cause no available server, currentServerAddr error


ijustyce:
**Describe the bug**  

This is a old story relate to `no available server, currentServerAddr ...` here is the code  

<img width="1209" alt="截屏2021-06-30 下午8 22 44" src="https://user-images.githubusercontent.com/6041166/123960496-ef43d080-d9e1-11eb-9648-ed5d85593ab5.png">

I think do-while should change if socketTimeoutException occur then it will have no chance to retry, endtime = now + timeout, if timeout then endtime also arrive and do-while break then the exception throw.

I think should make a change as below:

change  

`final long endTime = System.currentTimeMillis() + readTimeoutMs;`  

to  

  `final long endTime = System.currentTimeMillis() + (readTimeoutMs * retryTimes);`

KomachiSion:
2.0中已经变更为grpc了 httpagent即将移除

PR #6351, opened 2021-07-14
修复客户端请求超时后不重试的问题

ijustyce:
修复 #6215 客户端如果监听配置变更，那么会开启长轮询，长轮询失败后，client 会重试，但是重试时设置的重试超时有缺陷，太小，假设第一次读取超时，此时，do-while 循环就会结束，虽不影响业务，但是会抛出异常，不利于监控。


zrlw (commented):
final long endTime = System.currentTimeMillis() + ConfigHttpClientManager.getInstance().getConnectTimeoutOrDefault(3000) + this.maxRetry * readTimeoutMs;
may be better.

ijustyce:
> final long endTime = System.currentTimeMillis() + ConfigHttpClientManager.getInstance().getConnectTimeoutOrDefault(3000) + this.maxRetry * readTimeoutMs;
> may be better.

Yes, you are right, I also suggest to change the log level to warning if the connection exception occur (line 179), because we will retry later, if the retry fails, nacos will also throw an exception. The error log may be uploaded to sentry, and then I will receive text messages and phone calls. I really don’t want this to happen again just because of a misleading log. Thks again.

Issue #6461, opened 2021-07-26
[Reduce dependence] Replace "google.common.base. Objects" API with jdk API


zzq1314zll:
Replace "google.common.base. Objects" . Sub issue for #6448

mmm0171:
0xa36fD8fF0524f287422337d99bABdf05923f581B

PR #6463, opened 2021-07-26
Replace "google.common.base.Objects" API with jdk API

zzq1314zll:
for  #6461

### DIFF
--- a/api/src/main/java/com/alibaba/nacos/api/naming/pojo/healthcheck/impl/Http.java
+++ b/api/src/main/java/com/alibaba/nacos/api/naming/pojo/healthcheck/impl/Http.java
@@ -21,7 +21,7 @@ import com.alibaba.nacos.api.naming.pojo.healthcheck.AbstractHealthChecker;
 import com.alibaba.nacos.api.utils.StringUtils;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 
-import java.util.Arrays;
+import java.util.Objects;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/alibaba/nacos/api/naming/pojo/healthcheck/impl/Http.java
```diff
@@ -89,7 +89,7 @@ public void setHeaders(final String headers) {
     
     @Override
     public int hashCode() {
-        return Objects.hashCode(path, headers, expectedResponseCode);
+        return Arrays.hashCode(new Object[]{path, headers, expectedResponseCode});
```
KomachiSion, line 92: In `com.google.common.base.Objects` comment, guava recommend 

```
<p><b>Note for Java 7 and later:</b> This method should be treated as deprecated; use {@link
   * java.util.Objects#hash} instead.
```

KomachiSion (changes requested):

api/src/main/java/com/alibaba/nacos/api/naming/pojo/healthcheck/impl/Http.java
mmm0171: 0xa36fD8fF0524f287422337d99bABdf05923f581B

mmm0171 (commented):
0xbbf33a3c83cf86d0965a66e108669d272dfe4214

KomachiSion (approved):

PR #6470, opened 2021-07-27
Add nacos client version 0.8 support

goodjava:
Add nacos client version 0.8 support


KomachiSion:
The 0.8 client is too old, community will do not support it anymore. 
And the `dom` field is removed after nacos1.0 early, add it in 2.0 is unreasonable.

Issue #6497, opened 2021-07-28
[Reduce dependence] Replace "google.common.base.Joiner" API with jdk API 


zzq1314zll:
Replace "google.common.base. Joiner" . Sub issue for #6448

PR #6501, opened 2021-07-28
Replace "google.common.base.Joiner" API with jdk API

zzq1314zll:
Replace "google.common.base.Joiner" API with jdk API.
for #6497


KomachiSion:
Test failed. Need Fix your change.

KomachiSion (approved):

Issue #6498, opened 2021-07-28
[Reduce dependence] Replace "google.common.base.Preconditions" API with jdk API


zzq1314zll:
Replace "google.common.base.Preconditions" . Sub issue for #6448

PR #6502, opened 2021-07-29
replace "google.common.base.Preconditions" API with jdk API

zzq1314zll:
replace "google.common.base.Preconditions" API with jdk API,
for #6498


common/src/main/java/com/alibaba/nacos/common/utils/Preconditions.java
```diff
@@ -0,0 +1,54 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.alibaba.nacos.common.utils;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import static com.google.common.base.Strings.lenientFormat;
```
KomachiSion, line 21: still depend guava?

KomachiSion (changes requested):
It seems that full copy from guava's Preconditions?

common/src/main/java/com/alibaba/nacos/common/utils/Preconditions.java
KomachiSion: I think the Nullable is no need, we can judge in method.

And the guava depend can be refactor.

zzq1314zll:
No, 

> It seems that full copy from guava's Preconditions?

NO! 我仅仅是借鉴，没有有copy

KomachiSion (approved):

Issue #5095, opened 2021-03-12
Add unit tests for module nacos-common


JackSun-Developer:
This is a sub-issue of [ISSUE #4981]
The nacos-common module coverage rate is too low, this ISSUE aims to adding more unit tests to nacos-common module.

SunJiFengPlus:
Please assign it to me, I will solve it.

SunJiFengPlus:
At present, [the test coverage of common module](https://htmlpreview.github.io/?https://raw.githubusercontent.com/SunJiFengPlus/nacos/common-coverage/docs/index.html)

PR #6526, opened 2021-08-02
[ISSUE #5095] Add unit test for RpcClient

SunJiFengPlus:

The nacos-common in nacos 2.0 module coverage rate is too low.
For #5095

## Brief changelog

- remove unnecessary public
- add powermock
- add unit test to RpcClient
- replace explicit type argument with <>
-  replace anonymous class with lambda
- remove redundant initializer
- format comment & log and fix typo

## Verifying this change
[before coverage](https://htmlpreview.github.io/?https://raw.githubusercontent.com/SunJiFengPlus/nacos/common-coverage/docs/com.alibaba.nacos.common.remote.client/.classes/RpcClient.html)
[after coverage](https://htmlpreview.github.io/?https://raw.githubusercontent.com/SunJiFengPlus/nacos/common-coverage/common-coverage/common-current/ns-10/sources/source-2.html)

### DIFF
--- a/common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java
+++ b/common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java
@@ -284,7 +284,7 @@ public abstract class RpcClient implements Closeable {
         // connection event consumer.
         clientEventExecutor.submit(() -> {
             while (!clientEventExecutor.isTerminated() && !clientEventExecutor.isShutdown()) {
-                ConnectionEvent take = null;
+                ConnectionEvent take;
                 try {
                     take = eventLinkedBlockingQueue.take();
                     if (take.isConnected()) {
@@ -504,7 +504,7 @@ public abstract class RpcClient implements Closeable {
             
             int reConnectTimes = 0;
             int retryTurns = 0;
-            Exception lastException = null;
+            Exception lastException;
             while (!switchSuccess && !isShutdown()) {
                 
                 //1.get a new server
@@ -630,7 +630,7 @@ public abstract class RpcClient implements Closeable {
      */
     public Response request(Request request, long timeoutMills) throws NacosException {
         int retryTimes = 0;
-        Response response = null;
+        Response response;
         Exception exceptionThrow = null;
         long start = System.currentTimeMillis();
         while (retryTimes < RETRY_TIMES && System.currentTimeMillis() < timeoutMills + start) {
@@ -902,7 +902,7 @@ public abstract class RpcClient implements Closeable {
     private ServerInfo resolveServerInfo(String serverAddress) {
         String property = System.getProperty("nacos.server.port", "8848");
         int serverPort = Integer.parseInt(property);
-        ServerInfo serverInfo = null;
+        ServerInfo serverInfo;
         if (serverAddress.contains(Constants.HTTP_PREFIX)) {
             String[] split = serverAddress.split(Constants.COLON);
             String serverIp = split[1].replaceAll("//", "");
--- a/common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java
+++ b/common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java
@@ -46,7 +46,6 @@ import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
-import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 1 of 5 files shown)


KomachiSion:
CI failed, please check and fix

SunJiFengPlus:
> CI failed, please check and fix

The unit test in ```ClientWorkerTest``` uses the exception message from ```RpcClient```. I formatted the exception message, and then the unit test of ```ClientWorkerTest``` failed. Now fix completed.

common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java
```diff
@@ -275,110 +278,101 @@ public final void start() throws NacosException {
...
+                                LoggerUtils.printIfInfoEnabled(LOGGER,
+                                        "[{}] Server healthy check fail, currentConnection = {}", name,
+                                        currentConnection.getConnectionId());
+                                
+                                RpcClientStatus rpcClientStatus = RpcClient.this.rpcClientStatus.get();
+                                if (RpcClientStatus.SHUTDOWN.equals(rpcClientStatus)) {
+                                    break;
+                                }
+                                
+                                boolean success1 = RpcClient.this.rpcClientStatus
```
KomachiSion, line 330: why rename to `success1`?

common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java
```diff
@@ -913,25 +904,17 @@ protected ServerInfo currentRpcServer() {
...
-            }
-            serverInfo = new ServerInfo(serverIp, serverPort);
-        } else {
-            String[] split = serverAddress.split(Constants.COLON);
-            String serverIp = split[0];
-            if (split.length > 1 && StringUtils.isNotBlank(split[1])) {
-                serverPort = Integer.parseInt(split[1]);
-            }
-            serverInfo = new ServerInfo(serverIp, serverPort);
+        Matcher matcher = EXCLUDE_PROTOCOL_PATTERN.matcher(serverAddress);
```
KomachiSion, line 907: Why change this logics?

Your Pr is add unit test. Not refactor.

pom.xml
```diff
@@ -632,6 +633,18 @@
             <artifactId>mockito-inline</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
KomachiSion, line 636: Why add powermock?

KomachiSion (changes requested):

common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java
SunJiFengPlus: i rename it to ```statusFlowSuccess```

common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java
SunJiFengPlus: already dropped

pom.xml
SunJiFengPlus:
```java
        clientEventExecutor = new ScheduledThreadPoolExecutor(2, r -> {
            Thread t = new Thread(r);
            t.setName("com.alibaba.nacos.client.remote.worker");
            t.setDaemon(true);
            return t;
        });
```

```ScheduledThreadPoolExecutor::new``` cannot be mock by mockito, so i use PowerMock to mock the return value of ```ScheduledThreadPoolExecutor::new```.

KomachiSion: use reflection to get clientEventExecutor and replace it with mock `ScheduledThreadPoolExecutor`

KomachiSion: powermock will change the compiled file may cause some unexpected problem.

SunJiFengPlus: In the test case, I can mock the ```clientEventExecutor``` and assign it to the ```clientEventExecutor``` through reflection. However, in the run of ```Rpcclient:: start``` method, it will reassign a value to the ```clientEventExecutor``` which will overwrite the ```clientEventExecutor``` of the previous mock.

The testability of this method is low. I didn't think of a better way to test it. Should i delete all test cases related to this method? Or put the initialization of ```clientEventExecutor``` out of the method for testability?

KomachiSion:
I think this PR should be closed.

The PR do much things, If you want to add unit test, you should only add unit test as much as possible. And do no change any line for RpcClient.

But you change many lines of RpcClient and add more dependency.

So I think you should split this PR to three .
the first is add unit test as much as possible without more dependency.
then do some refactor for RpcClient
finally if the unit test still hard to do, we discuss whether add new dependency.

SunJiFengPlus:
> I think this PR should be closed.
> 
> The PR do much things, If you want to add unit test, you should only add unit test as much as possible. And do no change any line for RpcClient.
> 
> But you change many lines of RpcClient and add more dependency.
> 
> So I think you should split this PR to three .
> the first is add unit test as much as possible without more dependency.
> then do some refactor for RpcClient
> finally if the unit test still hard to do, we discuss whether add new dependency.

 I agree with you

PR #6537, opened 2021-08-03
fix #1733 again

### LINKED ISSUE #1733: still get the instance after register twice and deregister
## Issue Description
i use namingService to register a instance, somehow i register a same instance twice, after i degister this instance, this instance is still in nacos
Type: bug report

### Describe what happened (or what feature you want)
when i debug the program, I found that after i degister this instance, i can see the beatTask is removed from    dom2Beat map, but i found there is still a beatTask running to send beatInfo to nacos server, as i see, every time register a instance will cause a new BeatInfo created, and this new BeatInfo will override the old one in dom2beat map, but the old one is not cancelled, so when i deregister this instance, the old beatTask still send beatInfo to nacos server,this will cause register again.

image

image

### Describe what you expected to happen
when add beatInfo , we should consider if there is already exist one, and cancel it

zrlw:
fix #1733 again to avoid creating beat task more than once by doing ConcurrentHashMap.put concurrently.
i met the issue again by using nacos sync:
1. restart N instances of one sync task at the same time,  it trigged sync event more than once,
2. for each sync event, nacos sync called registerInstance concurrently, 
3. ConcurrentHashMap.put  was called concurrently for the same instance.

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/naming/beat/BeatReactor.java
+++ b/client/src/main/java/com/alibaba/nacos/client/naming/beat/BeatReactor.java
@@ -96,11 +96,10 @@ public class BeatReactor implements Closeable {
         NAMING_LOGGER.info("[BEAT] adding beat: {} to beat map.", beatInfo);
         String key = buildKey(serviceName, beatInfo.getIp(), beatInfo.getPort());
         BeatInfo existBeat;
-        //fix #1733
-        if ((existBeat = dom2Beat.remove(key)) != null) {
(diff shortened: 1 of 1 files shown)


KomachiSion (changes requested):
BeatReactor has no used after 2.0.0.

You may submit Pr to 1.X-develop branch.

zrlw:
> BeatReactor has no used after 2.0.0.
> 
> You may submit Pr to 1.X-develop branch.

ok， i submit a new PR #6557

PR #6566, opened 2021-08-05
merge service change event publish operation

### LINKED ISSUE #6565: [nacos 1.x] service change event merge algorithm lost efficacy when multiple instances of the same service changed at the same time
**Describe the bug**
when multiple instances of the same Service changed at the same time,  service change event merge algorithm  of Nacos 1.x lost efficacy because futureMap had nothing yet.
it will cause the main process almost hang if too many events of the same serivce are not merged at serviceChange().

zrlw:
fix #6565

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/push/PushService.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/push/PushService.java
@@ -48,7 +48,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.zip.GZIPOutputStream;
@@ -85,7 +84,9 @@ public class PushService implements ApplicationContextAware, ApplicationListener
     
     private static DatagramSocket udpSocket;
     
-    private static ConcurrentMap<String, Future> futureMap = new ConcurrentHashMap<>();
+    private static ConcurrentMap<String, Integer> eventPublishedMap = new ConcurrentHashMap<>();
+    
+    private static final int EVENT_PUBLISHED = 1;
     
     static {
         try {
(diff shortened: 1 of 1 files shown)


naming/src/main/java/com/alibaba/nacos/naming/push/PushService.java
```diff
@@ -122,64 +123,66 @@ public void onApplicationEvent(ServiceChangeEvent event) {
         String serviceName = service.getName();
         String namespaceId = service.getNamespaceId();
         
-        Future future = GlobalExecutor.scheduleUdpSender(() -> {
```
KomachiSion, line 125: The future is useful, can't be removed.

KomachiSion (changes requested):

naming/src/main/java/com/alibaba/nacos/naming/push/PushService.java
zrlw: future
why?

KomachiSion: I mean, the future is been used to merge change event. No need to do this change.

KomachiSion (changes requested):

naming/src/main/java/com/alibaba/nacos/naming/push/PushService.java
zrlw: servcie change event merge alogrithm is changed to ConcurrentMap<String, Long> that has nothing to do with future.
please take a full look at all changes.

zrlw: we have more than 100 instances per service,  all the instances will be started or  restarted by one touch click on our control system, the sevice merge mechanism using Future will be broken: all the futureMap.contains() calling almost return false because the future is not been put into the futureMap.
we brought the merge mechanism forward to publish event stage by doing theses changes, the push frequency was reduced eventually.
@KomachiSion

KomachiSion: the target of futureMap.contains() is to merge the frequently changes for same service. the future will be create even you ignore the `return`. So Why don't you use it but use a new `Long`? I has no idea the benefits of your changes.

zrlw (commented):
when the futureMap.containsKey() is called at serviceChanged(), the future is not created.
it just like:
```
void a() { if (i==0) b(); }
void b() { i++ };
```
it's ok if only one thread run it, but if more than one thread, it result will be wrong.

naming/src/main/java/com/alibaba/nacos/naming/push/PushService.java
KomachiSion: `onApplicationEvent` is a single thread I remember. spring event system will keep the event orders.

KomachiSion (changes requested):

naming/src/main/java/com/alibaba/nacos/naming/push/PushService.java
zrlw: serviceChanged() is not running at single thread:
Service.onChange -> updateIPs ->> PushService.serviceChanged is not running  at single thread mode, if more than one instances of the same service start up at the same time, more than one serviceChangeEvent of the same service might be created and published to ApplicationContext, all the events will be processed by onApplicationEvent()  successively.

zrlw (commented):
but serviceChanged() is not running at single thread

naming/src/main/java/com/alibaba/nacos/naming/push/PushService.java
KomachiSion: I think there is no problem, If you think it will cause your problem, should we double check futureMap in `onApplicationEvent` rather than do much in `serviceChanged`?

KomachiSion (changes requested):

zrlw (commented):
we have 400+ services, each service has more than 100 instances on one zone, when start all services of one zone at the same time, the event number might peak to 40000, it will cause many problems: nacos service slow down, service register slowly and so on. 
our test result showed reducing event number is matter.

KomachiSion:
> we have 400+ services, each service has more than 100 instances on one zone, when start all services of one zone at the same time, the event number might peak to 40000, it will cause many problems: nacos service slow down, service register slowly and so on.
> our test result showed reducing event number is matter.

restart all instance and service at the same times? without batches? Is it function service situation? If not, maybe you need change the usage or upgrade your nodes of nacos.

Of cource the codes can enhance, but I think we can double check in `onApplicationEvent` not in `serviceChange`

KomachiSion:
> we have 400+ services, each service has more than 100 instances on one zone, when start all services of one zone at the same time, the event number might peak to 40000, it will cause many problems: nacos service slow down, service register slowly and so on.
> our test result showed reducing event number is matter.

restart all instance and service at the same times? without batches? Is it function service situation? If not, maybe you need change the usage or upgrade your nodes of nacos.

Of cource the codes can enhance, but I think we can double check in `onApplicationEvent` not in `serviceChange`

zrlw:
we have a control manage system to switch zones or disaster switching by one button click.
if you tested the code with 20000+ serviceChangeEvent, you might understand why we put the merge mechanism forward to serviceChange(), it will cause the service almost hang.

zrlw (commented):
By default, spring event handles run at sync mode, the applicationContext.publishEvent() calling waits until onApplicationContext() return, it will block the main process if too many events published.
it's might be better to change the event handle mode from sync to async.

zrlw:
if using spring event, the event handler should run at async mode to avoid blocking the main process.
if using theadpool, there are no need to use spring event mechanism.

zrlw (commented):
test result showed that service subscribes receive udp push data quickly than before.

KomachiSion:
You can do this for yourself. But for community, submit PR before discussion issue is illegal.

From your problem, I think do double check in `onApplicationEvent` is enough. the `onApplicationEvent` only create a async task and cache the future, which will cost much time and the event bus is a single thread. So even you has thousands events, the handler will not cost much time and it also can slow down push.

You told that `serviceChange` will be call async many time and make do much push make your node unavailable, it shows that the `onApplicationEvent` has no performance problem. 

So only double check in `onApplicationEvent` is best way to enhance this problem.

zrlw:
maybe i didn't write clearly,  what i mean is:
```
it will cause the service almost hang if the events of the same serivce are not merged at serviceChange().
```
Frankly,  we might confuse the difference between PR and issue, 
we just want to remain the problem that we met and provide what we have done that were proved by our testing.
that's all.

KomachiSion:
> maybe i didn't write clearly, what i mean is:
> 
> ```
> it will cause the service almost hang if the events of the same serivce are not merged at serviceChange().
> ```
> 
> Frankly, we might confuse the difference between PR and issue,
> we just want to remain the problem that we met and provide what we have done that were proved by our testing.
> that's all.

So you need to submit issue first to describe the problem is and why this problem cause. Based the reason of problem, discuss the solution in issue. Finally submit the PR the implement the solution.

You submit two PR but I don't think the solutions is right.

KomachiSion:
I will close PR first and please submit issue to describe your problem and the reason clearly.

PR #6575, opened 2021-08-05
When read-index read timeout, go to leader to read.

horizonzy:
detail #6583 

When use raft read-index, the follower will wait self apply index reach leader's commitIndex. If follower fall behind leader much, the read-index read will timeout(default timeout is 2000ms).

If read-index read timeout, request to leader directly.


KomachiSion:
Can you submit an issue to describe the situation first?

horizonzy:
> Can you submit an issue to describe the situation first?

fine.

KomachiSion (approved):

PR #6576, opened 2021-08-05
[ISSUE #6572] add client info for OperatorController metrics interface

### LINKED ISSUE #6572: [nacos 2.0.3] add client info for OperatorController metrics interface
image
There are only service and instance info in  OperatorController metrics interface.
However, if nacos server run in 2.X mode, and all the clients use nacos-client 2.X, the field `responsibleServiceCount` and `responsibleInstanceCount` is always zero.
So I want to add client info for this interface, include :
image
After this change, we can get client info from this interface. 
For example, if we want to shutdown nacos-server gracefully,  all the client should switch to other nacos-server node at first,  and the filed `responsibleClientCount` will be zero. After that we can safely shutdown this server.

MajorHe1:

fix #6572 

## Brief changelog

XX

## Verifying this change

XXXX

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/core/v2/client/manager/ClientManagerDelegate.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/core/v2/client/manager/ClientManagerDelegate.java
@@ -16,6 +16,7 @@
 
 package com.alibaba.nacos.naming.core.v2.client.manager;
 
+import com.alibaba.nacos.naming.constants.ClientConstants;
 import com.alibaba.nacos.naming.core.v2.client.Client;
 import com.alibaba.nacos.naming.core.v2.client.ClientAttributes;
 import com.alibaba.nacos.naming.core.v2.client.impl.IpPortBasedClient;
@@ -103,7 +104,7 @@ public class ClientManagerDelegate implements ClientManager {
         if (isConnectionBasedClient(clientId)) {
             return connectionBasedClientManager;
         }
-        return clientId.endsWith(IpPortBasedClient.PERSISTENT_SUFFIX) ? persistentIpPortClientManager : ephemeralIpPortClientManager;
+        return clientId.endsWith(ClientConstants.PERSISTENT_SUFFIX) ? persistentIpPortClientManager : ephemeralIpPortClientManager;
     }
     
     private boolean isConnectionBasedClient(String clientId) {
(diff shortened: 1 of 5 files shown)


naming/src/main/java/com/alibaba/nacos/naming/core/v2/client/manager/ClientManagerDelegate.java
```diff
@@ -105,7 +103,7 @@ private ClientManager getClientManagerById(String clientId) {
         if (isConnectionBasedClient(clientId)) {
             return connectionBasedClientManager;
         }
-        return clientId.endsWith(SUFFIX) ? persistentIpPortClientManager : ephemeralIpPortClientManager;
+        return clientId.endsWith(IpPortBasedClient.PERSISTENT_SUFFIX) ? persistentIpPortClientManager : ephemeralIpPortClientManager;
```
KomachiSion, line 106: No need to move `SUFFIX` to `IpPortBasedClient`

KomachiSion (changes requested):

naming/src/main/java/com/alibaba/nacos/naming/core/v2/client/manager/ClientManagerDelegate.java
MajorHe1: If we don't move `SUFFIX` to `IpPortBasedClient`, we should declare another variable named `SUFFIX` in `OperatorController` to judge whether `clientId` is persistent clientId.
So I think `PERSISTENT_SUFFIX` is a property of `IpPortBasedClient`.
Otherwise, we should change `SUFFIX` from private  to public, in order `OperatorController` can reference it.

KomachiSion: Can we move it to ClientConstants?

KomachiSion: And replace all `false` with same meaning usage.

MajorHe1: Move it to ClientConstants is a good choice.
It seems that persistent clientId judgement is only occurs in `ClientManagerDelegate` and `OperatorController` so far.

MajorHe1: Should we move `ID_DELIMITER` from `IpPortBasedClient` to `ClientConstants` too?

KomachiSion (approved):

PR #6615, opened 2021-08-10
remove event mechanism from push service

### LINKED ISSUE #6565: [nacos 1.x] service change event merge algorithm lost efficacy when multiple instances of the same service changed at the same time

zrlw:
fix #6565 
another fix method  for comparison with #6566


KomachiSion:
duplicate with #6566

PR #6629, opened 2021-08-11
complete the warn log

secbr:

the warn log is incomplete, because it loses the event value.

## Brief changelog

add the "event" value to the LOGGER.warn method.


KomachiSion (changes requested):
Your commit log has some problem, Please fix

secbr (commented):
complete the warn log when the event loses.

KomachiSion:
Your commit log still has some problem. If you don't know how to fix. Please checkout new branch from develop branch and do modification.

PR #6701, opened 2021-08-19
optimize the HashMap’s initialCapacity

secbr:

optimize the HashMap’s initialCapacity

## Brief changelog

HashMap will automatically expand, when it‘s size greater than initialCapacity * 0.75.  And the initialCapacity’s value should be 2^n.

The following code will expand：

（1）NamingHttpClientProxy#registerService method：

The size of "params"  will be 13 or 16. Because the reqApi method will put 2 or 5 value into the it. So the initialCapacity should be 32.

(2)NamingHttpClientProxy#deregisterService method:

The size of "params"  will be 8 or 11. So the initialCapacity should be 16.

(3)NamingHttpClientProxy#updateInstance method:

The size of "params"  will be 12 or 15. So the initialCapacity should be 32.

(4)NamingHttpClientProxy#queryInstancesOfService method:

The size of "params"  will be 8 or 11. So the initialCapacity should be 16.

(5)NamingHttpClientProxy#queryService method:

The size of "params"  will be 5 or 8. So the initialCapacity should be 16.

(6)NamingHttpClientProxy#createService method:

The size of "params"  will be 8 or 11. So the initialCapacity should be 16.

(7)NamingHttpClientProxy#deleteService method:

The size of "params"  will be 5 or 8. So the initialCapacity should be 16.

(8)NamingHttpClientProxy#updateService method:

The size of "params"  will be 8 or 11. So the initialCapacity should be 16.

(9)NamingHttpClientProxy#sendBeat method:

The size of "params"  will be 7 or 10. So the initialCapacity should be 16.

(10)NamingHttpClientProxy#serverHealthy method:

The size of "params"  will be 3 or 6. So the initialCapacity should be 8.

(11)NamingHttpClientProxy#getServiceList method:

The size of "params"  will be 7 or 10. So the initialCapacity should be 16.


KomachiSion:
I think only the last commit has effects, can you checkout new branch from `develop` branch  and submit again?

Or learn to use `git rebase`  to fix the commit logs

PR #6710, opened 2021-08-20
[ISSUE #6693] Fix the group name send in heartbeats must be DEFAULT_GROUP

### LINKED ISSUE #6693: nacos 2.x open api: register service and send heartbeats must use DEFAULT_GROUP as groupName
**Describe the bug**
Here is the situation.

1. I started nacos-server-2.0.0 on my PC;
2. I use nacos open api to register a service;
3. I keep sending heartbeat to keep the service healthy (every three seconds and using open api).

**Expected behavior**

The service I registered keeps a health status of true in nacos.

**Acutally behavior**

The problem is, if I use a groupName other than DEFAULT_GROUP when I register the service and send heartbeats, the service's health status will be sometimes true and sometimes false. If I use the groupName DEFAULT_GROUP exactly, the problem wouldn't occur. 

**How to Reproduce**

I was actually use nacos-python-sdk when I found this issue, but to make this issue clearer, I used curl to request nacos open apis and reproduced this issue.

Here I wrote a srcipt `test_nacos_open_api.sh` to reproduce this issue:

```
#! /bin/bash
if [ $# -ne 2 ]
then
    echo -e "Usage: $(basename $0) service_name group_name\n"
    exit 1
fi

echo -e "register service: serviceName: $1, groupName: $2\n"
curl -X POST "http://localhost:8848/nacos/v1/ns/instance?username=nacos&password=nacosb&ip=10.1.2.75&port=18000&serviceName=$1&weight=1&enable=True&healthy=True&clusterName=DEFAULT&ephemeral=True&groupName=$2&metadata=%7B%7D&namespaceId=public"

while sleep 3;
do
  echo -e "send heartbeat: serviceName: $1, groupName: $2\n"
  curl -X PUT "http://localhost:8848/nacos/v1/ns/instance/beat?serviceName=$1&beat=%7B%22serviceName%22%3A+%22$1%22%2C+%22ip%22%3A+%2210.1.2.75%22%2C+%22port%22%3A+18000%2C+%22weight%22%3A+1.0%2C+%22ephemeral%22%3A+true%2C+%22cluster%22%3A+%22DEFAULT%22%7D&groupName=$2&namespaceId=public&username=nacos&password=nacos"
done
```

You just need to start nacos-server-2.0.0 (and above version), then run these commands in a bash terminal:

```
./test_nacos_open_api.sh test-service ANOTHER_GROUP
./test_nacos_open_api.sh test-service DEFAULT_GROUP
```

The first command will reproduce this issue and the second command should work fine.

I also tested on nacos-server-1.4.2, and this problem does not occur.

 - OS: Mac OS 11
 - Version nacos-server-2.0.0 nacos-server-2.0.3
 - Module naming
 - SDK original

**Additional context**

用中文再简单说一下，问题在 nacos 2.x 的 open api 上，使用 open api 注册服务并持续发送心跳（每3秒发一次），如果传的 groupName 不是 DEFAULT_GROUP，那么服务健康状态一会儿是好的，一会儿是坏的。groupName 传DEFAULT_GROUP时，在各个版本上都没有问题，服务一直都是好的。

在 nacos-server-1.4.2 上测试了没有出现这个问题。

上面给的脚本可以很容易复现这个问题。

zarkzheng:

Fix #6693 
If `serviceName` in beat send with open api is not `group@@serviceName`, then it will always be `DEFAULT_GROUP@@serviceName`.

## Brief changelog
- `ServiceNameFilter` replace the service name with `groupName@@serviceName`, so the `serviceName` parameter in beat request is also this format;
```java
// com.alibaba.nacos.naming.controllers.InstanceController.beat
// the serviceName must be groupName@@serviceName
String serviceName = WebUtils.required(request, CommonParams.SERVICE_NAME);
NamingUtils.checkServiceNameFormat(serviceName);
```
- In addition, sending the heartbeat data of another service in the request has no practical meaning, so 
just set the serviceName of `clientBeat` to `serviceName`, even if other service name or group name may specified in `beat`.
```java
if (clientBeat != null) {
    clientBeat.setServiceName(serviceName);
}
```
- And, should consider the situation where the instance cannot be obtained:
```java
// com.alibaba.nacos.naming.healthcheck.heartbeat.ClientBeatProcessorV2.run
HealthCheckInstancePublishInfo instance = (HealthCheckInstancePublishInfo) client.getInstancePublishInfo(service);
if (instance != null && instance.getIp().equals(ip) && instance.getPort() == port) {
    ...
}
```

## Verifying this change
Heartbeat send with open api will keep your instance health, regardless of whether this instance is in DEFAULT_GROUP.

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/controllers/InstanceController.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/controllers/InstanceController.java
@@ -393,6 +393,10 @@ public class InstanceController {
         String namespaceId = WebUtils.optional(request, CommonParams.NAMESPACE_ID, Constants.DEFAULT_NAMESPACE_ID);
         String serviceName = WebUtils.required(request, CommonParams.SERVICE_NAME);
         NamingUtils.checkServiceNameFormat(serviceName);
+        // fix https://github.com/alibaba/nacos/issues/6693
+        if (clientBeat != null) {
+            clientBeat.setServiceName(serviceName);
+        }
         Loggers.SRV_LOG.debug("[CLIENT-BEAT] full arguments: beat: {}, serviceName: {}, namespaceId: {}", clientBeat,
                 serviceName, namespaceId);
         BeatInfoInstanceBuilder builder = BeatInfoInstanceBuilder.newBuilder();
--- a/naming/src/main/java/com/alibaba/nacos/naming/healthcheck/heartbeat/ClientBeatProcessorV2.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/healthcheck/heartbeat/ClientBeatProcessorV2.java
@@ -57,7 +57,7 @@ public class ClientBeatProcessorV2 implements BeatProcessor {
         String groupName = NamingUtils.getGroupName(rsInfo.getServiceName());
         Service service = Service.newService(namespace, groupName, serviceName, rsInfo.isEphemeral());
         HealthCheckInstancePublishInfo instance = (HealthCheckInstancePublishInfo) client.getInstancePublishInfo(service);
-        if (instance.getIp().equals(ip) && instance.getPort() == port) {
+        if (instance != null && instance.getIp().equals(ip) && instance.getPort() == port) {
             if (Loggers.EVT_LOG.isDebugEnabled()) {
                 Loggers.EVT_LOG.debug("[CLIENT-BEAT] refresh beat: {}", rsInfo);
             }


KomachiSion:
If you send heartbeat with openAPI by yourself, please send with `group` in `groupName` parameter or format it in `serviceName` parameter. otherwise server will think you use `DEFAULT_GROUP` default.

zarkzheng:
Did you really look at this issue and my PR carefully?

zarkzheng:
@KomachiSion Here are my reasons:
1. The `beat` parameter in heartbeat request, which corresponding entity class is `RsInfo`, it doesn't even have a `groupName` field. The `groupName` parameter in the outer layer of `beat` has no effect on it at all. The only way you can set `groupName` for `beat` is to set its `serviceName` to `groupName@@serviceName`.
https://github.com/alibaba/nacos/blob/5e545c8d379d24a0576e067bf21d86b4039d39a4/naming/src/main/java/com/alibaba/nacos/naming/healthcheck/RsInfo.java#L28

2. Isn't it weird that the `serviceName` or `groupName` parameter has already been set in the request,  and you have to declare it again in `beat`?

3. For this code :
https://github.com/alibaba/nacos/blob/7a7f267ee9bc0e1a447d3e88cd11d2145c77b498/naming/src/main/java/com/alibaba/nacos/naming/healthcheck/heartbeat/ClientBeatProcessorV2.java#L58-L60
This `instance` will be `null` if you don't pass the parameters correctly. Doesn't it need special treatment?

PR #6824, opened 2021-09-08
remove the if statement

secbr:

The if statement is optional in slf4j. So remove them.

## Brief changelog

remove the if statement.


KomachiSion (changes requested):
The if statement is useful, please learn more about the logger.

PR #6987, opened 2021-09-26
Fix response success

horizonzy:
Now the response `success` is false, the logic is not right.

If remote processor response `success` is false, we should judge the value before future.complete.

Now the judgement is in whenComplete, it's unuseful.


chuntaojun (approved):

chuntaojun:
Thank you for finding and fixing my low-level mistakes

chuntaojun:
LGTM

PR #7012, opened 2021-09-29
fix https://github.com/alibaba/nacos/issues/6876

Cczzzz:
#6876


Cczzzz:
> you need to solve the CLA first.

How can i fix it ，the CLA

KomachiSion:
There is no response and changes from author, the main problem is cla not sign. So close PR first.

PR #7105, opened 2021-10-20
[ISSUE #4639] Fix Unlimited log files in log4j2 scenerio of nacos-client

zcz3313:
Fix Unlimited log files in log4j2 scenerio

to fix this [issue#4639](https://github.com/alibaba/nacos/issues/4639)
I think these log files should not be unlimited.

## Brief changelog

change rotation strategy of log4j2 to avoid too many log files


KomachiSion:
`<DefaultRolloverStrategy max="${sys:JM.LOG.RETAIN.COUNT:-7}"/>`
So the log will not unlimited. If your log is unlimited, there must be some custom config overwrite nacos log config.

PR #7165, opened 2021-11-02
Not publish server list change event when the servers is not change.

horizonzy:
Not publish server list change event when the servers is not change.


KomachiSion (approved):
CI need redo

PR #7248, opened 2021-11-17
Fix update task can't stop after unsubscribe.

horizonzy:
In final code block, it will execute self again.  Although service is not subscribed again.


KomachiSion (changes requested):
This change has problem.

After the task cancel, and not schedule again, the `futureMap` still contain the future and serviceKey, which may cause subscribe this services again work wrong.

horizonzy:
maybe not, see com.alibaba.nacos.client.naming.core.ServiceInfoUpdateService#stopUpdateIfContain

KomachiSion (approved):
ok

PR #7229, opened 2021-11-13
[ISSUE #7225]  Enhance the constants for client module

onewe:
Please do not create a Pull Request without creating an issue first.

## What is the purpose of the change

#7225 
Enhance the constants for client module

## Brief changelog

- Create a Constants.java
- Delete the IdentifyConstants.java
- Create a ConfigConstants.java in the package scope
- Enhance all the constants and follow the c3p

## Verifying this change

XXXX

Follow this checklist to help us incorporate your contribution quickly and easily:

* [x] Make sure there is a Github issue filed for the change (usually before you start working on it). Trivial changes like typos do not require a Github issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.
* [x] Format the pull request title like `[ISSUE #123] Fix UnknownException when host config not exist`. Each commit in the pull request should have a meaningful subject line and body.
* [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
* [x] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/alibaba/nacos/tree/master/test).
* [x] Run `mvn -B clean package apache-rat:check findbugs:findbugs -Dmaven.test.skip=true` to make sure basic checks pass. Run `mvn clean install -DskipITs` to make sure unit-test pass. Run `mvn clean test-compile failsafe:integration-test`  to make sure integration-test pass.

PR #7255, opened 2021-11-18
[ISSUE #7229] Optimize the constants to the config package for the client module. Part 1

onewe:

sub pr #7229 

* [x ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/config/filter/impl/ConfigRequest.java
+++ b/client/src/main/java/com/alibaba/nacos/client/config/filter/impl/ConfigRequest.java
@@ -22,11 +22,11 @@ import com.alibaba.nacos.api.config.filter.IConfigRequest;
 import java.util.HashMap;
 import java.util.Map;
 
-import static com.alibaba.nacos.client.config.common.ConfigConstants.CONTENT;
-import static com.alibaba.nacos.client.config.common.ConfigConstants.DATA_ID;
-import static com.alibaba.nacos.client.config.common.ConfigConstants.GROUP;
-import static com.alibaba.nacos.client.config.common.ConfigConstants.TENANT;
-import static com.alibaba.nacos.client.config.common.ConfigConstants.TYPE;
+import static com.alibaba.nacos.client.config.filter.impl.ConfigMetadata.CONTENT;
+import static com.alibaba.nacos.client.config.filter.impl.ConfigMetadata.DATA_ID;
+import static com.alibaba.nacos.client.config.filter.impl.ConfigMetadata.GROUP;
+import static com.alibaba.nacos.client.config.filter.impl.ConfigMetadata.TENANT;
+import static com.alibaba.nacos.client.config.filter.impl.ConfigMetadata.TYPE;
 
 /**
  * Config Request.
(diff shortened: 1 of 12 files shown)


client/src/main/java/com/alibaba/nacos/client/config/filter/impl/ConfigRequest.java
```diff
@@ -22,11 +22,11 @@
 import java.util.HashMap;
 import java.util.Map;
 
-import static com.alibaba.nacos.client.config.common.ConfigConstants.CONTENT;
-import static com.alibaba.nacos.client.config.common.ConfigConstants.DATA_ID;
-import static com.alibaba.nacos.client.config.common.ConfigConstants.GROUP;
-import static com.alibaba.nacos.client.config.common.ConfigConstants.TENANT;
-import static com.alibaba.nacos.client.config.common.ConfigConstants.TYPE;
+import static com.alibaba.nacos.client.config.filter.impl.ConfigMetadata.CONTENT;
```
KomachiSion, line 25: Why use new one? I think the old one has no problem.

client/src/main/java/com/alibaba/nacos/client/config/filter/impl/ConfigResponse.java
```diff
@@ -22,12 +22,12 @@
 import java.util.HashMap;
 import java.util.Map;
 
-import static com.alibaba.nacos.client.config.common.ConfigConstants.CONFIG_TYPE;
-import static com.alibaba.nacos.client.config.common.ConfigConstants.CONTENT;
-import static com.alibaba.nacos.client.config.common.ConfigConstants.DATA_ID;
-import static com.alibaba.nacos.client.config.common.ConfigConstants.ENCRYPTED_DATA_KEY;
-import static com.alibaba.nacos.client.config.common.ConfigConstants.GROUP;
-import static com.alibaba.nacos.client.config.common.ConfigConstants.TENANT;
+import static com.alibaba.nacos.client.config.filter.impl.ConfigMetadata.CONFIG_TYPE;
```
KomachiSion, line 25: same as above.

KomachiSion (changes requested):

client/src/main/java/com/alibaba/nacos/client/config/filter/impl/ConfigRequest.java
onewe: Because I don't want to  reference var of other module. It doesn't flow the p3c rule.

KomachiSion: The common module define these class and constants for all module used, which to reduce the duplicate codes and constants. 

These  constants both used in server and client. So I think you should use common ConfigConstants.

onewe: Emmm, Yes. I think you are right. I will put all the common constants into the common module, but it's big project for me. So I want to do this one by one. I'm going to optimize all the modules then put extract common constants into the common module. It's my plan.

onewe: Who can lead me that do this? I feel it too hard.😭

PR #7285, opened 2021-11-24
fix(#7284): log print ex details.

### LINKED ISSUE #7284: [1.4.2] Bug: some place log do not print ex details.
**Describe the bug**
some place log do not print ex details. like this:
```java
Loggers.SRV_LOG.error("[IP-DEAD] failed to delete ip automatically, ip: {}, error: {}", instance.toJson(),  throwable);
```

CherishCai:
fix #7284

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/core/ServiceManager.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/core/ServiceManager.java
@@ -298,7 +298,7 @@ public class ServiceManager implements RecordListener<Service> {
                 updatedHealthStatus(namespaceId, serviceName, serverIP);
             } catch (Exception e) {
                 Loggers.SRV_LOG
-                        .warn("[DOMAIN-UPDATER] Exception while update service: {} from {}, error: {}", serviceName,
+                        .warn("[DOMAIN-UPDATER] Exception while update service: {} from {}, error: ", serviceName,
                                 serverIP, e);
             }
         }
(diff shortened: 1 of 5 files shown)


KomachiSion (changes requested):
Almost these log is async task, which will make the error log print many lines. We should discuss it first.

naming/src/main/java/com/alibaba/nacos/naming/core/ServiceManager.java
```diff
@@ -298,7 +298,7 @@ public void run() {
                 updatedHealthStatus(namespaceId, serviceName, serverIP);
             } catch (Exception e) {
                 Loggers.SRV_LOG
-                        .warn("[DOMAIN-UPDATER] Exception while update service: {} from {}, error: {}", serviceName,
```
realJackSun, line 301: Why to remove {}?

realJackSun (changes requested):

naming/src/main/java/com/alibaba/nacos/naming/core/ServiceManager.java
CherishCai: placeholder {} let e.toString(), will lost error stackTrace details.

CherishCai:
> Almost these log is async task, which will make the error log print many lines. We should discuss it first.

Yes, Nacos should use the slf4j facade api for logging. Should not recommend ExceptionUtil.getStackTrace()

CherishCai:
I'm sorry it's late to sign the CLA.

KomachiSion (changes requested):
Whether the original author want to log message only?

CherishCai:
> Whether the original author want to log message only?

If you want to know the details of the error, I think it needs to be handled in this way. Maybe the original author didn't notice it.

KomachiSion (changes requested):
The PushService will print so many error logs, because the push will always retry for failed tasks. should we only print the message not all stack?

CherishCai:
> The PushService will print so many error logs, because the push will always retry for failed tasks. should we only print the message not all stack?

Okay, I'll go back and look at it tonight and change back to the old logic

KomachiSion (approved):

Issue #3046, opened 2020-06-12
Enhancement for constant variables.


jzdayz:
## Issue Description

Type: code review

### Describe what happened (or what feature you want)

I noticed that there are some magic variables in applicationUtils and PropertyUtil, can it be written in one place,it's so messy. i think all config magic variables should be one place 

![image](https://user-images.githubusercontent.com/35563513/84479897-41765e80-acc6-11ea-9342-2e89e53137c2.png)
![image](https://user-images.githubusercontent.com/35563513/84479924-4d622080-acc6-11ea-8294-ebdb979854e9.png)

### Describe what you expected to happen

If it passes, I can do this

jzdayz:
![image](https://user-images.githubusercontent.com/35563513/84480443-29530f00-acc7-11ea-8dfa-cd6149b36ad8.png)
so messy

jzdayz:
- plan1
```java
public interface Constants {
    class Raft{

    }

    class Nacos{
        public static String FUNCTION_MODE_PROPERTY_NAME = "nacos.functionMode";
    }

    class Web{
        public static String ACCEPT_CHARSET = "Accept-Charset";
        public static String SERVER_PORT = "server.port";
    }
}
```
- plan2
```java

public class Constants {
    public static class Raft{

    }

    public static class Nacos{
        public static String FUNCTION_MODE_PROPERTY_NAME = "nacos.functionMode";
    }

    public static class Web{
        public static String ACCEPT_CHARSET = "Accept-Charset";
        public static String SERVER_PORT = "server.port";
    }
}

```

KomachiSion:
Looks great, but I can't get the difference between the two plan, I think plan two is more used.

If you want to do this, please pay attention to the scope of these constants.

jzdayz:
> Looks great, but I can't get the difference between the two plan, I think plan two is more used.
> 
> If you want to do this, please pay attention to the scope of these constants.

ok i will do it

KomachiSion:
@jzdayz Any progress？

jzdayz:
> @jzdayz Any progress？



Sorry, in my idea, I want to put all the constants in common module, but I found that there are constant classes in different modules, so I think that if the code is modified, the range of changes will be larger. I think the code in the future It can be unified, but I dare not move the current code

KomachiSion:
@jzdayz you can split this issue as multiple sub task. each module each task. And refactor for each module.
After all modules has finished, then to think about merge constants and unified them.

[...]

PR #7294, opened 2021-11-25
[ISSUE #3046] Unify the treatment of constant variables and reduce the usage of "magic numbers"

ArthurMograbi:
To resolve issue #3046, unify the treatment of constant variables and reduce the usage of "magic numbers"

## Brief changelog
Replacement of some of the usage of "magic numbers" for a cleaner and more maintainable approach and unify some discordand usages of the same variables

## Verifying this change
eclipse-pmd was used to verify the changes

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/config/impl/ClientWorker.java
+++ b/client/src/main/java/com/alibaba/nacos/client/config/impl/ClientWorker.java
@@ -113,7 +113,8 @@ public class ClientWorker implements Closeable {
     /**
      * groupKey -> cacheData.
      */
-    private final AtomicReference<Map<String, CacheData>> cacheMap = new AtomicReference<>(new HashMap<>());
+    private final AtomicReference<Map<String, CacheData>> cacheMap = new AtomicReference<Map<String, CacheData>>(
+            new HashMap<String, CacheData>());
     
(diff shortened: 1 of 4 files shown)


client/src/main/java/com/alibaba/nacos/client/config/impl/ClientWorker.java
```diff
@@ -113,7 +113,8 @@
     /**
      * groupKey -> cacheData.
      */
-    private final AtomicReference<Map<String, CacheData>> cacheMap = new AtomicReference<>(new HashMap<>());
+    private final AtomicReference<Map<String, CacheData>> cacheMap = new AtomicReference<Map<String, CacheData>>(
```
KomachiSion, line 116: Why add type?

KomachiSion (changes requested):
I think your jdk version is too old. Please upgrade jdk version to at lease 1.8.

ArthurMograbi:
> I think your jdk version is too old. Please upgrade jdk version to at lease 1.8.

Thank you for the heads up! There was something wrong with my workspace that caused java to default to an earlier version

realJackSun:
1 Do not use magic number
2 Remove the types as sion says

KomachiSion:
CLA not sign for a long time.

PR #7404, opened 2021-12-12
Catch exception when push data to client in cancelled listener.

horizonzy:
Catch exception when push data to client in cancelled listener.


core/src/main/java/com/alibaba/nacos/core/remote/grpc/GrpcBiStreamRequestAcceptor.java
```diff
@@ -73,6 +73,12 @@ private void traceDetailIfNecessary(Payload grpcRequest) {
     
     @Override
     public StreamObserver<Payload> requestBiStream(StreamObserver<Payload> responseObserver) {
+        ((ServerCallStreamObserver) responseObserver).setOnCancelHandler(new Runnable() {
+            @Override
+            public void run() {
+                Loggers.REMOTE_DIGEST.warn("Call already cancel");
```
shiyiyue1102, line 79: it should print more detailed log ,such as connection id ,client ip .

KomachiSion:
CLA no signed for a long time.

Issue #4070, opened 2020-10-27
The nacos user is not locked after three consecutive login failures.


stephenxe:
**Is your feature request related to a problem? Please describe.**

Locked upon multiple login failures will enhance system security.

**Describe the solution you'd like**

The system locks an account for a period of time after three consecutive login failures.

**Describe alternatives you've considered**

Do not expose the NACOS management page.

**Additional context**

None.

KomachiSion:
This feature is a strong public security feature. for nacos, it should be a private net app console, so I don't think it's a high priority feature for community.

We will mark this issue to see whether community is interested for it.

KomachiSion:
There is no body want to enhance this issue, and After 2.1 nacos support auth plugin, so I think this issue can be closed.

PR #7439, opened 2021-12-16
(ISSUE #4070) Handle multiple login attempts by same user with new LockedUsers Class

ArthurMograbi:

To prevent possible attacks, both to the host and brute force attacks against users, it is crit8ical to have some sort of safeguard against multiple sequential logins by the same user.

## Brief changelog

Added new LockedUsers class and it's implementation in UserController


KomachiSion:
CLA no signed for a long time.

PR #7513, opened 2021-12-28
The closure of a stream should be in the finally method

hcraderGH:
In order to prevent an exception occurred while closing the stream.

## Brief changelog
Move the closure of Hessian2Output and Hessian2Input steam to the finally method block.


li-xiao-shuang:
> 

sign CLA

hcraderGH:
> ## What is the purpose of the change
> In order to prevent an exception occurred while closing the stream.
> 
> ## Brief changelog
> Move the closure of Hessian2Output and Hessian2Input steam to the finally method block.

PR #7564, opened 2022-01-06
catch exception when register self to cluster.

horizonzy:
catch exception when register self to cluster.

### DIFF
--- a/core/src/main/java/com/alibaba/nacos/core/distributed/raft/JRaftServer.java
+++ b/core/src/main/java/com/alibaba/nacos/core/distributed/raft/JRaftServer.java
@@ -368,7 +368,7 @@ public class JRaftServer {
                 }
                 Loggers.RAFT.warn("Failed to join the cluster, retry...");
             } catch (Exception e) {
-                Loggers.RAFT.warn("Failed to join the cluster, retry...", e);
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/alibaba/nacos/core/distributed/raft/JRaftServer.java
```diff
@@ -357,15 +357,19 @@ public void readFromLeader(final ReadRequest request, final CompletableFuture<Re
...
+                if (peerIds.contains(selfIp)) {
+                    return;
+                }
+                Status status = cliService.addPeer(groupId, conf, selfIp);
+                if (status.isOk()) {
+                    return;
+                }
+                Loggers.RAFT.warn("Failed to join the cluster, retry...");
+            } catch (Exception e) {
+                Loggers.RAFT.warn("Failed to join the cluster, retry...", e);
```
li-xiao-shuang, line 371: catch 到异常 应该打error日志比较好

li-xiao-shuang (dismissed):

core/src/main/java/com/alibaba/nacos/core/distributed/raft/JRaftServer.java
horizonzy: done.

li-xiao-shuang (approved):

PR #7604, opened 2022-01-10
fix #7427. the function of ClusterManagement should be limited to non…

### LINKED ISSUE #7427: Any user can leave a cluster node in the management UI by clicking Leave button
1. Create a user who can access public and any other namespace. 
2. Goto ClusterManagement section, choose a node, click the Leave button. Then the node leave the cluster, and I cannot get it back. 

I think the function of ClusterManagement should be limited to non admin users.

caomingjie-code:
What is the purpose of the change

fix #7427. the function of ClusterManagement should be limited to non admin users.
Brief changelog

the function of ClusterManagement should be limited to non admin users.
Verifying this change

nacos用戶
图片

普通用戶
图片


KomachiSion:
[删除无用的代码 解决7653](https://github.com/alibaba/nacos/pull/7604/commits/1e54ef48c1676332aeb216ba247bfb623e1e56ea)

is not relative with this PR.

PR #7723, opened 2022-01-28
[develop] log what causes I/O reactor status to STOPPED

zrlw:
log what causes I/O reactor status to STOPPED
see #7698


common/src/main/java/com/alibaba/nacos/common/http/client/request/DefaultAsyncHttpClientRequest.java
```diff
@@ -50,30 +60,43 @@ public DefaultAsyncHttpClientRequest(CloseableHttpAsyncClient asyncClient) {
...
+                public void failed(Exception ex) {
+                    callback.onError(ex);
+                }
+                
+                @Override
+                public void cancelled() {
+                    callback.onCancel();
+                }
+            });
+        } catch (IllegalStateException e) {
```
KomachiSion, line 88: Only catch IllegalStateException?

zrlw: CloseableHttpAsyncClientBase#ensureRunning

KomachiSion (approved):

PR #7805, opened 2022-02-21
[1.x] Patch v1.x DistroConsistencyServiceImpl

zrlw:
modify v1.x DistroConsistencyServiceImpl to eliminate redundant tasks as many as possible.

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/consistency/ephemeral/distro/DistroConsistencyServiceImpl.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/consistency/ephemeral/distro/DistroConsistencyServiceImpl.java
@@ -39,8 +39,6 @@ import com.alibaba.nacos.naming.misc.GlobalExecutor;
 import com.alibaba.nacos.naming.misc.Loggers;
 import com.alibaba.nacos.naming.misc.SwitchDomain;
 import com.alibaba.nacos.naming.pojo.Record;
-import org.apache.commons.lang3.StringUtils;
-import org.javatuples.Pair;
 import org.springframework.context.annotation.DependsOn;
 
 import javax.annotation.PostConstruct;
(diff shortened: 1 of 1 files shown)


naming/src/main/java/com/alibaba/nacos/naming/consistency/ephemeral/distro/DistroConsistencyServiceImpl.java
```diff
@@ -380,11 +380,14 @@ public boolean isInitialized() {
...
-            if (services.containsKey(datumKey) && action == DataOperation.CHANGE) {
-                return;
-            }
             if (action == DataOperation.CHANGE) {
-                services.put(datumKey, StringUtils.EMPTY);
+                if (services.putIfAbsent(datumKey, StringUtils.EMPTY) != null) {
+                    // if previous change task is waiting for handling.
+                    return;
+                }
+            } else if (action == DataOperation.DELETE) {
```
KomachiSion, line 388: I think the change may cause lose DELETE action.

```
                } else if (action == DataOperation.DELETE) {
                    if (services.contains(datumKey)) {
                        // if new change task is waiting for handling.
                        return;
                    }
                }
```

KomachiSion (changes requested):

naming/src/main/java/com/alibaba/nacos/naming/consistency/ephemeral/distro/DistroConsistencyServiceImpl.java
zrlw: DELETE action will be ignored only when it's followed by a CHANGE action.

zrlw: ```services``` should be separated into two groups that one group only has DELETE actions and the other only has CHANGE actions if bypass duplicated DELETE actions.

PR #7806, opened 2022-02-21
[1.x] Patch v1.x member manager

zrlw:
1. report metadata to other nodes concurrently to eliminate the side effect of PR #4948 - which made the initial status of all other nodes to DOWN, changing their status all back to UP took much more time if report metadata only to one of them at each report task.  And even worse than that if the number of other nodes is more than 7, some beat requests will not be transfer to the correct node till timeout.
2. eliminate concurrent issue between MemberUtil#onFail and DistroMapper#onEvent when two or more nodes is DOWN and  reporting metadata to other nodes concurrently.
3. apply PR #7236 from develop branch to ignore the WebServerInitializedEvent of management.

### DIFF
--- a/core/src/main/java/com/alibaba/nacos/core/cluster/MemberUtil.java
+++ b/core/src/main/java/com/alibaba/nacos/core/cluster/MemberUtil.java
@@ -135,7 +135,6 @@ public class MemberUtil {
     public static void onFail(final ServerMemberManager manager, final Member member, Throwable ex) {
         manager.getMemberAddressInfos().remove(member.getAddress());
         final NodeState old = member.getState();
-        member.setState(NodeState.SUSPICIOUS);
         member.setFailAccessCnt(member.getFailAccessCnt() + 1);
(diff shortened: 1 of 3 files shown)


KomachiSion (changes requested):
If execute report to all nodes each time, whether will cause high pressure? The report only interval 2 seconds.

zrlw:
> If execute report to all nodes each time, whether will cause high pressure? The report only interval 2 seconds.

it's a trade-off solution. 
let's say there are 10 nodes in nacos registry cluster. 2 seconds * 10 nodes = 20 seconds, 
it means one restart nacos node must wait 20 seconds to accomplish setting all other nodes status from DOWN  to UP, 
1. beat forwarding between nodes returns to normal after 20 seconds.
2. default value of beat timeout is 15 seconds
so some instance beat request might not be sent to the responsible node and the status of instance will be set unhealthy.
it will cause no provider issue if using tag routing and the instances of tag XXX all became unhealthy.

PR #7811, opened 2022-02-21
[1.x] Patch v1.x apache sync http client factory

zrlw:
modify connection pool parameters of ApacheSyncHttpClientFactory.
1. setConnectionTimeToLive to -1: keep connection alive as long as possible to support http1.1 keep alive mode.
2. setMaxConnPerRoute to 200: equal the default threads number of Tomcat, it might be set larger if having ample CPUs.
2. setMaxConnTotal to 2000:  it might be set larger if having more than 10 nacos nodes.
this PR is not ideal because MaxConnPerRoute and MaxConnTotal should match with running enviroment.

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/misc/HttpClientManager.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/misc/HttpClientManager.java
@@ -148,9 +148,9 @@ public class HttpClientManager {
         
         @Override
         protected HttpClientConfig buildHttpClientConfig() {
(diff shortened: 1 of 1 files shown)


KomachiSion:
Will too many threads be created?

zrlw:
> Will too many threads be created?

The original value of these parameters are too small to support large scale beat forwarding between nacos nodes for Distro protocol. These values are not general solution, they were obtained under a special stress test scenario:
Nacos cluster:  5 nodes * 8c16g (each node)
Testing target: 
1. support beat requests as many as possible (mock beat request by Jmeter java request class).
2. each naming-event.log remain stable.
3. each nacos node cpu < 80%, memory < 80%.

PR #7863, opened 2022-02-28
[1.x] add request type field at the head of task data, exchange parse order between Log and GetRequest 

zrlw:

see #7757


zrlw:
> Can you provide some unit tests to verify that this implementation is guaranteed to be compatible with old and new?

done

KomachiSion (approved):

Issue #7925, opened 2022-03-10
1.x client still requests to the server after calling unsubscribe()


gaofeihang:
**Describe the bug**
1.x client still requests to the server after calling unsubscribe(). The UpdateTask in HostReactor has no way to be stopped.

**Expected behavior**
After calling unsubscribe() and removing all listeners, the client stops to request to the server and stops to poll `/nacos/v1/ns/instance/list`. This expected behavior can reduce the QPS and the pressure of the Nacos server in some cases.

**Acutally behavior**
After calling unsubscribe(), the client is still polling the server. UpdateTask is scheduled in a finally block in the run() method. It can never be stopped.

**How to Reproduce**
1. Start a 1.4.3 Nacos server
2. Register an instance as tutorial
3. Discover the sevice, then sleep for a while（10s）
4. Call unsubscribe() using the same listener as subscribe() and therefore remove all listeners.
5. Debug UpdateTask or observe the access log of the sever. The reqeusts to `/nacos/v1/ns/instance/list` still exist.

**Desktop (please complete the following information):**
 - OS: macOS Big Sur
 - Version [nacos-server 1.4.3, nacos-client 1.4.3]
 - Module [naming / nacos-client]
 - SDK [original]

**Additional context**
Codes for reproducing the problem.
```
NamingService naming = NamingFactory.createNamingService("127.0.0.1:8848");
EventListener listener = event -> {
    if (event instanceof NamingEvent) {
        System.out.println(((NamingEvent) event).getServiceName());
        System.out.println(((NamingEvent) event).getInstances());
    }
};

naming.subscribe("nacos.test.3", listener);

Thread.sleep(1000 * 10);

naming.unsubscribe("nacos.test.3", listener);

Thread.sleep(1000 * 600);
```

PR #7927, opened 2022-03-10
[ISSUE #7925] Client stops the UpdateTask after a service is unsubscr…

gaofeihang:

FIX https://github.com/alibaba/nacos/issues/7925

## Brief changelog

[ISSUE #7925] Client stops the UpdateTask after a service is unsubscribed.

## Verifying this change

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/naming/core/HostReactor.java
+++ b/client/src/main/java/com/alibaba/nacos/client/naming/core/HostReactor.java
@@ -62,7 +62,7 @@ public class HostReactor implements Closeable {
     
     private static final long UPDATE_HOLD_INTERVAL = 5000L;
     
-    private final Map<String, ScheduledFuture<?>> futureMap = new HashMap<String, ScheduledFuture<?>>();
+    private final Map<String, ScheduledFuture<?>> futureMap = new ConcurrentHashMap<String, ScheduledFuture<?>>();
     
     private final Map<String, ServiceInfo> serviceInfoMap;
     
@@ -148,6 +148,12 @@ public class HostReactor implements Closeable {
      */
     public void unSubscribe(String serviceName, String clusters, EventListener eventListener) {
         notifier.deregisterListener(serviceName, clusters, eventListener);
+
+        // when all listeners are removed, stop the update task
+        if (!notifier.isSubscribed(serviceName, clusters)) {
+            // remove the flag for judging whether to stop the update task
+            futureMap.remove(ServiceInfo.getKey(serviceName, clusters));
+        }
     }
     
     public List<ServiceInfo> getSubscribeServices() {
@@ -450,6 +456,7 @@ public class HostReactor implements Closeable {
         
         @Override
         public void run() {
+            boolean stop = false;
             long delayTime = DEFAULT_DELAY;
             
             try {
@@ -474,6 +481,9 @@ public class HostReactor implements Closeable {
                 if (!notifier.isSubscribed(serviceName, clusters) && !futureMap
                         .containsKey(ServiceInfo.getKey(serviceName, clusters))) {
                     // abort the update task
+                    stop = true;
+                    // clear the service info cache
+                    serviceInfoMap.remove(ServiceInfo.getKey(serviceName, clusters));
                     NAMING_LOGGER.info("update task is stopped, service:" + serviceName + ", clusters:" + clusters);
                     return;
                 }
@@ -487,7 +497,9 @@ public class HostReactor implements Closeable {
                 incFailCount();
                 NAMING_LOGGER.warn("[NA] failed to update serviceName: " + serviceName, e);
             } finally {
-                executor.schedule(this, Math.min(delayTime << failCount, DEFAULT_DELAY * 60), TimeUnit.MILLISECONDS);
+                if (!stop) {
+                    executor.schedule(this, Math.min(delayTime << failCount, DEFAULT_DELAY * 60), TimeUnit.MILLISECONDS);
+                }
             }
         }
     }
--- a/client/src/test/java/com/alibaba/nacos/client/naming/core/HostReactorTest.java
+++ b/client/src/test/java/com/alibaba/nacos/client/naming/core/HostReactorTest.java
@@ -174,6 +174,11 @@ public class HostReactorTest {
         hostReactor.processServiceJson(CHANGE_DATA_EXAMPLE);
         
         Assert.assertEquals(0, count.intValue());
+
+        Thread.sleep(1000);
+
+        Assert.assertFalse(hostReactor.getServiceInfoMap().containsKey(
+                ServiceInfo.getKey("testName", "testClusters")));
     }
     
     @Test


client/src/main/java/com/alibaba/nacos/client/naming/core/HostReactor.java
```diff
@@ -471,9 +472,12 @@ public void run() {
                 
                 lastRefTime = serviceObj.getLastRefTime();
                 
-                if (!notifier.isSubscribed(serviceName, clusters) && !futureMap
-                        .containsKey(ServiceInfo.getKey(serviceName, clusters))) {
+                if (!notifier.isSubscribed(serviceName, clusters)) {
```
KomachiSion, line 475: Why remove futureMap.containsKey?

KomachiSion (changes requested):

client/src/main/java/com/alibaba/nacos/client/naming/core/HostReactor.java
gaofeihang: I searched for the usage of futureMap. There are only get and put operations. Once a service is put into futureMap, it can never be removed. This will make futureMap.containsKey() always return true and make the if block can not be entered.

KomachiSion: If judgement like this, It will cause problem when I call NamingService.getAllIsntance(subscribe = true).

The cache in HostReactor is always old one, will not be refresh.

KomachiSion (changes requested):

client/src/main/java/com/alibaba/nacos/client/naming/core/HostReactor.java
gaofeihang: If the cache refers to serviceInfoMap，is it OK to clear the cache after the judgement? Like HostReactor.java Line 479 in the PR.

KomachiSion: If you real want to remove cache. Maybe you clear it when you call unsubscribe not in HostReactor.

KomachiSion (changes requested):

client/src/main/java/com/alibaba/nacos/client/naming/core/HostReactor.java
gaofeihang: When unsubscribe is called, the next round UpdateTask may be already scheduled. If the cache is cleared at this moment, the next round UpdateTask will fill the cache again.
I think it's better to clear the cache exactly when the UpdateTask can be terminated, to make sure the UpdateTask will not run afterwards.

KomachiSion: > If judgement like this, It will cause problem when I call NamingService.getAllIsntance(subscribe = true).

> The cache in HostReactor is always old one, will not be refresh.

can't solve this problem. It will cause data no right. So I think only do once more query is no much problem.

gaofeihang: I try to summarize the related logic as below.

1. NamingService.getAllIsntance(subscribe = true) reads cache data from HostReactor.serviceInfoMap. If serviceInfoMap has no data that belongs to a service, there is no old data problem.

2. When unsubscribe clears all listeners in HostReactor.InstancesChangeNotifier, it indicates that the UpdateTask can be stoped. The original judgement `futureMap.containsKey()` has no effect, it always returns true and should be removed.

3. When UpdateTask can be judged to be stopped, change the flag stop = true so that no more UpdateTask will run and no more cache filling operation can be executed. Then clear the cache in serviceInfoMap and make no old data problem caused. The key codes are as follows
```
if (!notifier.isSubscribed(serviceName, clusters)) {
    // abort the update task
    stop = true;
    String serviceKey = ServiceInfo.getKey(serviceName, clusters);
    serviceInfoMap.remove(serviceKey);
    futureMap.remove(serviceKey);
    return;
}
``` 

4. If cache is cleared after unsubscribe will case problem. Suppose that 
10:00:01 unsubscribe is called, serviceInfoMap is cleared.
10:00:02 UpdateTask runs, fills the cache before the judgement by 
```
ServiceInfo serviceObj = serviceInfoMap.get(ServiceInfo.getKey(serviceName, clusters));
                
if (serviceObj == null) {
    updateService(serviceName, clusters);
    return;
}
```
Afterwards the task is stopped after judgement without clearing the cache, the old data will remain in serviceInfoMap.

It would be appreciated if you tell me which of the 4 points is wrong. Then I can better understand where the problem is. Thanks.

KomachiSion: > I try to summarize the related logic as below.
> 
> 1. NamingService.getAllIsntance(subscribe = true) reads cache data from HostReactor.serviceInfoMap. If serviceInfoMap has no data that belongs to a service, there is no old data problem.
> 2. When unsubscribe clears all listeners in HostReactor.InstancesChangeNotifier, it indicates that the UpdateTask can be stoped. The original judgement `futureMap.containsKey()` has no effect, it always returns true and should be removed.
> 3. When UpdateTask can be judged to be stopped, change the flag stop = true so that no more UpdateTask will run and no more cache filling operation can be executed. Then clear the cache in serviceInfoMap to avoid old data problem The key codes are as follows
> 
> ```
> if (!notifier.isSubscribed(serviceName, clusters)) {
>     // abort the update task
>     stop = true;
>     String serviceKey = ServiceInfo.getKey(serviceName, clusters);
>     serviceInfoMap.remove(serviceKey);
>     futureMap.remove(serviceKey);
>     return;
> }
> ```
> 
> 4. If cache is cleared after unsubscribe will case problem. Suppose that
>    10:00:01 unsubscribe is called, serviceInfoMap is cleared.
>    10:00:02 UpdateTask runs, fills the cache before the judgement by
> 
> ```
> ServiceInfo serviceObj = serviceInfoMap.get(ServiceInfo.getKey(serviceName, clusters));
>                 
> if (serviceObj == null) {
>     updateService(serviceName, clusters);
>     return;
> }
> ```
> 
> Afterwards the task is stopped after judgement without clearing the cache, the old data will remain in serviceInfoMap.
> 
> It would be appreciated if you tell me which of the 4 points is wrong. Then I can better understand where the problem is. Thanks.

If like this. Why nacos add `NamingService.getAllIsntance(subscribe = true)`? After your changes, `NamingService.getAllIsntance(subscribe = true)` is same as `NamingService.getAllIsntance(subscribe = false)`. And use one more thread and task to do. So I think we can't change like this.

gaofeihang: After these changes, NamingService.getAllIsntance(subscribe = true) can work as before. It still return instances and start to subscribe. There is no difference before unsubscribe() is called. NamingService.getAllIsntance(subscribe = false) does not start subscribing, these two invocations do not have the same behavior.

But when unsubscribe operation is needed, these changes can make sense. Without these changes, even if a service is unsubscribed and all related listeners are deregistered, the UpdateTask is still running. This will case some (maybe a lot of) useless UpdateTasks run continuously. Both client side and server side have to pay the cost of useless polling. This also breaks the definition of unsubscribe: poling is expected to be stopped after unsubscribe in common sense.

I think it is a problem worth to be fixed. If these changes are not good enough, other fix plans are still recommended. If this problem can be ignored or it is intentionally designed like that, it's also OK to me.

KomachiSion: `NamingService.getAllIsntance(subscribe = true)` no add subscriber to notify. So the first time updateTask run will remove cache and task after your change.

gaofeihang: Yes, you are right. Now I know where the problem is. Thanks a lot for explaining that to me. Meanwhile I have changed the pr as follows.

1. Keep the original judgement, use futureMap as a flag to judge whether to stop the update task. Therefore `NamingService.getAllIsntance(subscribe = true)` can work as before. Update task will not stop when there is no listener.
2. Remove the futureMap flag when unsubscribe is called and all listeners are removed in order to fix the issue.

Hope it's a better solution now.

KomachiSion (approved):

PR #8102, opened 2022-04-08
Fix Flaky test

plzdoo:
The method `testToJson1` is flaky due to it is depend on method `toJson` in class `JacksonUtils`. This test use method `writeValueAsString` which may result indeterministic order when parsing object to json string. 

I added a configuration for ObjectMapper which sort the object alphabetically and change the expected value of json string which fix the flakiness of the test. 

I use `mvn -pl common edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=JacksonUtilsTest#testToJson1 --DnondexRuns=10 ` to detect the test is flakiness.


KomachiSion (changes requested):
CLA and PR is unreasonable

PR #8116, opened 2022-04-08
[Optimize] Set interrupted flag to caller

Oliverwqcwrw:

Set interrupted flag to caller

## Brief changelog

XX

## Verifying this change

XXXX

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/config/utils/ConcurrentDiskUtil.java
+++ b/client/src/main/java/com/alibaba/nacos/client/config/utils/ConcurrentDiskUtil.java
@@ -220,6 +220,8 @@ public class ConcurrentDiskUtil {
             Thread.sleep(time);
         } catch (InterruptedException e) {
             LOGGER.warn("sleep wrong", e);
+            // set the interrupted flag
+            Thread.currentThread().interrupt();
         }
     }
 }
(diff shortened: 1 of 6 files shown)


KomachiSion (commented):
I can't get why change in this PR.

Oliverwqcwrw:
If we catch InterruptedException but do nothing,This can cause the thread closing delay and lost the interrupted thread info,So we should return interrupted flag to caller and deal the interrupted thread

KomachiSion:
I don't think so, InterruptedException only stop the sleep and wait lock operation, the thread need to do other logic. Not means the thread need be interrupted

KomachiSion:
> InterruptedException not only stop the sleep and wait lock operation,It can be interrupted by any thread if need,It should be return the interrupted flag to the caller,Otherwise,The thread can not know what do next

Whether cause the following logic will not execute or cause other problem? I think the flag be set will cause the thread has some different habavior.

Oliverwqcwrw:
If the interrupt status is not set, the thread will be delayed close and may occupy thread resources

KomachiSion:
delayed close can be accepted, but if set flag cause thread early recycling can't be accepted.

Oliverwqcwrw:
Delay close is also an optimistic idea and may not be closed and occupy thread resources.

So I think return the interrupted flag to caller better

KomachiSion (approved):
ok

PR #8148, opened 2022-04-13
[1.x] accept node return beat response and forward beat request

zrlw:
To reduce beat retransmission times and bandwidth consumption between nodes,  and improve horizontal expansion capability. 
see #8147
1. the node which received client beat answers directly instead of responsible node.
2. each node merge, compress and retransmit beat requests that they answered at step 1 to the responsible nodes.


KomachiSion:
Close reason in issue。

PR #8171, opened 2022-04-15
[1.x] optimize client beat processor to avoid FullGC

zrlw:
see #8139

## Brief changelog
1. doing client beat processing at the thread which run InstanceController to avoid FullGC which will be trigged by huge client beat requestion.
2. get client beat instances identified by key instead of full searching.
3. merge same service change events by queuing.


KomachiSion:
We need discuss first in issue.

PR #8202, opened 2022-04-19
issue#8163 fix multi instance share the same local snapshot

shiyiyue1102:
fix bug isuue 8163

add nacos.cache.data.init.snapshot properties ,to determine if load snapshot content on  startup  which will result in a default notification on initialing  ,default is true

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/config/impl/ServerListManager.java
+++ b/client/src/main/java/com/alibaba/nacos/client/config/impl/ServerListManager.java
@@ -132,28 +132,28 @@ public class ServerListManager implements Closeable {
         for (String serverAddr : fixed) {
             String[] serverAddrArr = InternetAddressUtil.splitIPPortStr(serverAddr);
             if (serverAddrArr.length == 1) {
-                serverAddrs.add(serverAddrArr[0] + InternetAddressUtil.IP_PORT_SPLITER + ParamUtil.getDefaultServerPort());
+                serverAddrs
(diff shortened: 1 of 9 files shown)


client/src/main/java/com/alibaba/nacos/client/config/impl/ServerListManager.java
```diff
@@ -132,28 +132,28 @@ public ServerListManager(List<String> fixed, String namespace) {
         for (String serverAddr : fixed) {
             String[] serverAddrArr = InternetAddressUtil.splitIPPortStr(serverAddr);
             if (serverAddrArr.length == 1) {
-                serverAddrs.add(serverAddrArr[0] + InternetAddressUtil.IP_PORT_SPLITER + ParamUtil.getDefaultServerPort());
+                serverAddrs
```
li-xiao-shuang, line 135: I don't think this change is necessary

shiyiyue1102: the modification on ServerListManager  has several purposes:
1.optimize the naming rule of server name  when  the serverlistmanager is using endpoint , port,contentpath ,cluster name which construct the whole address server. 
2.support endpoint query string parameter after the address server url to satisfy more  usage scenarios , for example, this query   string is essential in alibaba internal scene when using address server.
3.support user defined server name to serverlistmanager and client worker , to make our log print more graceful in multi instances  scenarios.

li-xiao-shuang (approved):

PR #8322, opened 2022-05-09
Support for ibm J9

lyh200:
Now system metrics are obtained from com.sun.management.OperatingSystemMXBean for HotSpot JVM, but these are in com.ibm.lang.management.OperatingSystemMXBean for Ibm J9, so need to deduce the target source.


KomachiSion:
Please only contain your own commit in your PR.

Issue #8424, opened 2022-05-20
2.x的auth插件模块的Spring Bean存在循环依赖


liqipeng:
The dependencies of some of the beans in the application context form a cycle:

   nacosAuthManager (field private com.alibaba.nacos.plugin.auth.impl.JwtTokenManager com.alibaba.nacos.plugin.auth.impl.NacosAuthManager.tokenManager)
┌─────┐
|  jwtTokenManager (field private com.alibaba.nacos.plugin.auth.impl.NacosAuthConfig com.alibaba.nacos.plugin.auth.impl.JwtTokenManager.nacosAuthConfig)
↑     ↓
|  nacosAuthConfig (field private com.alibaba.nacos.plugin.auth.impl.JwtTokenManager com.alibaba.nacos.plugin.auth.impl.NacosAuthConfig.tokenProvider)
└─────┘

liqipeng:
#8184

KomachiSion:
Fixed by https://github.com/alibaba/nacos/pull/8418.

PR #8425, opened 2022-05-20
[ISSUE #8424] solve circular reference between `JwtTokenManager` and `NacosAuthConfig`(2.x)

liqipeng:
#8424

#8184


KomachiSion:
There are some one fixed the problem. #8418

PR #8437, opened 2022-05-22
[ISSUE #8436] start distro verify task after load data

liqipeng:
#8436 数据丢失场景1


KomachiSion:
refuse reason in issue

PR #8438, opened 2022-05-22
[ISSUE #8436] Not allow /v1/ns/distro/datums be visited util server is UP

liqipeng:
#8436 数据丢失场景2


KomachiSion:
Refuse reason in issue。

Issue #8471, opened 2022-05-28
[clean code]: has no bolt, jraft-core exclude bolt and use rpc-grpc-impl.


CherishCai:
Has no`bolt`, `jraft-core` exclude`bolt` and use `rpc-grpc-impl`. May be clean under code.

https://github.com/alibaba/nacos/blob/0eb530ef99b994be1ef1d4e620f69747df62cf19/core/src/main/java/com/alibaba/nacos/core/distributed/raft/JRaftServer.java#L109-L116

KomachiSion:
What's the difference？

CherishCai:
> What's the difference？

sofa-jraft include grpc , excludes bolt, So clean up the above code.

```xml
           <!-- JRaft -->
            <dependency>
                <groupId>com.alipay.sofa</groupId>
                <artifactId>jraft-core</artifactId>
                <version>${jraft-core.version}</version>
                <exclusions>
                    <exclusion>
                        <groupId>com.alipay.sofa</groupId>
                        <artifactId>bolt</artifactId>
                    </exclusion>
                    <exclusion>
                        <groupId>org.apache.logging.log4j</groupId>
                        <artifactId>log4j-api</artifactId>
                    </exclusion>
                    <exclusion>
                        <groupId>org.apache.logging.log4j</groupId>
                        <artifactId>log4j-core</artifactId>
                    </exclusion>
                    <exclusion>
                        <groupId>org.apache.logging.log4j</groupId>
                        <artifactId>log4j-slf4j-impl</artifactId>
                    </exclusion>
                    <exclusion>
                        <groupId>org.apache.logging.log4j</groupId>
                        <artifactId>log4j-jcl</artifactId>
                    </exclusion>
                </exclusions>
            </dependency>
            
            <dependency>
                <groupId>com.alipay.sofa</groupId>
                <artifactId>rpc-grpc-impl</artifactId>
                <version>${rpc-grpc-impl.version}</version>
            </dependency>

```

PR #8475, opened 2022-05-28
[ISSUE #8471] has no bolt, remove bolt properties

CherishCai:
[clean code]: has no bolt, jraft-core exclude bolt and use rpc-grpc-impl. (#8471)

So remove bolt properties.


horizonzy (approved):
LGTM.

PR #8477, opened 2022-05-28
[ISSUE #8472]: Dependency Governance, delete no use plugin versions.

CherishCai:
feat(#8472): [Dependency Governance] revision with flatten, delete no use plugin versions. (#8472)


KomachiSion (changes requested):
This plugin has useful when we need upgrade the versions

CherishCai:
> This plugin has useful when we need upgrade the versions

Sorry, 我以为使用 revision with flatten , versions 是多余的

PR #8573, opened 2022-06-14
Make InstanceChangeNotifier is not notified by different namespace.

horizonzy:

Fixes-#6864


liqipeng:
Maybe I had fixed a same issue?  #8433

PR #8596, opened 2022-06-19
Explicitly specify spring-boot-maven-plugin version same as spring-boot

liqipeng:
Explicitly specify ```spring-boot-maven-plugin``` version same as spring-boot by pluginManagement.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,6 +158,15 @@
     <!-- =========================================================Build plugins================================================ -->
     <!-- == -->
     <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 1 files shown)


KomachiSion:
What‘s the reason add `spring-boot-maven-plugin`?

liqipeng:
> What‘s the reason add `spring-boot-maven-plugin`?

@KomachiSion 
```spring-boot-maven-plugin``` is used in console, address, config module, and no version specified.
This PR is not to add ```spring-boot-maven-plugin``` plugin to parent pom, but add  spring-boot-maven-plugin to pluginManagement to make the version specified.

I meet a similar error describe in https://www.jianshu.com/p/878a8b62e70f when I build nacos in a differenct PC.
So I think it is better to specify a version to keep version fixed with spring-boot version.

KomachiSion (approved):

PR #8805, opened 2022-07-26
[ISSUE#8481]Unified TRACE capacity building 

scienceZ:
[[ISSUE#8481]Unified TRACE capacity building]
1. Event type add to traceEvent.
2. The event about instance add instancePort.
3. Add health check type to healthChangeEvent.
4. Other fixup.


common/src/main/java/com/alibaba/nacos/common/trace/event/TraceEvent.java
```diff
@@ -17,13 +17,16 @@
...
 import com.alibaba.nacos.common.notify.Event;
+import com.alibaba.nacos.common.trace.TraceEventType;
 
 /**
  * Trace event.
  *
  * @author yanda
  */
 public class TraceEvent extends Event {
+    private TraceEventType type;
```
KomachiSion, line 28: I think the type is `String` will be better.

common/src/main/java/com/alibaba/nacos/common/trace/publisher/TraceEventPublisher.java
```diff
@@ -103,7 +103,6 @@ public boolean publish(Event event) {
         boolean success = this.queue.offer(event);
         if (!success) {
             LOGGER.warn("Unable to plug in due to interruption, synchronize sending time, event : {}", event);
-            handleEvent(event);
```
KomachiSion, line 106: remove call handleEvent logic, the warning log should be change also.

KomachiSion (changes requested):

KomachiSion (approved):

PR #8864, opened 2022-08-03
fix group with point . bug issues#8860

moonciki:
fix bug when group contains point mark "."
see https://github.com/alibaba/nacos/issues/8860


KomachiSion:
I think we should more discuss in issue.

PR #8876, opened 2022-08-04
fix group contains point mark bug issues#8860

moonciki:
fix bug when group contains point mark "."
see https://github.com/alibaba/nacos/issues/8860


Robyn37 (approved):

KomachiSion:
现在issue里细化一下吧。

PR #8932, opened 2022-08-10
[ISSUE #8924]optimize subscriber get generic event type

### LINKED ISSUE #8924: Optimizing abstract class Subscriber suggestion
I found that many Subscriber implementation classes obtain subscribeType as generic classes during reading the source code. The abstract class Subscriber can optimize the subscribeType() method to obtain the generic class of subclasses by default, reducing unnecessary code in the project.
example:
image
image

karsonto:
Close #8924

### DIFF
--- a/auth/src/main/java/com/alibaba/nacos/auth/config/AuthConfigs.java
+++ b/auth/src/main/java/com/alibaba/nacos/auth/config/AuthConfigs.java
@@ -18,7 +18,6 @@ package com.alibaba.nacos.auth.config;
 
 import com.alibaba.nacos.common.JustForTest;
 import com.alibaba.nacos.common.event.ServerConfigChangeEvent;
-import com.alibaba.nacos.common.notify.Event;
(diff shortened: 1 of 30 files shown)


KomachiSion (changes requested):
I don't think it's a good way. I will contain many hide informations. I think the old way can make readers fast understand the subscriber sub which events.

KomachiSion:
Please has some discuss in issue first before you submit PR.

vergilyn:
It is not recommended to obtain the `subscribeType` by reflection **every time**.

can reference：<https://github.com/baomidou/mybatis-plus/blob/master/mybatis-plus-extension/src/main/java/com/baomidou/mybatisplus/extension/service/impl/ServiceImpl.java>
image

karsonto:
@vergilyn I know, but the Subscriber will register to publisher once when initialization in constructor,so it is unnecessary to waste memory  to save event class.
image
image

Issue #8993, opened 2022-08-18
Close input stream and output stream by try with resource[nacos-common]


openonx:
Close input stream and output stream by try with resource[nacos-common]

PR #8995, opened 2022-08-18
[ISSUE #8993]Close input stream and output stream by try with resourc…

openonx:
[ISSUE #8993]Close input stream and output stream by try with resource[nacos-common]


openonx:
code style error，I will recommit

Issue #8460, opened 2022-05-25
[ASoC 2022] Nacos configuration modify hook plugin (Nacos 配置变更Hook插件化)


KomachiSion:
# Background
This is a Basic subject of [ASoC 2022](https://opensource.alibaba.com/asoc2022) and #8456 .

As a configuration center, Nacos can publish, modify, and delete application configurations. Many users in the community want to add some special processes to the configuration change process of Nacos, such as: [webhook](https://github.com/alibaba/nacos/issues/8338), [format verification](https://github.com/alibaba/nacos/issues/8044), [whitelist](https://github.com/alibaba/nacos/issues/4912) and other functions.

This type of function has the particularity of each user, so it is not suitable to add all of them to the configuration change process. Therefore, we want to add a plugin to meet the different needs of different users before and after configuration changes.

# Target
Design and implement a Hook plugin, use the Hook plugin in the main process of configuration changes (publish, modify, delete), so that users can implement plugin capabilities according to their own needs.

# Difficulty

*Basic*

# Mentor

- @[yanlinly](https://github.com/yanlinly) water.lyl@alibaba-inc.com

# Output Requirements
- Design Hook plugin interface for configuration changes;
- Select the appropriate design pattern to implement the plugin;
- Provide complete plugin development documentation and add unit tests;

# Technical Requirements
- Familiar with Java;
- Familiar with design patterns;
- Get to know Nacos;
- Understand the SPI;

----------------

# 背景
这是一个[阿里巴巴编程之夏 2022](https://opensource.alibaba.com/asoc2022) 的基础课题 #8456 。

Nacos作为配置中心，能够发布，修改，删除应用配置，社区中许多用户希望在Nacos的配置变更流程中增加一些特殊流程，如：[webhook](https://github.com/alibaba/nacos/issues/8338)，[格式校验](https://github.com/alibaba/nacos/issues/8044)，[白名单](https://github.com/alibaba/nacos/issues/4912)等功能。

这类功能具有各个用户的业务特性，因此不合适全部加入到配置变更的流程当中，因此我们想加入一个插件机制，来满足不同用户对于配置变更前后的不同需求。

# 目标
设计并实现一个Hook插件，在配置变更（发布，修改，删除）的主流程中使用该Hook插件，让用户可以根据自己的需求实现插件能力。

# 难度

*基础*

# 导师

- @[yanlinly](https://github.com/yanlinly) water.lyl@alibaba-inc.com

# 产出要求
[...]

PR #9029, opened 2022-08-24
[ISSUE #8460]add config change plugin

lyf712:

#8460

In order to meet the additional needs of users for configuration change, SPI mechanism is used to implement the configuration change plugin, which can flexibly expand the plugin service of configuration change

## Brief changelog

1. add nacos-config-plugin module,Introducing relevant dependencies （aop,cloudevent,fastjson）
2. add aspect to pointcut the interfaces of configuration change in nacos-config-plugin module
3. define config change serivce inteface,and abstract major config change plugin service in nacos-config-plugin module
4. implements the default major configuration change plugin service in nacos-plugin-default-impl module
5. add some unit tests to validate the service

### DIFF
--- a/console/src/main/resources/application.properties
+++ b/console/src/main/resources/application.properties
@@ -13,77 +13,55 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
 #*************** Spring Boot Related Configurations ***************#
 ### Default web context path:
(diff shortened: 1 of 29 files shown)


KomachiSion:
Should we add this PR into a single branch like Asoc#8460 first? @yanlinly

PR #9080, opened 2022-09-02
Fix the actual number of health check retries less than the set one.

### LINKED ISSUE #9079: Bug about health check retry times.
https://github.com/alibaba/nacos/blob/0aa695e0d3ae5209f015bf9fd23ea7f06632c002/common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java#L466-L483

I think there's a problem here, if healthCheckRetryTimes is equal to 1, it doesn't actually retry.
The judgment condition should be：reTryTimes >= 0

stone-98:

Fix #9079.

## Brief changelog

XX

## Verifying this change

XXXX

### DIFF
--- a/common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java
+++ b/common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java
@@ -469,7 +469,7 @@ public abstract class RpcClient implements Closeable {
             return false;
         }
         int reTryTimes = healthCheckRetryTimes;
-        while (reTryTimes > 0) {
+        while (reTryTimes >= 0) {
             reTryTimes--;
             try {
                 Response response = this.currentConnection.request(healthCheckRequest, healthCheckTimeOut);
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java
```diff
@@ -469,7 +469,7 @@ private boolean healthCheck() {
             return false;
         }
         int reTryTimes = healthCheckRetryTimes;
-        while (reTryTimes > 0) {
+        while (reTryTimes >= 0) {
```
Oliverwqcwrw, line 472: I think if it is equal to 0 it will cause one more retry

stone-98: I don't understand what you mean. When `healthCheckRetryTimes` is equal to 0, it means that there will be no retry, only a normal request will be initiated, and no retry will be performed.

KomachiSion: The old code whether no do request when `healthCheckRetryTimes` is 0?

stone-98: > The old code whether no do request when `healthCheckRetryTimes` is 0?

Yes, in the old code if healthCheckRetryTimes equals 0, no request will be made. The healthCheck method returns false directly.

KomachiSion (approved):

PR #9100, opened 2022-09-06
Fix flaky test for JacksonUtilsTest.toJsonBytes and JacksonUtilsTest.toJson1

huan1372:
**There are two flaky tests found in `JacksonUtilsTest` :**
- `JacksonUtils.toJsonBytes`
- `JacksonUtils.toJson1`

**Reason :**

Both functions are using `writeValueasString()` which return Json Formatting string that is not ordered by ObjectMapper, causing nondeterministic behavior in different computer settings and detected by `NonDex`.

**To fix it :**
- `MapperFeature.SORT_PROPERTIES_ALPHABETICALLY` needs to be enabled in the `configure`  function of `ObjectMapper` in class of `JacksonUtils`

- Reorder the testcase in `JacksonUtilsTest.testToJsonBytes1` and `JacksonUtils.toJson1` to be alphabetical by their key value to pass the test.

**To Verify it :** 
`mvn edu.illinois:nondex-maven-plugin:1.1.2:nondex -pl common -Dtest=com.alibaba.nacos.common.utils.JacksonUtilsTest#testToJson1`
**AND**
`mvn edu.illinois:nondex-maven-plugin:1.1.2:nondex -pl common -Dtest=com.alibaba.nacos.common.utils.JacksonUtilsTest#testToJsonBytes1`


KomachiSion (changes requested):
I think the json is not need to sorted, If want to fix test, I think use contains replace equals in test is better.

KomachiSion:
No more response from author, if want fix this problem, please submit other PR with comment's suggestion or more discusion

PR #9103, opened 2022-09-06
Feature #8941 Support Fuzzy Query in Authority Control.

zhouchunhai:

Feature #8941 Support Fuzzy Query in Authority Control.

## Brief changelog

support fuzzy query for user, role, permission management


KomachiSion:
Can you split this PR to at least two PR?

zhouchunhai:
ok, i will submit another PR to fix underscore problem

zhouchunhai:
I revert the second commit, which is meant to fix the underscore problem.But the two commits are associated, so only the first was merged, I can submit another.

KomachiSion:
I see this PR still contain the ui and server plugin, I want you split this PR at lease two PR(one is server plugin api change, and other is ui. If you can split server plugin change to more PR will be better).

Issue #9282, opened 2022-10-09
Remove redundant judgments in ClientWorker


RapperCL:
#com.alibaba.nacos.client.config.impl.ClientWorker

1 The cacheMap.get() method will not be empty
<img width="672" alt="image" src="https://user-images.githubusercontent.com/44110731/194753184-d7f20db8-aa9c-47c4-aca6-77eccb02141c.png">

2 configChangeBatchListenResponse will not be empty
<img width="715" alt="image" src="https://user-images.githubusercontent.com/44110731/194753204-5e157f9c-89ca-4a1e-a929-bdfd38e780e9.png">

PR #9284, opened 2022-10-09
[ISSUE #9282]remove redundant judgment

RapperCL:
[issue](https://github.com/alibaba/nacos/issues/9282)

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/config/impl/ClientWorker.java
+++ b/client/src/main/java/com/alibaba/nacos/client/config/impl/ClientWorker.java
@@ -425,8 +425,8 @@ public class ClientWorker implements Closeable {
     }
     
     private void refreshContentAndCheck(String groupKey, boolean notify) {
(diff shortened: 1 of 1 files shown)


KomachiSion (changes requested):
How about `cacheMap.get()` is null or `configChangeBatchListenResponse` is null?

RapperCL:
> How about `cacheMap.get()` is null or `configChangeBatchListenResponse` is null?

1  As a global variable, cacheMap is initialized when it is created.
` private final AtomicReference<Map<String, CacheData>> cacheMap = new AtomicReference<>(new HashMap<>());`

RapperCL:
2 requestProxy()-->rpcClientInner.request
image
If it is null, a null exception will be thrown in advance, so it will not return null

You can also see that there is no null judgment in other places where this method is called.
for example
image

KomachiSion (approved):

PR #9296, opened 2022-10-11
[ISSUE #8619] add CustomEnvironmentPlugin (#8619)

godhth:
* feature issue #8619 .

* custom properties.

add custom  environment plugin


sys/src/main/java/com/alibaba/nacos/sys/env/EnvUtil.java
```diff
@@ -36,159 +39,182 @@
...
+import java.util.Arrays;
 
 /**
  * Its own configuration information manipulation tool class.
  *
  * @author <a href="mailto:liaochuntao@live.com">liaochuntao</a>
  */
 public class EnvUtil {
-    
+
```
KomachiSion, line 56: please use nacos code style to reformat codes

plugin-default-impl/src/main/java/com/alibaba/nacos/plugin/environment/impl/NacosCustomEnvPluginService.java
```diff
@@ -0,0 +1,58 @@
...
+import com.alibaba.nacos.plugin.environment.spi.CustomEnvironmentPluginService;
+
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Nacos default auth plugin service implementation.
+ *
+ * @author xiweng.yy
```
KomachiSion, line 30: author error

KomachiSion (changes requested):

sys/src/main/java/com/alibaba/nacos/sys/env/EnvUtil.java
godhth: ok👌

plugin-default-impl/src/main/java/com/alibaba/nacos/plugin/environment/impl/NacosCustomEnvPluginService.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Nacos default auth plugin service implementation.
+ *
+ * @author huangtianhui
+ */
+
+public class NacosCustomEnvPluginService implements CustomEnvironmentPluginService {
```
KomachiSion, line 33: Should we change the plugin name like databasePasswordDecryptPluginService ?

plugin-default-impl/src/main/java/com/alibaba/nacos/plugin/environment/impl/NacosCustomEnvPluginService.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Override
+    public Map<String, Object> customValue(Map<String, Object> property) {
+        String pwd = (String) property.get("db.password.0");
```
KomachiSion, line 37: So only handle `0` db? if I set multiple db like `1`, `2`. Don't solve them?

plugin-default-impl/src/main/java/com/alibaba/nacos/plugin/environment/impl/NacosCustomEnvPluginService.java
godhth: I don't think so，Because it's not just about encrypting the database password, it's about customizing all the environment variables

plugin-default-impl/src/main/java/com/alibaba/nacos/plugin/environment/impl/NacosCustomEnvPluginService.java
godhth: This is just an example, you can select "db.password.0","db.password.1" or any other environment variable to customize

godhth: Is aimed at environment rather than just the database password

console/src/main/java/com/alibaba/nacos/console/config/ConsoleConfig.java
```diff
@@ -51,6 +52,7 @@ public class ConsoleConfig {
      */
     @PostConstruct
     public void init() {
+        EnvUtil.customEnvironment();
```
KomachiSion, line 55: I has a problem about this. Should we move this to nacos-core module?

KomachiSion (changes requested):

console/src/main/java/com/alibaba/nacos/console/config/ConsoleConfig.java
godhth: The purpose of this is to execute after loading application.properties and before initializing the controller.

I haven't yet found where nacos-core can achieve this effect.

If you do, you can move it to nacos-core.

KomachiSion: I'm not sure, I just think about the feature will no effect when users not start with console. The custom method will not be called.

godhth: Okay, I'll try .

godhth: down

sys/src/main/java/com/alibaba/nacos/sys/env/EnvUtil.java
```diff
@@ -99,15 +104,36 @@ public class EnvUtil {
...
     
     private static ConfigurableEnvironment environment;
-    
+
+    /**
+     * customEnvironment.
+     */
+    public static void customEnvironment() {
+        Boolean enableCustom = getProperty(NACOS_CUSTOM_ENVIRONMENT_ENABLED, Boolean.class);
+        if (Boolean.TRUE.equals(enableCustom)) {
```
KomachiSion, line 125: `if (enableCustom) ` is enough?

KomachiSion (changes requested):

sys/src/main/java/com/alibaba/nacos/sys/env/EnvUtil.java
godhth: Yes, that's enough. I'll fix it

KomachiSion:
```
boolean enableCustom = getProperty(NACOS_CUSTOM_ENVIRONMENT_ENABLED, Boolean.class, false);
if (enableCustom)
```
I remember EnvUtil has method with Classes and default value.

KomachiSion (changes requested):

sys/src/main/java/com/alibaba/nacos/sys/env/EnvUtil.java
godhth: Yes, there is, but I don't think there is much difference between the two methods.

KomachiSion (dismissed):

KomachiSion (approved):

PR #9365, opened 2022-10-21
modify table his_config_info column group_id length to 255

503612012:
modify table his_config_info column group_id length to 255


KomachiSion:
Please discuss in #9334 first, community should decide change to 128 or 255.

KomachiSion:
Should set length as 128 and fixed by https://github.com/alibaba/nacos/pull/9412

Issue #9391, opened 2022-10-25
chooser optimization reduces array usage and redundant for loops


RapperCL:
path:com.alibaba.nacos.client.naming.utils.Chooser

chooser optimization :
reduces array usage and redundant for loops
1 The first for loop stores the nodes whose probability has been calculated through the double array;

2 The second for loop calculates the overall probability distribution array based on the probability of the first step;
The first double array is not needed, it can be replaced directly with a variable, and the two for loops should be combined into one.

![image](https://user-images.githubusercontent.com/44110731/197723329-d2559bfd-7b57-4861-8029-0545912a6336.png)

KomachiSion:
I read this codes, I think this array size might be `index` no items.size()

RapperCL:
> I read this codes, I think this array size might be `index` no items.size()

Supplement the screenshot after the modification. In fact, the change in general is that the original exactWeights array is replaced by the variable exactWeight, and then the two for loops are merged into one
补充上修改之后的截图，其实改动总的来说就是，将原本的exactWeights数组通过变量exactWeight替代了，然后将两个for循环合并为一个了
![image](https://user-images.githubusercontent.com/44110731/198182429-44118785-0714-48ca-b52c-6f76ea4046fc.png)

RapperCL:
@KomachiSion  hey bro, what do you think about this now

KomachiSion:
我知道你的改动意思， 我回复的意思是我读代码感觉后一个数组的长度不应该是item.size，
如果后一个数组的长度不应该是item.size，那么你就不能按照你这个修改来改动

RapperCL:
> 我知道你的改动意思， 我回复的意思是我读代码感觉后一个数组的长度不应该是item.size，
> 如果后一个数组的长度不应该是item.size，那么你就不能按照你这个修改来改动

嗯，目前而言，后一个数组的长度一定是item.size， 但是你的考虑是对的，我重新修改了一下，同时对构造方法做了一点
增强，一方面避免空异常，另一方面让itemsWithWeight这个集合的初始化有了意义。
<img width="530" alt="image" src="https://user-images.githubusercontent.com/44110731/198929758-197cc21c-f79e-4387-85c8-271abc725618.png">

PR #9392, opened 2022-10-25
[ISSUE #9391] Chooser optimization

RapperCL:
[issue](https://github.com/alibaba/nacos/issues/9391)


client/src/main/java/com/alibaba/nacos/client/naming/utils/Chooser.java
```diff
@@ -143,23 +143,20 @@ public void refresh() {
...
             for (Pair<T> item : itemsWithWeight) {
                 double singleWeight = item.weight();
                 //ignore item which weight is zero.see test_randomWithWeight_weight0 in ChooserTest
                 if (singleWeight <= 0) {
                     continue;
                 }
-                exactWeights[index++] = singleWeight / originWeightSum;
-            }
-            
-            weights = new double[items.size()];
```
KomachiSion, line 158: I read this codes, I think this array size might be `index` no items.size()

KomachiSion (changes requested):

KomachiSion:
I'm contacting the author for this part codes. After making sure the codes meaning, I will response this PR and determine whether change or how to change this codes. Before making sure, The PR will be pending, Sorry.

PRANRAVEE (approved):

KomachiSion (changes requested):
Can you add some Unit test to ensure your changes has no effect for this method?

RapperCL:
> Can you add some Unit test to ensure your changes has no effect for this method?

OK, I'll add it later

KomachiSion:
> I'm contacting the author for this part codes. After making sure the codes meaning, I will response this PR and determine whether change or how to change this codes. Before making sure, The PR will be pending, Sorry.

I talk about this problem with author, he think the length of weight also should be `index`. So I think after your unit test finished, the PR can be merged.

RapperCL:
> > I'm contacting the author for this part codes. After making sure the codes meaning, I will response this PR and determine whether change or how to change this codes. Before making sure, The PR will be pending, Sorry.
> 
> I talk about this problem with author, he think the length of weight also should be `index`. So I think after your unit test finished, the PR can be merged.

Well, thank you. Interestingly, when I was writing the test case, I found that the chooser would have an exception in a special scenario. In order to prevent confusion, I raised an issue again [issue 9478](https://github.com/alibaba/nacos/issues/9478) and the corresponding pr submitted the solution and test case.

KomachiSion (approved):

Issue #9463, opened 2022-11-04
Flaky test in Common Module caused by Json Ordering Permutations


bmk15897:
**Describe the bug**
The test file
```
common/
com.alibaba.nacos.common.utils.JacksonUtilsTest.
```
has flakiness in the test **testToJsonBytes**

**Expected behavior**
The assertion should be passed whatever the order of the JSON string is.
We use the [NonDex.](https://github.com/TestingResearchIllinois/NonDex) as a test tool and we expected it passed.

**Actually behavior**
Fails while using Nondex to run the test with the following error -
```
[ERROR] Failures: 
[ERROR]   JacksonUtilsTest.testToJsonBytes:460 expected:<{"code":0,"data":{"[string":"你好，中国！","integer":999]}}> but was:<{"code":0,"data":{"[integer":999,"string":"你好，中国！"]}}>
[INFO] 
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0
```
**How to Reproduce**
The flakiness was found with [NonDex.](https://github.com/TestingResearchIllinois/NonDex)
After installing NonDex and run the following code.
```
cd nacos
mvn install -pl common -am -DskipTests
mvn -pl common edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=com.alibaba.nacos.common.utils.JacksonUtilsTest#testToJsonBytes
```
**Desktop (please complete the following information):**
 - OS: Windows
 - Version: 10
 - Maven Version: 3.8.1
 - Nacos Module - common
 - SDK: Original

KomachiSion:
Can you provide a fix solution way?

bmk15897:
Yes, the proposed solution is as follows: (https://github.com/alibaba/nacos/pull/9505 is my pull request for the same.)
There are two points where non-determinism is introduced. Proposed fixes are attached herewith.
1. Using HashMap introduces non-determinism because HashMap doesn't guarantee the order of elements.
Fix: Use a LinkedHashMap instead of a HashMap to guarantee the preservation of the order of insertion.
2. Conversion of the ResultRet object into a JSON string without mentioning the order of the elements in the ResultSet.
Fix: Annotate the class ResultSet with @JsonPropertyOrder(alphabetic = true) to ensure the order of properties while converting the object into JSON.

Can you please review the same and approve?

KomachiSion:
The second way is un-reasonable, the problem is test case problem, so the solution should not change the original codes.

PR #9505, opened 2022-11-08
[ISSUE #9643] Fixed Possible JSON Ordering Permutations Problem in Tests

bmk15897:
Related to issue https://github.com/alibaba/nacos/issues/9463

To fix the flaky test caused by possible JSON ordering permutations as explained in the issue https://github.com/alibaba/nacos/issues/9463.

## Brief changelog
1. Use a LinkedHashMap instead of HashMap to preserve the order of insertion.
2. Use the [@JsonPropertyOrder(alphabetic = true)](https://fasterxml.github.io/jackson-annotations/javadoc/2.8/com/fasterxml/jackson/annotation/JsonPropertyOrder.html#alphabetic()) to set order at the top of the ResultSet class.

## Verifying this change
This test avoids nested or different orders of JSON bytes that cause flaky errors.
There is 1 change in the source code - Annotation of the class ResultSet with [@JsonPropertyOrder(alphabetic = true)](https://fasterxml.github.io/jackson-annotations/javadoc/2.8/com/fasterxml/jackson/annotation/JsonPropertyOrder.html#alphabetic()) to ensure the order of properties while converting the object into a JSON string. The other change is in the test code, which changes HashMap to LinkedHashMap.
The changes passed the local tests that the guidelines mentioned.

### DIFF
--- a/common/src/main/java/com/alibaba/nacos/common/model/RestResult.java
+++ b/common/src/main/java/com/alibaba/nacos/common/model/RestResult.java
@@ -16,6 +16,7 @@
 
 package com.alibaba.nacos.common.model;
 
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import java.io.Serializable;
 
 /**
@@ -23,6 +24,7 @@ import java.io.Serializable;
  *
  * @author <a href="mailto:liaochuntao@live.com">liaochuntao</a>
  */
+@JsonPropertyOrder(alphabetic = true)
 public class RestResult<T> implements Serializable {
     
     private static final long serialVersionUID = 6095433538316185017L;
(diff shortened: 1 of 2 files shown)


li-xiao-shuang (approved):

KomachiSion (changes requested):
No reasonable fix.

The way should change the test case , not change original code.

bmk15897:
I understand. I will attempt to fix the flaky test without changing the original code.

KomachiSion:
I think we can use contain to replace equals, for example, we can assertTrue `actual.contains("\"code\":0")` to replace assertEquals with expected and actual.

It will not depend order again.

KomachiSion (approved):

PR #9552, opened 2022-11-11
perf: Replace the dead loop with while

zhantiao:
Because this is a conditional loop, it's better to use while, and it also simplifies the code

### DIFF
--- a/common/src/main/java/com/alibaba/nacos/common/notify/DefaultPublisher.java
+++ b/common/src/main/java/com/alibaba/nacos/common/notify/DefaultPublisher.java
@@ -102,18 +102,12 @@ public class DefaultPublisher extends Thread implements EventPublisher {
             int waitTimes = 60;
             // To ensure that messages are not lost, enable EventHandler when
             // waiting for the first Subscriber to register
-            for (; ; ) {
-                if (shutdown || hasSubscriber() || waitTimes <= 0) {
-                    break;
-                }
+            while (!shutdown && !hasSubscriber() && waitTimes > 0) {
                 ThreadUtils.sleep(1000L);
                 waitTimes--;
             }
-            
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/alibaba/nacos/common/notify/DefaultPublisher.java
```diff
@@ -102,18 +102,12 @@ void openEventHandler() {
             int waitTimes = 60;
             // To ensure that messages are not lost, enable EventHandler when
             // waiting for the first Subscriber to register
-            for (; ; ) {
-                if (shutdown || hasSubscriber() || waitTimes <= 0) {
-                    break;
-                }
+            while (!shutdown && !hasSubscriber() && waitTimes > 0) {
```
li-xiao-shuang, line 105: The logic here doesn't seem right

KomachiSion (dismissed):

zhantiao:
image
My pr failed the integration test, I want to know why? Do I need to be concerned about this?

common/src/main/java/com/alibaba/nacos/common/notify/DefaultPublisher.java
```diff
@@ -102,18 +102,12 @@ void openEventHandler() {
             int waitTimes = 60;
             // To ensure that messages are not lost, enable EventHandler when
             // waiting for the first Subscriber to register
-            for (; ; ) {
-                if (shutdown || hasSubscriber() || waitTimes <= 0) {
```
Rigavinil201, line 106: >

zhantiao (commented):
我准备把 CredentialWatcher.java 这个改动另外提交一个新PR

KomachiSion (approved):

PR #9709, opened 2022-12-14
support list instances which is unenabled.

### LINKED ISSUE #9703: can "/nacos/v2/ns/instance/list" return instance record which  enabled was false?
I noticed that the endpoint “/nacos/v2/ns/instance/list” which Open-API offered only return enabled instance, if return all instances will be fantastic for me, Any plan to support that feature?

YunWZ:
fix #9703.


KomachiSion:
It‘s a unreasonable feature request. Close PR. If has some discussion please comment in issue.

PR #9719, opened 2022-12-16
The requestMappingHandlerMapping conflict will occur during startup

twelvet-s:
bean but found 2: requestMappingHandlerMapping,controllerEndpointHandlerMapping


Salhazry (commented):
Salhazry@gmail.com

PR #9813, opened 2023-01-07
Extract the cluster of AbstractClusterRequest as a constant

sixsixsix516:

To unify the code style.

Because I see that the implementation of the subclasses of com. alibaba. nacos. api. remote. request. Request # getModule uses the constants defined in the Constants, and only AbstractClusterRequest is a constant directly defined in the current class, which is inconsistent with the system code style.

因为我看见 com.alibaba.nacos.api.remote.request.Request#getModule 的子类的实现都是使用的 Constants 中定义的常量，而只有AbstractClusterRequest 是直接在当前类定义的一个常量，这与系统现代码风格不符

## Brief changelog
com.alibaba.nacos.naming.cluster.remote.request.AbstractClusterRequest#getModule 的返回值定义到了 

com.alibaba.nacos.api.common.Constants.Cluster#CLUSTER_MODULE

## Verifying this change

XXXX

### DIFF
--- a/api/src/main/java/com/alibaba/nacos/api/common/Constants.java
+++ b/api/src/main/java/com/alibaba/nacos/api/common/Constants.java
@@ -245,6 +245,14 @@ public class Constants {
     }
     
     /**
+     * The constants in Cluster directory.
+     */
+    public static class Cluster {
+        
+        public static final String CLUSTER_MODULE = "cluster";
+    }
+    
+    /**
      * The constants in exception directory.
      */
     public static class Exception {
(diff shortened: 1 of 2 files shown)


The-Gamer-01 (approved):

KomachiSion (changes requested):
I can't get the change reason, the constants only one place used, I think should be local constants not common constants.

KomachiSion:
Other module like `naming` and `config`, the constants will be used by sub module with a large of number. So use `Constants`, but `cluster` is only for this sub-request, so no need to use `Constants`.

The `Constants` should contain constants which may be used by many different classes. For some of local constants, they should be defined in class itself.

Issue #9937, opened 2023-02-13
when oldService.getLastRefTime() > serviceInfo.getLastRefTime()，The serviceInfo should not be updated


xuziyang:
**Describe the bug**
branch：v1.x-develop
com.alibaba.nacos.client.naming.core.HostReactor：line 190
![image](https://user-images.githubusercontent.com/8465969/218401075-b4e39697-e79c-46b4-a89d-8e12a0cef2d1.png)

**Expected behavior**
Return directly to the oldService

**Actually behavior**
serviceInfo performs the update

KomachiSion:
Which mean the local cache is newer than current data from server , so should ignore it.

KomachiSion:
Sorry, I make a mistake. I will make sure whether return old data.

KomachiSion:
I ask the original author, the answer is should not return old one.

The reason is that the RefTime is server defined, if the RefTime has some problem, like the time clock error. The old RefTime will large than new data for a long time, which will make the data not updated.

So only print log to trace the problem, but the cached data should be update also.

PR #9938, opened 2023-02-13
The client receives an old servicerInfo and should not perform the update

xuziyang:
fix https://github.com/alibaba/nacos/issues/9937


KomachiSion:
Close, reason in issue.

PR #10007, opened 2023-02-24
Specification partial code

4yDX3906:

1.The if statement should be followed by parentheses.
2.The equals method of Object is prone to throwing null pointer exceptions. equals should be called using constants or objects that have definite values.


KomachiSion (changes requested):
grpc/auto package is auto generate from protobuf. No need to change.

4yDX3906:
> grpc/auto package is auto generate from protobuf. No need to change.

Issue #10018, opened 2023-02-28
Utility classes should not have public constructors


4yDX3906:
**Describe the bug**
Utility classes, which are collections of static members, are not meant to be instantiated.   Even abstract utility classes, which can be extended, should not have public constructors.
Java adds an implicit public constructor to every class which does not define at least one explicitly.   Hence, at least one non-public constructor should be defined.

I found the com/alibaba/nacos/common/utils package under classes have a lot of Utility classes without private structure constructor.

**Reference material**
![image](https://user-images.githubusercontent.com/89574863/221809969-f65dcf4a-0d9c-4920-bb6b-27c150b5922d.png)

KomachiSion:
when you add the private constructor, how to do unit test the constructor?

I have no idea for this rule, the public constructor has no effect.

PR #10019, opened 2023-02-28
fix:Utility classes should not have public constructors

4yDX3906:

Utility classes should not have public constructors.

## Brief changelog

To com/alibaba/nacos/common/utils package under the utility class add  private constructor.

## Associated issue
#10018

### DIFF
--- a/common/src/main/java/com/alibaba/nacos/common/utils/AbstractAssert.java
+++ b/common/src/main/java/com/alibaba/nacos/common/utils/AbstractAssert.java
@@ -58,6 +58,9 @@ import java.util.function.Supplier;
  */
 public abstract class AbstractAssert {
 
+    private AbstractAssert() {
+    }
+
     /**
      * Assert a boolean expression, throwing an {@code IllegalStateException}
      * if the expression evaluates to {@code false}.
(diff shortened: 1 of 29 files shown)


KomachiSion (changes requested):
when you add the private constructor, how to do unit test the constructor?

I have no idea for this `rule`, the public constructor has no effect.

4yDX3906:
> when you add the private constructor, how to do unit test the constructor?
> 
> I have no idea for this `rule`, the public constructor has no effect.



> when you add the private constructor, how to do unit test the constructor?
> 
> I have no idea for this `rule`, the public constructor has no effect.

A utility class is a class that is just a namespace for functions.  No instances of it can exist,   and all its members are static. I don't think I need to unit test its constructor. Just make sure the static methods  are available. For example,  `java.lang.Math` and `java.util.Collections` are well known utility classes. Their constructor is private.

KomachiSion:
It seems auto ignore private constructor after jacoco 0.8.0.

https://github.com/jacoco/jacoco/wiki/FilteringOptions

we use 0.8.7.

KomachiSion (approved):

PR #10343, opened 2023-04-19
Specify the GC as the CMS when jdk version less 9.

aloneboat:

When Jdk version less than 9 , for example 8, the GC default is Parallel GC，the application threads will be pause for long time (as I understand it) ,that is not we want。
I find this change [Support JDK 11](https://github.com/alibaba/nacos/commit/b6e1ad91b8700facc966e01b7b5e3830e9f87401)，recovered some code  

## Brief changelog

Specify the GC as the CMS when jdk version less that 9.

### DIFF
--- a/distribution/bin/startup.sh
+++ b/distribution/bin/startup.sh
@@ -108,6 +108,7 @@ JAVA_MAJOR_VERSION=$($JAVA -version 2>&1 | sed -E -n 's/.* version "([0-9]*).*$/
 if [[ "$JAVA_MAJOR_VERSION" -ge "9" ]] ; then
   JAVA_OPT="${JAVA_OPT} -Xlog:gc*:file=${BASE_DIR}/logs/nacos_gc.log:time,tags:filecount=10,filesize=100m"
 else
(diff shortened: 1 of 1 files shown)


KomachiSion (approved):

iJIAJIA:
看了下gc配置, 有个疑惑点想讨论下, 这里为什么去除 -XX:-UseParNewGC. 我这边8c20g的3节点集群,跟默认使用UseParNewGC进行了3000的并发心跳压测对比, 吞吐量有10k左右的差距. 同时gc的频率也有差距. 
启动时jvm也给了警告. 不推荐 cms 跟DefNew 搭配使用.

iJIAJIA:
image
image
看下来cpu使用率是降下来了, 但这个感觉是因为gc频次高导致吞吐量上不去

iJIAJIA:
image
image
对比下来, cms + ParNewGC的gc频次是远低于 cms + DefNew 的

iJIAJIA:
@aloneboat

aloneboat:
@iJIAJIA Sorry for the late reply. you should be right, I just restored the earlier code，the purpose is simply to use the CMS garbage collector when jdk version , I didn't look closely at whether it was a plus or a minus before "ParNewGC",actually I agree with you.

PR #10631, opened 2023-06-12
Remove disable "UseParNewGC" option when use CMS

aloneboat:

In this change [Specify the GC as the CMS when jdk version less 9](https://github.com/alibaba/nacos/pull/10343),  I just restored the earlier code,but I  didn't look closely at whether it was a plus or a minus before "UseParNewGC", @iJIAJIA  points out the problem.

## Brief changelog

Specify the GC as the CMS when jdk version less than 9.


KomachiSion:
There is conflict for your PR, please fix them first

KomachiSion (approved):

Issue #9302, opened 2022-10-11
将配置清理过期历史配置的逻辑插件化


realJackSun:
# 背景

目前配置模块的功能耦合性比较强，后面希望做出来多款插件，将配置内部数据流转的各个过程解耦。
本次先将配置清理过期历史配置的逻辑抽象成一个插件。

# 方案设计
// TODO

# 时间节奏

10月31日发布

Daydreamer-ia:
清除过期历史配置好像是在 `DumpService` 里用定时任务完成的，“配置清理过期历史配置的逻辑”是指具体调用清除的逻辑吗？如果集群不同节点运行不同插件会不会有问题？

vividfish:
我可以尝试下

realJackSun:
@vividfish welcome

HACKED-need-HELP:
For the past 3 years ive had people hacking/ stalking me and my children.  Every electronic we got these people some how got into them.  These people have all out personal information, they have stolen 30, 000 in these 3 years. Yes ive bought over 30 phone and electronic for me and my children. Im a single mother.  These people jave threatened to kill us. I dont know these people i dont think. I HAVE NO IDEAL HAVE THEY GOT ALL OUR INFORMATION BUT THEY HAVE DESTROYED ME. AMD MY Kids have been so scared for the past 3 years.  We have had naked pictures took of us and put on the inernet. And now o have tracked them down to an group app on this app.  So does anyone know MARTIN MAYAN? HES THE MAIN ONE with all his little followers behind him.

PR #10654, opened 2023-06-15
Feature/plugin cleaner config history sl ready to push

realJackSun:
fix https://github.com/alibaba/nacos/issues/9302


KomachiSion:
Not good design and implementation.

The jdbcTemplate directly import to plugin, which will cause the plugins consider the database type. It's hard to implement plugin.

Please discuss the design more.

Issue #10717, opened 2023-07-01
There is a serialization issue with com.alibaba.nacos.common.utils.JacksonUtils#toJson.


Bo-Qiu:
**Describe the bug**
There seems to be a bug in the toJson method of the com.alibaba.nacos.common.utils.JacksonUtils class. When the method is called with a parameter that is a JSON-formatted string, the return value is no longer a JSON-formatted string. This method is invoked in the com.alibaba.nacos.common.http.client.request.JdkHttpClientRequest#execute method, where the requestHttpEntity.getBody() is parsed into a string. If the body itself is already a JSON-formatted string, it causes the bodyStr to no longer be a JSON-formatted string. Consequently, this can lead to exceptions when making a request with Content-Type=application/json, such as in the com.alibaba.nacos.common.http.client.NacosRestTemplate#postJson method.

**Expected behavior**
In the com.alibaba.nacos.common.utils.JacksonUtils#toJson method, if the parameter is a String, it should be returned directly as the result without any additional parsing.

**Actually behavior**
When the method is called with a parameter that is a JSON-formatted string, the return value is no longer a JSON-formatted string.

**How to Reproduce**
It can be reproduced using the code as follow:
```
 /**
     * When the parameter is a String in Json format, the returned result will no longer be a String in Json format.
     * The parsed String cannot be parsed into an object by calling the com.alibaba.nacos.common.utils.JacksonUtils#toObj(java.lang.String) method
     */
    @Test
    public void testToJson3() {
        String originalJson = "{\"user\":\"qb\",\"age\":18}";
        String parsedJson = JacksonUtils.toJson(originalJson);
        System.out.println("originalJson:    " + originalJson);
        System.out.println("parsedJson:    " + parsedJson);
        JacksonUtils.toObj(originalJson, HashMap.class);
        JacksonUtils.toObj(parsedJson, HashMap.class);
    }
```

Bo-Qiu:
Please review this issue, and if it is indeed a bug, kindly assign it to me, and I will fix it as soon as possible. Thank you.

horizonzy:
Close via #10718

PR #10718, opened 2023-07-01
fix(#10717): com.alibaba.nacos.common.http.client.request.JdkHttpClie…

Bo-Qiu:

fix the issue  #10717 

## Brief changelog

Actually, there is a serialization issue with the com.alibaba.nacos.common.utils.JacksonUtils#toJson method. When the parameter is a JSON-formatted string, the returned result is no longer a JSON-formatted string. This causes a problem in the com.alibaba.nacos.common.http.client.request.JdkHttpClientRequest#execute method when the body is a JSON string and the Content-Type is set to application/json. It leads to incorrect parsing of the body, causing the server-side to be unable to properly interpret the body. Ideally, the adjustment should be made internally in the com.alibaba.nacos.common.utils.JacksonUtils#toJson method. However, considering that this method is public, directly modifying it may affect users who rely on this method. Therefore, the adjustment will only be made to the com.alibaba.nacos.common.http.client.request.JdkHttpClientRequest#execute method to fix this issue. If necessary, a separate pull request will be submitted later to fix the com.alibaba.nacos.common.utils.JacksonUtils#toJson method.


horizonzy (approved):
LGTM

Issue #10727, opened 2023-07-04
When the doPushWithCallback push task fails, it no longer retries forever, but only the default number of retries


Bo-Qiu:
**Is your feature request related to a problem? Please describe.**

For the same pushTask, the com.alibaba.nacos.naming.push.v2.executor.PushExecutor#doPushWithCallback method will not retry forever after onFail, but will only retry the default number of times. 
When the same pushTask fails and retries the default number of doPush times, the task will not be added back to the delayTaskEngine's task queue. In this way, you can avoid repeatedly pushing invalid tasks that are destined to fail, and you can even avoid the problem of server OOM exceptions caused by a large number of invalid tasks that cannot be successfully removed.

**Describe the solution you'd like**

DoPushWithCallback will retry after failing to push the same task. If the task fails every retry, it will retry forever. This will waste resources on invalid tasks that are bound to fail, and at the same time, when a large number of invalid tasks appear, it will cause the server to OOM.

After doPushWithCallback fails to push the same task, it will only perform the default number of retries. After the default number of times is exceeded, even if the task still fails to be pushed, it will not be added to the task queue.

**Describe alternatives you've considered**

After doPushWithCallback fails to push the same task, it will only perform the default number of retries. After the default number of times is exceeded, even if the task still fails to be pushed, it will not be added to the task queue.

Users can customize the number of retries. If not configured, the default value will be used. The default value is 3.

When the number of times configured by the user is -1, it is consistent with the logic of the current version. If the task push fails, it will keep retrying until the task push succeeds.

Bo-Qiu:
If this feature is necessary, I'll take it over and implement it as soon as possible, please assign it to me，thx!

KomachiSion:
I see the issue, but the request is not reasonable.

The push check will see the subsribers and connection whether existed. So It is not push forever to client and not cause OOM.

If the connection and subsribers existed but the push continuely failed. Please `make sure` your subsribers is ok and network is ok first.

If there is limit of push count, the data in subsribers will different with server, which make bigger problem.

PR #10731, opened 2023-07-04
feat(#10727): when the doPushWithCallback push task fails, it no longer retries forever, but only the default number of retries

Bo-Qiu:

feat(#10727) When the doPushWithCallback push task fails, it no longer retries forever, but only the default number of retries.

## Brief changelog

DoPushWithCallback will retry after failing to push the same task. If the task fails every retry, it will retry forever. This will waste resources on invalid tasks that are bound to fail, and at the same time, when a large number of invalid tasks appear, it will cause the server to OOM.

After doPushWithCallback fails to push the same task, it will only perform the default number of retries. After the default number of times is exceeded, even if the task still fails to be pushed, it will not be added to the task queue.

## Verifying this change
After doPushWithCallback fails to push the same task, it will only perform the default number of retries. After the default number of times is exceeded, even if the task still fails to be pushed, it will not be added to the task queue.

Users can customize the number of retries. If not configured, the default value will be used. The default value is 3.

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/push/v2/task/PushDelayTask.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/push/v2/task/PushDelayTask.java
@@ -36,6 +36,8 @@ public class PushDelayTask extends AbstractDelayTask {
     
     private Set<String> targetClients;
     
+    private int retriedCount = 0;
+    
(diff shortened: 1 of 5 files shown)


KomachiSion:
Reject reason in issue.

Issue #10750, opened 2023-07-08
public namespaceId should directly shown on console


Joey777210:
**Is your feature request related to a problem? Please describe.**
Since we use namespaceId to determine a namespace , we should clearly show namespace ids on console.
Currently namespace id of 'public' is same to show name, but users do not know. 
![image](https://github.com/alibaba/nacos/assets/53630996/a600d293-664c-4e2a-8d1d-1627d7e24288)

![image](https://github.com/alibaba/nacos/assets/53630996/22fb2f57-1319-4f9b-90ab-c02b14a435fe)

**Describe the solution you'd like**
Show public's namespace id on console.

KomachiSion:
The public id should be `` for now. Only naming module will do more handle to transform `` to `public`. And This `GetDefaultNamespace` is datasource plugin added, and no more usage. 

How to solve the `public` namespace problem is discussion in #9846. But no result for now.

wilsonwu:
Yes, the public namespace id is empty string, and if you use public as Id to register service, it works fine, so there are confusing for users, it is a legercy old problem.

Joey777210:
> The public id should be '' for now. Only naming module will do more handle to transform '' to `public`. And This `GetDefaultNamespace` is datasource plugin added, and no more usage.
> 
> How to solve the `public` namespace problem is discussion in #9846. But no result for now.

Then how to determine public namespace by namespaceId? Empty param seems not right..

PR #10751, opened 2023-07-08
[issue10750]Feature/issue 10750

Joey777210:

show public namespace id on console


Joey777210:
#10750

KomachiSion:
The public id should be `` for now. Only naming module will do more handle to transform `` to `public`. And This `GetDefaultNamespace` is datasource plugin added, and no more usage. 

How to solve the `public` namespace problem is discussion in #9846. But no result for now.

Issue #10223, opened 2023-03-29
Instance id is null and may cause trouble with spring cloud integration


liubao68:
**Describe the bug**

When integrate with spring cloud, e.g.  `spring-cloud-starter-alibaba-nacos-discovery`,  I found getInstanceId is always null. 
And find the following api of nacos do not return the expected instance id. 

```
List<Instance> selectInstances(String serviceName, String groupName, boolean healthy)
```

**Expected behavior**
selectInstances should return the correct instance id. 

**Actually behavior**
 instance id is null. 

**How to Reproduce**
using nacos 2.2.1 and  `spring-cloud-starter-alibaba-nacos-discovery`  2021.0.5.0

**Desktop (please complete the following information):**
NA

**Additional context**
NA

KomachiSion:
InstanceId should users set when register Instance. 

If want to generate instanceId, welcome submit PR in `InstanceRequestHandler`

KomachiSion:
It can be do with https://github.com/alibaba/nacos/issues/5680

liubao68:
I'd agree with https://github.com/alibaba/nacos/issues/5680 that client generate id is not a good practice, users need to consider distributed configurations.  But still thanks this can help me to solve the problem.

PR #10784, opened 2023-07-13
feat(#5608 and #10223): When the custom instance id is empty, the id will be automatically generated.

Bo-Qiu:

feat(#5608  and #10223 ): When the custom instance id is empty, the id will be automatically generated.

## Brief changelog

When the custom instance id is empty, the id will be automatically generated.

```
    public static String generateInstanceId(Instance instance) {
        return instance.getIp() + InternetAddressUtil.IP_PORT_SPLITER + instance.getPort() + "_"
                + System.currentTimeMillis();
    }
```

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/constants/Constants.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/constants/Constants.java
@@ -76,9 +76,9 @@ public final class Constants {
     public static final long ACK_TIMEOUT_NANOS = TimeUnit.SECONDS.toNanos(10L);
     
     /**
-     * The custom instance id key.
+     * The instance id key.
      */
-    public static final String CUSTOM_INSTANCE_ID = "customInstanceId";
+    public static final String INSTANCE_ID = "instanceId";
     
     /**
      * The weight of instance according to instance self publish.
--- a/naming/src/test/java/com/alibaba/nacos/naming/core/v2/service/ClientOperationServiceTest.java
+++ b/naming/src/test/java/com/alibaba/nacos/naming/core/v2/service/ClientOperationServiceTest.java
@@ -1,0 +1,48 @@
+/*
+ *  Copyright 1999-2018 Alibaba Group Holding Ltd.
+ *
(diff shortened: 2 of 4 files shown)


Bo-Qiu:
Is it necessary to use the snowflake algorithm to generate an instance ID that is basically guaranteed to be globally unique?

naming/src/test/java/com/alibaba/nacos/naming/core/v2/service/ClientOperationServiceTest.java
```diff
@@ -0,0 +1,48 @@
...
+public class ClientOperationServiceTest {
+    
+    @Test
+    public void getPublishInfoAutoGenerateInsId() {
+        ClientOperationServiceProxy clientOperationService = new ClientOperationServiceProxy(null, null);
+        Instance instance = new Instance();
+        String customInsId = "1.1.1.1_custom";
+        instance.setInstanceId(customInsId);
+        InstancePublishInfo publishInfo1 = clientOperationService.getPublishInfo(instance);
+        assert customInsId.equals(publishInfo1.getExtendDatum().get(Constants.INSTANCE_ID));
```
horizonzy, line 42: Please use Assert.equals()

naming/src/main/java/com/alibaba/nacos/naming/constants/Constants.java
```diff
@@ -76,9 +76,9 @@ public final class Constants {
     public static final long ACK_TIMEOUT_NANOS = TimeUnit.SECONDS.toNanos(10L);
     
     /**
-     * The custom instance id key.
+     * The instance id key.
      */
-    public static final String CUSTOM_INSTANCE_ID = "customInstanceId";
+    public static final String INSTANCE_ID = "instanceId";
```
horizonzy, line 81: I'm not sure that is this a breaking change. If we change the `public static` variable, the previous user upgrade to the new version, are there some differeces?

Bo-Qiu:
After discussing with zy, we decided to implement the logic of automatically generating IDs on the client side. 
@horizonzy @KomachiSion 
I'll open a new PR.

PR #10796, opened 2023-07-15
feat(#5608): When the custom instance id is empty, the id will be automatically generated.

Bo-Qiu:

feat(#5608): When the custom instance id is empty, the id will be automatically generated.

When the client initiates a request to register an instance, it will determine whether the instance ID is empty. If it is empty, the instance ID will be automatically generated and set.

```
public void setInstanceIdIfNeeded() {
        if (StringUtils.isEmpty(this.instanceId)) {
            long curIndex = index.getAndIncrement();
            this.instanceId = ip + Constants.COLON + port + Constants.NAMING_INSTANCE_ID_SPLITTER + curIndex
                    + Constants.NAMING_INSTANCE_ID_SPLITTER + System.currentTimeMillis();
        }
    }
```


Bo-Qiu:
According to our discussion, I rewrote the code and submitted a PR, please review it when you are free. @horizonzy @KomachiSion

KomachiSion:
Please do this feature in server like com.alibaba.nacos.naming.pojo.instance.HttpRequestInstanceBuilder#setInstanceId

KomachiSion (changes requested):
Wrong fixed way.

Issue #10539, opened 2023-05-24
当配置增删改查失败的时候，留下操作失败的ERROR日志


realJackSun:
![image](https://github.com/alibaba/nacos/assets/58767027/5162ffe7-ad94-4650-8265-81677bce2e2e)
写入数据库失败时，会导致：
![image](https://github.com/alibaba/nacos/assets/58767027/69d31c8b-6d3d-4b12-9f2b-db8fc1aa0d8f)
当有如上图所示发布配置失败时，留下ERROR日志

realJackSun:
@i will solve it@

wilsonwu:
This is a good first issue :)

realJackSun:
> This is a good first issue :)

Not first issue, hh

wilsonwu:
> > This is a good first issue :)
> 
> Not first issue, hh

Good first issue means anyone want to contribute this project can be start from this issue without any experience.

KomachiSion:
@realJackSun Please solve #10353 together, the catch exception has both problem in DuplicateKeyException and DataIntegrityViolationExcetion

PR #10804, opened 2023-07-16
feat(#10539): When the operation configuration fails, log.

Bo-Qiu:

feat(#10539): When the operation configuration fails, log.


KomachiSion:
Please reuse the log ( config-server.log)

Bo-Qiu:
> Please reuse the log ( config-server.log)

Thanks for your suggestion, I have adjusted the code, please review it again.

PR #10812, opened 2023-07-17
feat(#5608 && #10223): When the custom instance id is empty, the id will be automatically generated.

Bo-Qiu:

feat(#5608 && #10223): When the custom instance id is empty, the id will be automatically generated.

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/remote/rpc/handler/InstanceRequestHandler.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/remote/rpc/handler/InstanceRequestHandler.java
@@ -29,6 +29,7 @@ import com.alibaba.nacos.common.trace.event.naming.RegisterInstanceTraceEvent;
 import com.alibaba.nacos.core.remote.RequestHandler;
 import com.alibaba.nacos.naming.core.v2.pojo.Service;
 import com.alibaba.nacos.naming.core.v2.service.impl.EphemeralClientOperationServiceImpl;
+import com.alibaba.nacos.naming.utils.InstanceUtil;
 import com.alibaba.nacos.plugin.auth.constant.ActionTypes;
 import org.springframework.stereotype.Component;
 
@@ -51,6 +52,7 @@ public class InstanceRequestHandler extends RequestHandler<InstanceRequest, Inst
     public InstanceResponse handle(InstanceRequest request, RequestMeta meta) throws NacosException {
         Service service = Service
                 .newService(request.getNamespace(), request.getGroupName(), request.getServiceName(), true);
+        InstanceUtil.setInstanceIdIfEmpty(request.getInstance(), service.getGroupedServiceName());
         switch (request.getType()) {
             case NamingRemoteConstants.REGISTER_INSTANCE:
                 return registerInstance(service, request, meta);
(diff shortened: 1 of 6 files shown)


Bo-Qiu:
Based on our discussions, I re-wrote the code and submitted a new PR, please review it again, thx!
@KomachiSion

KomachiSion (changes requested):
Please use DefaultInstanceIdGenerator to generate default Id like http register

KomachiSion:
com.alibaba.nacos.naming.pojo.instance.HttpRequestInstanceBuilder#setInstanceId

Bo-Qiu:
> Please use DefaultInstanceIdGenerator to generate default Id like http register

In fact, my implementation is consistent with it. could you provide more information? thanks a lot.

image

Bo-Qiu:
> Please use DefaultInstanceIdGenerator to generate default Id like http register

Is your suggestion to directly set the ID using it regardless of whether the customID is empty or not?   Code as follow:
```
private void setInstanceId(Instance instance) {
        DefaultInstanceIdGenerator idGenerator = new DefaultInstanceIdGenerator(instance.getServiceName(),
                instance.getClusterName(), instance.getIp(), instance.getPort());
        instance.setInstanceId(idGenerator.generateInstanceId());
    }
```

I think there are some problems with using it directly. In GRPC mode, the filed serviceName of instance could be empty when registering an instance. During the process of registering an instance via Grpc, we are directly using the value of the service's name field.

By the way, whether it's necessary to maintain the logic of custom instance IDs is also a topic worth discussing.

KomachiSion (changes requested):
You can see com.alibaba.nacos.naming.pojo.instance.HttpRequestInstanceBuilder#setAttributesToBuilder

the instance service name in from the request. 
so grpc also should use service name in request.

service name in request should not be null acutally.

Bo-Qiu:
> 

Thanks for your information, I've changed it.

KomachiSion (approved):

naming/src/main/java/com/alibaba/nacos/naming/remote/rpc/handler/InstanceRequestHandler.java
```diff
@@ -51,6 +52,7 @@ public InstanceRequestHandler(EphemeralClientOperationServiceImpl clientOperatio
     public InstanceResponse handle(InstanceRequest request, RequestMeta meta) throws NacosException {
         Service service = Service
                 .newService(request.getNamespace(), request.getGroupName(), request.getServiceName(), true);
+        InstanceUtil.setInstanceIdIfEmpty(request.getInstance(), request.getServiceName());
```
KomachiSion, line 55: Is this will lose the group name?

KomachiSion (changes requested):

naming/src/main/java/com/alibaba/nacos/naming/remote/rpc/handler/InstanceRequestHandler.java
Bo-Qiu: > 

Thank you for your reminder, I have already fixed this issue, please review it again.

KomachiSion (approved):

Issue #10831, opened 2023-07-20
When using the deregisterInstance method to remove one of multiple instances registered by batchRegisterInstance, all instances registered by batchRegisterInstance will be removed.


Bo-Qiu:
**Describe the bug**
When using the deregisterInstance method to remove one of multiple instances registered by batchRegisterInstance, all instances registered by batchRegisterInstance will be removed.

**Expected behavior**
When you use the deregisterInstance method to deregister a specific instance from multiple instances registered through batchRegisterInstance, this will only remove that specific instance, and all other instances will remain.

**Actually behavior**
When using the deregisterInstance method to remove one of multiple instances registered by batchRegisterInstance, all instances registered by batchRegisterInstance will be removed.

**How to Reproduce**
1. batchRegisterInstance("demo1", "group", instances);
2. naming.getAllInstances("demo1", "group", false);
3. deregisterInstance("demo1", "group", instance1);
4. naming.getAllInstances("demo1", "group", false);
5. you will find that the instance list is empty.

**Desktop (please complete the following information):**
 - OS: [MAC M1]
 - Version [nacos-server 2.2.4, nacos-client 2.2.4]
 - Module [naming]
 - SDK [original]

Bo-Qiu:
I will create a PR to fix this bug as quickly as possible.

PR #10836, opened 2023-07-21
fix(#10831): When using the deregisterInstance method to remove one o…

Bo-Qiu:

fix(#10831  ) When using the deregisterInstance method to remove one of multiple instances registered by batchRegisterInstance, all instances registered by batchRegisterInstance will be removed. 

## Brief changelog

fix(#10831  ) When using the deregisterInstance method to remove one of multiple instances registered by batchRegisterInstance, all instances registered by batchRegisterInstance will be removed. 

## Verifying this change

fix(#10831  ) When using the deregisterInstance method to remove one of multiple instances registered by batchRegisterInstance, all instances registered by batchRegisterInstance will be removed.

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/naming/remote/gprc/NamingGrpcClientProxy.java
+++ b/client/src/main/java/com/alibaba/nacos/client/naming/remote/gprc/NamingGrpcClientProxy.java
@@ -54,8 +54,8 @@ import com.alibaba.nacos.common.notify.NotifyCenter;
 import com.alibaba.nacos.common.remote.ConnectionType;
 import com.alibaba.nacos.common.remote.client.RpcClient;
 import com.alibaba.nacos.common.remote.client.RpcClientFactory;
-import com.alibaba.nacos.common.remote.client.ServerListFactory;
 import com.alibaba.nacos.common.remote.client.RpcClientTlsConfig;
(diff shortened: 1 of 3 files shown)


KomachiSion (changes requested):
Is it thread safe?

Bo-Qiu:
To ensure the thread safety of this method, we just need to maintain the thread safety of batchDeregisterService. Is the thread safety issue around batchDeregisterService fixed within this PR, or should we propose a new PR for it?

Bo-Qiu:
@KomachiSion  
By the way, I've noticed that there are many places where unnecessary 'synchronized' locks have been added. Since registeredInstances and subscribes are instances of ConcurrentHashMap, which is inherently thread-safe, there's no need to add extra locks for its atomic operations. Do you think it's necessary to open an issue to optimize this? 
e.g.
```
synchronized (subscribes) {
            subscribes.put(key, redoData);
        }
```
synchronized should be removed.

```
synchronized (subscribes) {
            SubscriberRedoData redoData = subscribes.get(key);
            if (null != redoData) {
                redoData.setRegistered(true);
            }
        }
```
It can be replaced by the computeIfPresent method

Bo-Qiu:
I have directly placed the implementation of batchDeregisterService within a synchronized block to ensure thread safety. Should I use a better way to ensure thread safety?

KomachiSion (approved):

PR #10882, opened 2023-07-29
refactor: isChangedServiceInfo method in ServiceInfoHolder has been r…

E1se2t:
The isChangedServiceInfo method primarily aims to compare instance information for any changes and then categorize them into three separate collections: new, deleted, and modified collections.

In the original code, adding and modifying instances were processed within one loop, while deleting instances were handled in another loop, resulting in unnecessary iterations. In the refactored code, I combined the keys of the newHostMap and oldHostMap collections and used a HashSet to remove duplicates. Then, within the loop of the allHostMapKeys collection, the following conditions were checked sequentially:

1. newInstance ！= oldInstance 
add modHosts

2. newInstance ！= null && oldInstance == null
add newHosts

3. newInstance == null && oldInstance ！= null
add remvHosts

Add them to the ArrayList, and there is no longer a need for HashSet to handle duplicate results，The improvement has enhanced the efficiency and code readability of the method, while also eliminating redundant processing.

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/naming/cache/ServiceInfoHolder.java
+++ b/client/src/main/java/com/alibaba/nacos/client/naming/cache/ServiceInfoHolder.java
@@ -32,10 +32,8 @@ import com.alibaba.nacos.common.utils.JacksonUtils;
 import com.alibaba.nacos.common.utils.StringUtils;
 
 import java.io.File;
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
@@ -178,7 +176,7 @@ public class ServiceInfoHolder implements Closeable {
     private boolean isEmptyOrErrorPush(ServiceInfo serviceInfo) {
         return null == serviceInfo.getHosts() || (pushEmptyProtection && !serviceInfo.validate());
     }
-    
+
     private boolean isChangedServiceInfo(ServiceInfo oldService, ServiceInfo newService) {
         if (null == oldService) {
             NAMING_LOGGER.info("init new ips({}) service: {} -> {}", newService.ipCount(), newService.getKey(),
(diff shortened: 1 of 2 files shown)


KomachiSion (changes requested):
I think the change will lead to increased time complexity.

The old implementation: loop newHosts and oldHosts twice for each.

But new implementation loop newHosts and oldHosts third times(host to map, map key set merge, and merged key set loop).

E1se2t:
image
如果我在将host转换为map的循环期间合并map的键，就可以避免您提到的三个循环的问题。是否支持此更改?

E1se2t:
> I think the change will lead to increased time complexity.
> 
> The old implementation: loop newHosts and oldHosts twice for each.
> 
> But new implementation loop newHosts and oldHosts third times(host to map, map key set merge, and merged key set loop).

image
If I merge the keys of the map during the loop that converts hosts to a map, it can avoid the issue of three loops that you mentioned. Is this change supported?

If this change is not allowed or considered unnecessary, I will close this PR, and the method will remain as it is. Thank you for your response.

KomachiSion:
Why not do enhance directly in this issue?

The situation prefer performance I think. And the old one I think the readability has no much different for new implementaton.

E1se2t:
> Why not do enhance directly in this issue?
> 
> The situation prefer performance I think. And the old one I think the readability has no much different for new implementaton.



> Why not do enhance directly in this issue?
> 
> The situation prefer performance I think. And the old one I think the readability has no much different for new implementaton.

I have reworked the code again, this time removing the step where I converted newService.getHosts() into a map. Within the loop iterating over newService.getHosts(), I've completed the comparison tasks for Instance objects:

1.
if oldHost ！= null && oldHost == newHost
oldHostMap.remove(host);

2.
if oldHost ！= null && newHost != oldHost
oldHostMap.remove(key);
and
modHosts.add(newHost);

3.
if oldHost == null
newHosts.add(newHost);

4.
After filtering oldHostMap within the loop, at the end of the iteration, only data where newHost == null and oldHosts != null remain in oldHostMap. These data are directly placed in the remvHosts.


Through this approach, there is no need to pre-convert newHosts into a map. Furthermore, it also reduces the requirement for two separate comparison and filtering loops.

Issue #10947, opened 2023-08-10
The logic of the com.alibaba.nacos.auth.config.AuthConfigs#validate method can be optimized.


Bo-Qiu:
**Describe the bug**

**Expected behavior**
When ‘nacos.core.auth.enable.userAgentAuthWhite’ is true, there is no need to verify ‘nacos.core.auth.plugin.nacos.token.secret.key’ and ‘nacos.core.auth.server.identity.value’.

**Actually behavior**
When ‘nacos.core.auth.enable.userAgentAuthWhite’ is true, the verification is still executed.

KomachiSion:
No need.

The enabledUserAgent is only for upgrade from old version and change the identity info.

Community force users to config the identity info when open auth.

PR #10948, opened 2023-08-10
fix(#10947): The logic of the com.alibaba.nacos.auth.config.AuthConfigs#validate method can be optimized.

Bo-Qiu:

fix(#10947 ): The logic of the com.alibaba.nacos.auth.config.AuthConfigs#validate method can be optimized.

## Brief changelog

fix(#10947 ): The logic of the com.alibaba.nacos.auth.config.AuthConfigs#validate method can be optimized.

## Verifying this change

fix(#10947 ): The logic of the com.alibaba.nacos.auth.config.AuthConfigs#validate method can be optimized.


KomachiSion (changes requested):
No need.

The enabledUserAgent is only for upgrade from old version and change the identity info.

Community **force** users to config the identity info when open auth.

Issue #11127, opened 2023-09-13
[Nacos Client 2.2.3] ClientWork removeListener 和 addListener 存在并发问题


MajorHe1:
先执行removeLister，再执行addListener，会出现如下并发问题：
![企业微信截图_16945907173768](https://github.com/alibaba/nacos/assets/53824889/a99b4692-e991-413c-971f-2e8865d78162)
![企业微信截图_16945909157303](https://github.com/alibaba/nacos/assets/53824889/dc408e76-8e9a-4129-a3fd-8300041d19ef)
后果就是用户后面执行的addListener方法，并不会真正发起对服务端的监听请求，也就是没有监听上

大家讨论一下修改方案

Bo-Qiu:
这么写应该就没问题了
![image](https://github.com/alibaba/nacos/assets/46370663/260bc8c8-0f1a-42c9-81e1-8b96cf04fe3b)

Bo-Qiu:
I created a PR, please review it, thx!

shiyiyue1102:
#11681 has been merged and includes the ClientWorker concurrency fix with tests. Could you confirm whether the removeListener/addListener case is resolved on your side?

If there is no further update in the next 3 days, we will close this issue as resolved.

---

#11681 已合并，其中包含 ClientWorker 并发修复和对应测试。麻烦确认一下 removeListener/addListener 这个场景在你这边是否已经解决？

如果接下来 3 天内没有进一步反馈，我们会按已解决来关闭这个 issue。

PR #11131, opened 2023-09-14
fix(11127): Fix ClientWork removeListener and addListener concurrency issue.

Bo-Qiu:
fix(#11127 ): Fix ClientWork removeListener and addListener concurrency issue.


Bo-Qiu:
@shiyiyue1102 老哥，空了帮忙review下代码哇，感谢

PR #11195, opened 2023-09-27
Optimize MapperManager

kang-hl:

XXXXX

## Brief changelog

XX

## Verifying this change

XXXX

### DIFF
--- a/plugin/datasource/src/main/java/com/alibaba/nacos/plugin/datasource/MapperManager.java
+++ b/plugin/datasource/src/main/java/com/alibaba/nacos/plugin/datasource/MapperManager.java
@@ -64,11 +64,10 @@ public class MapperManager {
     /**
      * The init method.
      */
-    public void loadInitial() {
+    public synchronized void loadInitial() {
         Collection<Mapper> mappers = NacosServiceLoader.load(Mapper.class);
         for (Mapper mapper : mappers) {
-            Map<String, Mapper> mapperMap = MAPPER_SPI_MAP.computeIfAbsent(mapper.getDataSource(), (r) -> new HashMap<>(16));
-            mapperMap.put(mapper.getTableName(), mapper);
+            putMapper(mapper);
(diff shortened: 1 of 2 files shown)


KomachiSion (changes requested):
Is necessary to change ConcurrentHashMap in MapperManager?  I think there is no need, the MAPPER_SPI_MAP should only put in init step which is single thread. And even plugin implementation use join, the times is not many and frequent. So I think HashMap is enough and join with synchronized.

And other think is indent problem, please use nacos code style to reformat codes.

kang-hl:
> Is necessary to change ConcurrentHashMap in MapperManager? I think there is no need, the MAPPER_SPI_MAP should only put in init step which is single thread. And even plugin implementation use join, the times is not many and frequent. So I think HashMap is enough and join with synchronized.
> 
> And other think is indent problem, please use nacos code style to reformat codes.

已修改

plugin/datasource/src/main/java/com/alibaba/nacos/plugin/datasource/MapperManager.java
```diff
@@ -64,11 +64,10 @@ public static MapperManager instance(boolean isDataSourceLogEnable) {
     /**
      * The init method.
      */
-    public void loadInitial() {
+    public synchronized void loadInitial() {
```
KomachiSion, line 67: 这方法有必要锁吗？ 只有初始化的时候单线程执行一次

KomachiSion (changes requested):

plugin/datasource/src/main/java/com/alibaba/nacos/plugin/datasource/MapperManager.java
kang-hl: > 这方法有必要锁吗？ 只有初始化的时候单线程执行一次

那就需要修改为private

KomachiSion (changes requested):
能帮忙加下UT吗

KomachiSion (approved):

Issue #11214, opened 2023-10-04
Optimization method com.alibaba.nacos.naming.core.v2.client.manager.ClientManager#syncClientConnected


Bo-Qiu:
**Is your feature request related to a problem? Please describe.**
Optimization method com.alibaba.nacos.naming.core.v2.client.manager.ClientManager#syncClientConnected, the client instance is no longer pre-created. A client instance is created only if it does not exist in the Clients collection.

KomachiSion:
拒绝， 详情见PR

PR #11219, opened 2023-10-05
feat(#11214): Optimization method com.alibaba.nacos.naming.core.v2.client.manager.impl.ConnectionBasedClientManager.clientConnected

Bo-Qiu:

feat(#11214): 
Optimization method com.alibaba.nacos.naming.core.v2.client.manager.impl.ConnectionBasedClientManager.clientConnected.
Before creating a client instance, first determine whether a client instance with the same clientId already exists in clientsMap. If it exists, no invalid client instance will be created.


KomachiSion (changes requested):
拒绝，多态方法用于方便扩展和复用使用设计，不给予删除

Issue #11382, opened 2023-11-16
Listener api doesn't support paging in Listening Query page 


0808bing:
**Describe the bug**
Listener api doesn't support paging in Listening Query page, but appear the pagination feature
![bug1](https://github.com/alibaba/nacos/assets/9526863/887b2545-3e8a-45a7-8df4-f33f0ed6519d)

**Expected behavior**
if listener api doesn't support paging in Listening Query page, please delete the pagination feature 

**Actually behavior**
The Listening Query page have pagination feature, then change the page number, the listen api still return all data.

**How to Reproduce**
1. Go to 'ConfigManagement -> Listening Query'
2. Click on 'page number'

PR #11383, opened 2023-11-16
[ISSUE#11382]Delete Listening Query page pagination, because listener api doesn't support paging

0808bing:
For issue #11382 

Listener api doesn't support paging in Listening Query page, so delete the pagination feature


985492783 (changes requested):
我建议是在前端做个分页，而不是把分页组件删了

0808bing:
> 我建议是在前端做个分页，而不是把分页组件删了

这个也可以

Issue #11502, opened 2023-12-13
Unexpected derby.log file generated in current user directory


yoyo837:
**Describe the bug**
In macOS,  run nacos with `/User/xxx/nacos/bin/start.sh`, unexpected `derby.log` file generated in current user directory.

**Expected behavior**
No unexpected log file generated.

**Actually behavior**
Unexpected derby.log file generated in current user directory.

**How to Reproduce**
Just run start nacos.

**Desktop (please complete the following information):**
 - OS: [macOS latest]
 - Version [2.2.4 with #10638 fixed source code]
 - Module [e.g. naming/config]
 - SDK [Oracle jdk 1.8.0_391 aarch64]

**Additional context**

yoyo837:
Ref: https://github.com/alibaba/nacos/issues/10593

DarrenLuo:
@i will solve it@

KomachiSion:
The latest version 2.3.0 may fix this version, can someone test for the 2.3.0?

KomachiSion:
I see, the derby.log is a logs when you use derby database.

so the derby.log will exist in work dir. It's match expectations.

yoyo837:
> I see, the derby.log is a logs when you use derby database.
> 
> so the derby.log will exist in work dir. It's match expectations.

On macOS, are the `$HOME` directory and the `workdir` directory you mentioned the same thing?

It appears under `$HOME`, is this expected? I think it should appear in a folder with a certain name, not appear directly.

KomachiSion:
I work on macOS, and the derby.log appear on the `nacos.work.dir`, not int `$HOME`.

KomachiSion:
It might be some config diff in your macOS system, so that the derby read the config and print the log to `$HOME`.

yoyo837:
> It might be some config diff in your macOS system, so that the derby read the config and print the log to `$HOME`.

Thanks for reply. Where should I check to find this difference config. I would be grateful if you could provide some guidance.

KomachiSion:
I have no idea for this problem. Because I don't find this problem, maybe you can check your nacos dir is whether same as `$HOME` or some env properties special.

And you can move to derby community to find the derby.log information.

KomachiSion:
No more response from author, and community can't reproduce this problem.

Issue #11497, opened 2023-12-12
2.3.0开启数据源sql日志启动报错的bug


309283175:
# use this config in application.properties
```
nacos.plugin.datasource.log.enabled=true
```
then start server  
Application run failed with this Exp:
Caused by: java.lang.ClassCastException: com.sun.proxy.$Proxy110 cannot be cast to com.alibaba.nacos.plugin.datasource.proxy.MapperProxy
```
public class MapperProxy implements InvocationHandler {
    //private static final Map<String, MapperProxy> SINGLE_MAPPER_PROXY_MAP = new ConcurrentHashMap<>(16);
    private static final Map<String, Mapper> SINGLE_MAPPER_PROXY_MAP = new ConcurrentHashMap<>(16);
}
```
change MapperProxy to  Mapper can solve this problem.

Issue #11461, opened 2023-12-02
Proxy110 cannot be cast to com.alibaba.nacos.plugin.datasource.proxy.MapperProxy


1148260649:
![image](https://github.com/alibaba/nacos/assets/45970803/51d91c0b-4062-4209-b79c-c55749cdba31)
启动失败，提示 mapperproxy 该如何修改
Caused by: com.alibaba.nacos.api.exception.NacosException: Nacos Server did not start because dumpservice bean construction failure :
com.sun.proxy.$Proxy110 cannot be cast to com.alibaba.nacos.plugin.datasource.proxy.MapperProxy

1148260649:
> What's the `com.sun.proxy.$Proxy110`? Can you template to describe how to happen?
I didn't do anything, but I encountered this error when starting the Nacos source code in standalone mode.

KomachiSion:
console.nacos.io is staring with standalone mode, no happen this problem, Please provide how to reproduce this problem.

rongxingsun:
> console.nacos.io is staring with standalone mode, no happen this problem, Please provide how to reproduce this problem.
jdk是17版本嘛

309283175:
#11497 
```
MapperProxy
Map<String, Mapper> SINGLE_MAPPER_PROXY_MAP
```

PR #11507, opened 2023-12-13
[ISSUE #11502] fix proxy mapper bug

985492783:
description: when set nacos.plugin.datasource.log.enabled=true, will create Proxy Object, but it was mistakenly changed to a different type

relate #11497 #11461


onewe:
LGTM

KomachiSion (approved):

PR #11658, opened 2024-01-15
optimize dump all logic and optimize dump change interface

shiyiyue1102:
1.optimize config dump all,do not select content in page query,single query config on md5 and ts updated.
2. remove convertDeleteConfig and convertConfigConfig ，use standard RowMapper instead.


shiyiyue1102:
https://github.com/alibaba/nacos/issues/11632

KomachiSion (changes requested):

KomachiSion (approved):

PR #11788, opened 2024-03-01
fix ClassCastException

xiaoshuai:
Controller returns Boolean type, ConfigChangeAspect returns RestResult type, trigger exception.

java.lang.ClassCastException: class com.alibaba.nacos.common.model.RestResult cannot be cast to class java.lang.Boolean (com.alibaba.nacos. common.model.RestResult is in unnamed module of loader org.springframework.boot.loader.LaunchedURLClassLoader @39c0f4a; java.lang.Boolean is in module java.base of loader "bootstrap')

## Brief changelog

Before: Controller returns Boolean type, ConfigChangeAspect returns RestResult type, trigger exception.
After: Both return RestResult type.

## Verifying this change

XXXX


KomachiSion (changes requested):
Can't change it, the API return change will cause some old version client can;t handle it .

you can see the ConfigControllerV2, and see whether should change in there.

KomachiSion:
No more response from author, current fix way is not reasonable.

Issue #11951, opened 2024-04-11
删除空服务时，发布DeregisterServiceTraceEvent 传参错误


Happy-26:
在 `ServiceNameFilter#doFilter()`方法中，使用`groupName@@serviceName`作为一个新的服务名，在`ServiceController#remove()`方法中，发布`DeregisterServiceTraceEvent`事件时，未对组装后的serviceName进行处理，源代码如下：

    @DeleteMapping
    @TpsControl(pointName = "NamingServiceDeregister", name = "HttpNamingServiceDeregister")
    @Secured(action = ActionTypes.WRITE)
    public String remove(@RequestParam(defaultValue = Constants.DEFAULT_NAMESPACE_ID) String namespaceId,
            @RequestParam String serviceName) throws Exception {
        getServiceOperator().delete(namespaceId, serviceName);
        NotifyCenter.publishEvent(
                new DeregisterServiceTraceEvent(System.currentTimeMillis(), namespaceId, "", serviceName));
        return "ok";
    }

修改后的代码如下：

    @DeleteMapping
    @TpsControl(pointName = "NamingServiceDeregister", name = "HttpNamingServiceDeregister")
    @Secured(action = ActionTypes.WRITE)
    public String remove(@RequestParam(defaultValue = Constants.DEFAULT_NAMESPACE_ID) String namespaceId,
            @RequestParam String serviceName) throws Exception {
        
        getServiceOperator().delete(namespaceId, serviceName);
        NotifyCenter.publishEvent(
                new DeregisterServiceTraceEvent(System.currentTimeMillis(), namespaceId,
                        NamingUtils.getGroupName(serviceName), NamingUtils.getServiceName(serviceName)));
        return "ok";
    }

KomachiSion:
有工具类：NamingUtils 来获取groupName和serviceName

PR #11955, opened 2024-04-11
[ISSUE #11951] Fix the problem that the servicename and groupname are not resolved correctly when deleting an empty service instance

Happy-26:
Bug  fix [#11951 ](https://github.com/alibaba/nacos/issues/11951) 修改删除空服务实例时，服务名和分组名没有正确解析的问题


KomachiSion (changes requested):
Can you help community to add unit test for this case?

Happy-26:
It seems that a unit test already exists here[ServiceControllerTest](https://github.com/alibaba/nacos/blob/develop/naming/src/main/java/com/alibaba/nacos/naming/controllers/ServiceController.java)

KomachiSion:
> It seems that a unit test already exists here：[ServiceControllerTest](https://github.com/alibaba/nacos/blob/develop/naming/src/test/java/com/alibaba/nacos/naming/controllers/ServiceControllerTest.java)

Yes I know it, I mean the old one only check the UpdateServiceTraceEvent , not check DeregisterServiceTraceEvent

KomachiSion:
> > It seems that a unit test already exists here：[ServiceControllerTest](https://github.com/alibaba/nacos/blob/develop/naming/src/test/java/com/alibaba/nacos/naming/controllers/ServiceControllerTest.java)
> 
> Yes I know it, I mean the old one only check the UpdateServiceTraceEvent , not check DeregisterServiceTraceEvent

Prefer you add the check for DeregisterServiceTraceEvent.

Happy-26:
> > > It seems that a unit test already exists here：[ServiceControllerTest](https://github.com/alibaba/nacos/blob/develop/naming/src/test/java/com/alibaba/nacos/naming/controllers/ServiceControllerTest.java)
> > 
> > 
> > Yes I know it, I mean the old one only check the UpdateServiceTraceEvent , not check DeregisterServiceTraceEvent
> 
> Prefer you add the check for DeregisterServiceTraceEvent in UT.

I'm sorry I didn't understand how to write unit tests for Nacos. Do you mean to change it to this?

    @Before
    public void before() {
        super.before();
        subscriber = new SmartSubscriber() {
            @Override
            public List<Class<? extends Event>> subscribeTypes() {
                List<Class<? extends Event>> result = new LinkedList<>();
                result.add(UpdateServiceTraceEvent.class);
                result.add(DeregisterServiceTraceEvent.class);
                return result;
            }
            
            @Override
            public void onEvent(Event event) {
                eventReceivedClass = event.getClass();
            }
        };
        NotifyCenter.registerSubscriber(subscriber);
    }
    
    @After
    public void tearDown() throws Exception {
        NotifyCenter.deregisterSubscriber(subscriber);
        NotifyCenter.deregisterPublisher(UpdateServiceTraceEvent.class);
        NotifyCenter.deregisterPublisher(DeregisterServiceTraceEvent.class);
        eventReceivedClass = null;
    }

KomachiSion (approved):

Issue #12231, opened 2024-06-15
There is an inconsistency between Nacos's local configuration information and the database configuration information, and this inconsistency cannot be automatically resolved.


Bo-Qiu:
**Describe the bug**
When inserting and updating configurations in the database, the gmt_modified field is set using the local timestamp of the Nacos node rather than the database's built-in time function. This can lead to inconsistencies between the server node data and the database data when different Nacos nodes concurrently modify the same configuration, and this inconsistency cannot be automatically resolved.

**Expected behavior**
When inserting and updating configurations in the database, the time-related field values need to be set using the time obtained from the database's built-in time function.

**Actually behavior**
When inserting and updating configurations in the database, the gmt_modified field is set using the local timestamp of the Nacos node rather than the database's built-in time function.

caoyanan666:
这个会让pg报错，
![image](https://github.com/user-attachments/assets/38212261-b940-499e-93d3-dbef3126fb0e)

caoyanan666:
https://github.com/nacos-group/nacos-plugin/pull/46/files

PR #12232, opened 2024-06-15
fix(#12231): When inserting and updating configurations in the database, the time-related field values need to be set using the time obtained from the database's built-in time function.

Bo-Qiu:

XXXXX

fix(#12231 ): When inserting and updating configurations in the database, the time-related field values need to be set using the time obtained from the database's built-in time function.

XX

fix(#12231 ): When inserting and updating configurations in the database, the time-related field values need to be set using the time obtained from the database's built-in time function.

XXXX

fix(#12231 ): When inserting and updating configurations in the database, the time-related field values need to be set using the time obtained from the database's built-in time function.


Bo-Qiu:
@shiyiyue1102 
Hi, please review it when you have time, thx!

Bo-Qiu:
@shiyiyue1102 
Hi, please review it when you have time, thx!

PR #12491, opened 2024-08-15
[ISSUE #12483] Configuration list adds configuration format

tonycody:
issue: #12483


common/src/main/java/com/alibaba/nacos/common/constant/Symbols.java
```diff
@@ -0,0 +1,31 @@
+/*
```
Nftcollecte, line 1: Gm

Nftcollecte (changes requested):

common/src/main/java/com/alibaba/nacos/common/constant/Symbols.java
tonycody: Not sure what it means? Please explain.

KomachiSion (approved):

Issue #12569, opened 2024-08-30
使用Oracle数据库运行，启动报错。


bes2008:
**Describe the bug**
使用Oracle数据库运行，启动报错。
报错内容：
2024-08-30 17:15:12,901 WARN HikariPool-1 - Connection oracle.jdbc.driver.T4CConnection@4653747d marked as broken because of SQLSTATE(0A000), ErrorCode(304                        7)

java.sql.SQLException: ORA-03047: number '1' is not syntactically valid following '...1 FROM config_info_tag LIMIT '

https://docs.oracle.com/error-help/db/ora-03047/
        at oracle.jdbc.driver.T4CTTIoer11.processError(T4CTTIoer11.java:709)
        at oracle.jdbc.driver.T4CTTIoer11.processError(T4CTTIoer11.java:609)
        at oracle.jdbc.driver.T4C8Oall.processError(T4C8Oall.java:1290)
        at oracle.jdbc.driver.T4CTTIfun.receive(T4CTTIfun.java:1100)
        at oracle.jdbc.driver.T4CTTIfun.doRPC(T4CTTIfun.java:408)
        at oracle.jdbc.driver.T4C8Oall.doOALL(T4C8Oall.java:483)
        at oracle.jdbc.driver.T4CStatement.doOall8(T4CStatement.java:170)
        at oracle.jdbc.driver.T4CStatement.executeForDescribe(T4CStatement.java:1029)
        at oracle.jdbc.driver.OracleStatement.prepareDefineBufferAndExecute(OracleStatement.java:1299)
        at oracle.jdbc.driver.OracleStatement.executeMaybeDescribe(OracleStatement.java:1224)
        at oracle.jdbc.driver.OracleStatement.executeSQLSelect(OracleStatement.java:1592)
        at oracle.jdbc.driver.OracleStatement.doExecuteWithTimeout(OracleStatement.java:1442)
        at oracle.jdbc.driver.OracleStatement.executeQuery(OracleStatement.java:1858)
        at oracle.jdbc.driver.OracleStatementWrapper.executeQuery(OracleStatementWrapper.java:382)
        at com.zaxxer.hikari.pool.ProxyStatement.executeQuery(ProxyStatement.java:111)
        at com.zaxxer.hikari.pool.HikariProxyStatement.executeQuery(HikariProxyStatement.java)
        at org.springframework.jdbc.core.JdbcTemplate$1QueryStatementCallback.doInStatement(JdbcTemplate.java:453)
[...]

PR #12571, opened 2024-08-31
fix Issue 12569

huangkemingyyds:
https://github.com/alibaba/nacos/issues/12569

Change the checklist existence check to use ‘count’.


KomachiSion:
The PR is invalid beacuse include others commit

PR #12929, opened 2024-12-06
[ISSUE#12644] Return cached instances when getAllInstance with subscribe exception

### LINKED ISSUE #12644: Naming local cache may be ignored in a rare scenario
**Describe the bug**

Naming local cache may be ignored in a rare scenario. So even when local cache is not empty, user's invocation would still get exception.

**Expected behavior**

If local cache is not empty, Nacos client should never throw exception.

**Actually behavior**

In a rare case, when local cache is not empty, user's invocation would get exception.

**How to Reproduce**

It's hard to reproduce but it did happen in our production environment. Here is the related logic:

in the method at [getServiceInfoBySubscribe](https://github.com/alibaba/nacos/blob/3a9003bc7cbb642ef836975bde65e596e87eb697/client/src/main/java/com/alibaba/nacos/client/naming/NacosNamingService.java#L336):
1. first it tried to get service info from local cache;
2. then it will check if local cache is null or client is not subscribed:
image
3. if yes, it will try to subscribe from remote Nacos server.

Usually it will work, because whenever `clientProxy.isSubscribed(...)` returns false, it means the [Nacos client](https://github.com/alibaba/nacos/blob/3a9003bc7cbb642ef836975bde65e596e87eb697/common/src/main/java/com/alibaba/nacos/common/remote/client/RpcClient.java#L516) has just reconnected to Nacos server and closed the old connection:
image

As new connection is ready, so the subscribe request would succeed.

But if the new connection is down immediately again so the subscribe request failed, then in the method `getServiceInfoBySubscribe`, an exception will be thrown.

 - OS: [e.g. Centos]
 - Version: nacos-server 2.1.2, nacos-client 2.1.2
 - Module: naming
 - SDK: spring-cloud-alibaba-nacos 2021.1

**Additional context**

I suggest to add a protection logic in the method `getServiceInfoBySubscribe`, so that whenever the local cache is not empty, the remote request error or any other exception will be ignored.

KomachiSion:

Fixes #12644 .

## Brief changelog

1. Return cached instances when getAllInstance with subscribe exception.
2. Throw subscribe exception if no cached instances .
3. getAllInstance and selectInstance with clusters new local selector but not subscribe again.

## Verifying this change

XXXX

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/naming/NacosNamingService.java
+++ b/client/src/main/java/com/alibaba/nacos/client/naming/NacosNamingService.java
@@ -24,6 +24,8 @@ import com.alibaba.nacos.api.naming.listener.EventListener;
 import com.alibaba.nacos.api.naming.pojo.Instance;
 import com.alibaba.nacos.api.naming.pojo.ListView;
 import com.alibaba.nacos.api.naming.pojo.ServiceInfo;
+import com.alibaba.nacos.api.naming.selector.NamingContext;
+import com.alibaba.nacos.api.naming.selector.NamingResult;
 import com.alibaba.nacos.api.naming.selector.NamingSelector;
 import com.alibaba.nacos.api.naming.utils.NamingUtils;
 import com.alibaba.nacos.api.selector.AbstractSelector;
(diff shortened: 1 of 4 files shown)


wuyfee:
$\color{green}{SUCCESS}$
[DETAILS](https://github.com/alibaba/nacos/actions/runs/12195253858)
 ✅ - docker:   success 
 ✅ - deploy (standalone & cluster & standalone_auth):   success 
 ✅ - e2e-java-test (standalone & cluster & standalone_auth):   success 
 ✅ - e2e-go-test (standalone & cluster):   success 
 ✅ - e2e-cpp-test (standalone & cluster):   success 
 ✅ - e2e-csharp-test (standalone & cluster):   success 
 ✅ - e2e-nodejs-test (standalone & cluster):   success 
 ✅ - e2e-python-test (standalone & cluster):   success 

PR #13717, opened 2025-08-17
Using the JDBC4 Connection.isValid() method to test connection

kuchikij:

For adapting to different databases

## Brief changelog

Using the JDBC4 Connection.isValid() method to test connection validity can be more efficient on some databases.

## Verifying this change

XXXX

### DIFF
--- a/persistence/src/main/java/com/alibaba/nacos/persistence/datasource/ExternalDataSourceProperties.java
+++ b/persistence/src/main/java/com/alibaba/nacos/persistence/datasource/ExternalDataSourceProperties.java
@@ -89,10 +89,7 @@ public class ExternalDataSourceProperties {
             poolProperties.setUsername(getOrDefault(user, index, user.get(0)).trim());
             poolProperties.setPassword(getOrDefault(password, index, password.get(0)).trim());
             HikariDataSource ds = poolProperties.getDataSource();
-            if (StringUtils.isEmpty(ds.getConnectionTestQuery())) {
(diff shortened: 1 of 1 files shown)


kuchikij:
> When to call `isValid`?

Sorry, I didn't describe it clearly.
Currently, each database implements the **Connection.isvalid** method, which conducts Connection checks based on the characteristics of the database itself.
In creating a connection, will call **PoolBase.CheckValidationSupport** method, this method will determine if there is no set ConnectionTestQuery, use the **Connection.isvalid** method, That is, to use the implementation of each database itself. 
It has better compatibility when adapting to different databases.

KomachiSion:
> > When to call `isValid`?
> 
> Sorry, I didn't describe it clearly. Currently, each database implements the **Connection.isvalid** method, which conducts Connection checks based on the characteristics of the database itself. In creating a connection, will call **PoolBase.CheckValidationSupport** method, this method will determine if there is no set ConnectionTestQuery, use the **Connection.isvalid** method, That is, to use the implementation of each database itself. It has better compatibility when adapting to different databases.

Do you make sure all database jdbc already implement this method and can switch or reconnect after isValid failed?

I don't think we should remove this test sql, the test sql is used by hikariCP to check the connections in pool is enabled.

If test failed, pool will close the one and generate a new one into pool.

If hikariCP replaced to use isValid to check rather than using test sql. We can follow. But I see the hikariCP not mark deprecated in test SQL feature, so I think we should not remove it.

PR #13718, opened 2025-08-17
Fix the issue where config info id cannot return in the oceanbase oracle

kuchikij:

For adapting to different databases

## Brief changelog

To solve the problem that using KeyHolder under oceanbase oracle cannot return increment value.

## Verifying this change

1. Prepare OceanBase For Oracle(3.2.x) environment,
2. When creating and saving a configuration in the configuration management menu, an error prompt appears. 


KomachiSion:
The keyholder way is spring.jdbc provided to enhance the changed id information.

I think if change to select and query, will change the meaning of the old logic.

It should be discuss in issue and found out the actual reason why oceanbase and oracle not support keyholder way but not fix it in nacos.

KomachiSion:
No more response from author, and it should be discuss first in issue.

If stiil want to do this change, please submit issue to discuss first. Thanks.

PR #13961, opened 2025-11-24
Fix invalid regex pattern in ValidatorUtils

RamsaiPolisetti:
This PR fixes a malformed regex pattern in ValidatorUtils.

The previous regex was incomplete due to missing quantifier and closing token,
causing incorrect context-path validation.

Updated:
- Corrected CONTEXT_PATH_MATCH pattern
- Verified successful compilation with: ./mvnw -DskipTests=true install

This resolves incorrect path validation and restores proper client behavior.


KomachiSion (changes requested):
I can't get the reason of this PR.

PR #14717, opened 2026-03-25
fix(test): fix flaky MemoryMcpCacheIndexTest by eliminating scheduler races

### LINKED ISSUE #14550: [Bug] MemoryMcpCacheIndexTest flaky: cleanup scheduler races with assertions
### Background

`MemoryMcpCacheIndexTest.testCleanupExpiredEntriesDoesNotAffectValidEntries` intermittently fails in CI with NPE on `this.properties`. Observed in PR #14470 CI run.

### Root Cause Analysis

Three factors combine to cause the flaky behavior:

**1. Missing `@AfterEach` cleanup — scheduler thread leak**

Each `@BeforeEach` creates a new `MemoryMcpCacheIndex` instance with a background cleanup scheduler, but old instances are never shut down. 30 test methods = 29 zombie schedulers running concurrently.

**2. Background scheduler races with assertions (primary cause)**

`testCleanupExpiredEntriesDoesNotAffectValidEntries` uses `cleanupIntervalSeconds=1` and `expireTimeSeconds=2`. Under CI load:

```
Ideal timeline:
T=0.0s  insert id1
T=1.1s  insert id2
T=2.2s  id1 expired (age=2.2s >= 2s), id2 safe (age=1.1s < 2s)
→ assertion passes ✓

CI load timeline:
T=0.0s  insert id1
T=1.4s  insert id2 (sleep overshot by 300ms)
         ↑ integer-second truncation: 1400ms/1000 = 1, same second as id2!
T=3.0s  ⚡ scheduler fires: id2 age >= 2s → also cleaned up!
T=3.1s  assert id2 exists → ❌ FAIL
```

Triple jeopardy: `Thread.sleep` imprecision + `System.currentTimeMillis() / 1000` integer truncation (~1s error) + async scheduler timing.

**3. `testCleanupExpiredEntriesWithException` has same race**

Sets `properties` to null via reflection, then relies on `Thread.sleep(1100)` hoping the scheduler fires after — non-deterministic.

### Suggested Fix

Replace scheduler-dependent assertions with manual cleanup invocation:

```java
@AfterEach
void tearDown() {
    cache.shutdown();
}
```

For time-sensitive tests, disable auto-cleanup and trigger manually:

```java
props.setCleanupIntervalSeconds(3600); // won't fire during test
// ... setup entries, sleep for expiration ...
// Manually invoke cleanup via reflection
Method cleanup = MemoryMcpCacheIndex.class.getDeclaredMethod("cleanupExpiredEntries");
cleanup.setAccessible(true);
cleanup.invoke(cache);
// Now assert deterministically
```

### Additional Note

There are also thread-safety concerns in the implementation class itself (`removeIndex` has no lock, `addToHead` not synchronized, read-lock methods performing writes, `cleanupExpiredEntries` has no lock). The author has acknowledged these via TODO comments (L39-53) with plans to potentially replace with Guava Cache. These are separate from the test fix.

daguimu:
## Problem

`MemoryMcpCacheIndexTest.testCleanupExpiredEntriesDoesNotAffectValidEntries` intermittently fails in CI with NPE on `this.properties`. As analyzed in #14550, three factors combine to cause flakiness:

1. **Missing `@AfterEach` cleanup** — each `@BeforeEach` creates a new `MemoryMcpCacheIndex` with a background scheduler, but old instances are never shut down, causing 29 zombie scheduler threads.
2. **Background scheduler races with assertions** — tests use `cleanupIntervalSeconds=1`, so the async scheduler can fire at unpredictable times under CI load, cleaning up entries that tests expect to still be valid.
3. **Sleep-based assertions are non-deterministic** — `Thread.sleep` + `System.currentTimeMillis() / 1000` integer truncation + async scheduler timing create a triple jeopardy of imprecision.

## Root Cause

Tests rely on the background cleanup scheduler firing at specific times, but `Thread.sleep` imprecision under CI load combined with integer-second truncation in `CacheNode.isExpired()` makes timing-dependent assertions unreliable.

## Fix

- **Add `@AfterEach`** to shut down the default `cache` instance, preventing scheduler thread leaks across all 30+ test methods.
- **Set `cleanupIntervalSeconds=3600`** for test-created cache instances so the background scheduler effectively never fires during tests.
- **Invoke `cleanupExpiredEntries` directly via reflection** for deterministic assertions instead of relying on `Thread.sleep` + scheduler timing.
- **Extract `invokeCleanupExpiredEntries` helper method** to avoid duplication across test methods.

## Tests Added

This PR fixes existing flaky tests rather than adding new ones. The following tests are made deterministic:

- `testCleanupExpiredEntries` — now uses manual cleanup invocation instead of scheduler
- `testCleanupExpiredEntriesDoesNotAffectValidEntries` — the primary flaky test; scheduler race eliminated
- `testConcurrentAccessDuringCleanup` — scheduler disabled to avoid interfering with concurrent test operations
- `testCleanupExpiredEntriesWithException` — manual invocation replaces sleep-and-hope pattern
- `testCleanupExpiredEntriesAfterShutdown` — manual invocation replaces sleep-and-hope pattern

## Impact

- Test-only change, no production code modified
- Eliminates all scheduler-dependent race conditions in the test class
- Prevents scheduler thread leaks that could affect other tests in the suite
- No behavioral changes to the `MemoryMcpCacheIndex` implementation

Fixes #14550

### DIFF
--- a/ai/src/test/java/com/alibaba/nacos/ai/index/MemoryMcpCacheIndexTest.java
+++ b/ai/src/test/java/com/alibaba/nacos/ai/index/MemoryMcpCacheIndexTest.java
@@ -18,11 +18,13 @@ package com.alibaba.nacos.ai.index;
 
 import com.alibaba.nacos.ai.config.McpCacheIndexProperties;
 import com.alibaba.nacos.ai.model.mcp.McpServerIndexData;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
 import org.springframework.test.util.ReflectionTestUtils;
 
+import java.lang.reflect.Method;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -51,6 +53,13 @@ class MemoryMcpCacheIndexTest {
         props.setMaxSize(3);
         props.setExpireTimeSeconds(2); // 2秒过期
         cache = new MemoryMcpCacheIndex(props);
+    }
+    
+    // Shutdown the default cache after each test to prevent zombie scheduler thread leaks,
+    // since each @BeforeEach creates a new MemoryMcpCacheIndex with a background scheduler.
+    @AfterEach
+    void tearDown() {
+        cache.shutdown();
     }
     
     @Test
@@ -570,12 +579,14 @@ class MemoryMcpCacheIndexTest {
     }
     
     @Test
-    void testCleanupExpiredEntries() throws InterruptedException {
+    void testCleanupExpiredEntries() throws Exception {
         // 创建一个具有短过期时间的缓存实例
         McpCacheIndexProperties shortExpireProps = new McpCacheIndexProperties();
         shortExpireProps.setMaxSize(100);
         shortExpireProps.setExpireTimeSeconds(1); // 1秒过期
-        shortExpireProps.setCleanupIntervalSeconds(1);
+        // Use large interval to prevent the background scheduler from firing during the test,
+        // which would cause non-deterministic cleanup and race conditions under CI load.
+        shortExpireProps.setCleanupIntervalSeconds(3600);
         MemoryMcpCacheIndex shortExpireCache = new MemoryMcpCacheIndex(shortExpireProps);
         
         try {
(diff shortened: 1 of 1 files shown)


ai/src/test/java/com/alibaba/nacos/ai/index/MemoryMcpCacheIndexTest.java
```diff
@@ -609,34 +616,34 @@ void testCleanupExpiredEntries() throws InterruptedException {
...
             assertEquals("id1", mixedCache.getMcpId("ns1", "name1"));
             assertEquals("id2", mixedCache.getMcpId("ns2", "name2"));
-            
+
             // 再等待1.1秒，使第一个条目过期但第二个不过期
             Thread.sleep(1100);
-            
-            // 触发清理（通过获取来间接触发）
-            mixedCache.getMcpId("ns1", "name1");
-
```
KomachiSion, line 635: Please don't change indent

KomachiSion (changes requested):

KomachiSion (approved):

wuyfee:
$\color{red}{FAILURE}$
[DETAILS](https://github.com/alibaba/nacos/actions/runs/23579964632)
 ✅ - docker:   success 
 ❌ - deploy (standalone & cluster & standalone_auth):   failure 
 ❌ - e2e-java-test (standalone & cluster & standalone_auth):   skipped 
 ❌ - e2e-go-test (standalone & cluster):   skipped 
 ❌ - e2e-cpp-test (standalone & cluster):   skipped 
 ❌ - e2e-csharp-test (standalone & cluster):   skipped 
 ❌ - e2e-nodejs-test (standalone & cluster):   skipped 
 ❌ - e2e-python-test (standalone & cluster):   skipped 

PR #14729, opened 2026-03-25
fix: fix InputStream leak in ConfigImportAndExportService.handleEntity()

daguimu:
## Summary

- In `ExportHttpClientResponseHandler.handleEntity()`, the `InputStream` obtained from `HttpEntity.getContent()` is not closed after reading, causing a stream resource leak.
- Included the `InputStream` in the existing try-with-resources block alongside `ByteArrayOutputStream` to ensure proper cleanup.

## Test plan

- [ ] Verify config export handles entity content stream cleanup correctly
- [ ] Existing unit tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/console/src/main/java/com/alibaba/nacos/console/handler/impl/remote/config/ConfigImportAndExportService.java
+++ b/console/src/main/java/com/alibaba/nacos/console/handler/impl/remote/config/ConfigImportAndExportService.java
@@ -176,8 +176,8 @@ public class ConfigImportAndExportService {
         
         @Override
         public ResponseEntity<byte[]> handleEntity(HttpEntity entity) throws IOException {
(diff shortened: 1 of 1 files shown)


KomachiSion (commented):
I don't think so, this http entity is controlled by httpClient. The http client is an CloseableHttpClient and covered by try finally. So it should be auto closed and no leak.

daguimu:
Thanks for the explanation! You're right — the `InputStream` lifecycle is already managed by `CloseableHttpClient.execute(request, handler)`, which calls `EntityUtils.consume()` and `response.close()` after the handler returns, plus the outer `try-with-resources` on `CloseableHttpClient` as an additional safety net.

Closing this PR.

PR #14741, opened 2026-03-26
fix(config): add ORDER BY to findConfigInfoLike4PageFetchRows for deterministic pagination

### LINKED ISSUE #14046: When nacos run on mysql, the result of  ConfigInfoMapperByMySql.findConfigInfoLike4PageFetchRows is wrong in some cases
无论是mysql 5.27 还是8.0 均可以稳定复现此BUG

用POSTMAN直接访问配置分页模糊查询的接口，设置 Pagesize = 50， dataId = "someStr*"，注意someStr左边没有*

这里跟nacos控制台发起的调用不一样，nacos控制台发起的调用如果是模糊搜索， dataId的值是 “\*someStr\*”，两边都有*

此时，nacos第一页查出来的是按dataId的字典序升序的结果，第二页查出来的是全部结果集扫描后按默认顺序（也就是ID字段升序）排序后第二页的结果

根本原因是mysql设计如此，看大模型的总结

Image

解决办法：ConfigInfoMapperByMySql.findConfigInfoLike4PageFetchRows方法拼接SQL应该加上orderBy，提供默认的排序实现，避免mysql乱排序

==============================

This bug can be reliably reproduced in both MySQL 5.27 and 8.0.

Use Postman to directly access the interface configured for paginated fuzzy queries, setting `Pagesize = 50` and `dataId = "someStr*"`. Note that `someStr` does not have an asterisk (*) on the left.

This differs from calls initiated through the Nacos console. In Nacos console calls, if it's a fuzzy search, the value of `dataId` is "\*someStr\*".

In this case, the first page of Nacos retrieves results in lexicographical ascending order by `dataId`, and the second page retrieves results from the entire result set after scanning and sorting by the default order (i.e., ascending order of the ID field).

Solution: The `ConfigInfoMapperByMySql.findConfigInfoLike4PageFetchRows` method should include `orderBy` in the SQL concatenation to provide a default sorting implementation and prevent MySQL from missorting the data.

Anybody willing  to fix this problem?

daguimu:
## Problem

When using `findConfigInfoLike4PageFetchRows` in the MySQL implementation with `LIMIT/OFFSET` pagination, the result ordering is non-deterministic across pages. This causes duplicate or missing records when paginating through fuzzy query results — for example, page 1 returns results sorted by `dataId` lexicographically, but page 2 returns results sorted by primary key `id`, leading to inconsistent pagination.

## Root Cause

The inner SQL query in `ConfigInfoMapperByMySql.findConfigInfoLike4PageFetchRows()` applies `LIMIT` without an `ORDER BY` clause. Without explicit ordering, MySQL does not guarantee consistent row ordering across paginated queries, especially when the query optimizer chooses different execution plans for different offsets.

## Fix

- Added `ORDER BY id` before `LIMIT` in the inner query of `findConfigInfoLike4PageFetchRows()`, consistent with other paginated methods in the same class (e.g., `findAllConfigKey`, `findAllConfigInfoBaseFetchRows`, `findChangeConfigFetchRows`, `listGroupKeyMd5ByPageFetchRows`, `findAllConfigInfoFetchRows`).

## Tests Added

| Change Point | Test |
|-------------|------|
| Added `ORDER BY id` before `LIMIT` in inner query | `testFindConfigInfoLike4PageFetchRows()` — verifies the exact generated SQL contains `ORDER BY id LIMIT` |
| Same change point | `testFindConfigInfoLike4PageFetchRowsWithDescAndTags()` — verifies ORDER BY is present when all parameters (dataId, group, appName, content, types) are provided |

## Impact

This fix ensures deterministic pagination ordering for fuzzy config queries on MySQL. Only the `findConfigInfoLike4PageFetchRows` method in the MySQL mapper is changed. No behavioral changes for other query methods or other database dialects.

Fixes #14046

### DIFF
--- a/plugin-default-impl/nacos-default-datasource-plugin/nacos-datasource-plugin-mysql/src/main/java/com/alibaba/nacos/plugin/datasource/impl/mysql/ConfigInfoMapperByMySql.java
+++ b/plugin-default-impl/nacos-default-datasource-plugin/nacos-datasource-plugin-mysql/src/main/java/com/alibaba/nacos/plugin/datasource/impl/mysql/ConfigInfoMapperByMySql.java
@@ -278,8 +278,8 @@ public class ConfigInfoMapperByMySql extends AbstractMapperByMysql implements Co
         }
         
         // 先分页，减少后续 JOIN 的数据量
-        innerSql.append(" LIMIT ").append(context.getStartRow()).append(",").append(context.getPageSize());
-        
+        innerSql.append(" ORDER BY id LIMIT ").append(context.getStartRow()).append(",").append(context.getPageSize());
+
         // 外层查询：对分页后的结果进行标签关联
         final String sql = "SELECT a.id,a.data_id,a.group_id,a.tenant_id,a.app_name,a.content,a.md5,"
                           + "a.encrypted_data_key,a.type,a.c_desc,a.gmt_modified,"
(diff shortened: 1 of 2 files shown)


wuyfee:
$\color{red}{FAILURE}$
[DETAILS](https://github.com/alibaba/nacos/actions/runs/23594568711)
 ✅ - docker:   success 
 ❌ - deploy (standalone & cluster & standalone_auth):   failure 
 ❌ - e2e-java-test (standalone & cluster & standalone_auth):   skipped 
 ❌ - e2e-go-test (standalone & cluster):   skipped 
 ❌ - e2e-cpp-test (standalone & cluster):   skipped 
 ❌ - e2e-csharp-test (standalone & cluster):   skipped 
 ❌ - e2e-nodejs-test (standalone & cluster):   skipped 
 ❌ - e2e-python-test (standalone & cluster):   skipped 

KomachiSion (approved):

PR #14754, opened 2026-03-26
fix(naming): include cluster field in InstancePublishInfo equals and hashCode

daguimu:
## Problem

`InstancePublishInfo.equals()` and `hashCode()` do not include the `cluster` field. This means two instances with the same IP and port but in different clusters are considered equal, which can cause Distro data sync to miss cluster changes across Nacos nodes.

## Root Cause

The `cluster` field is part of the instance identity — it is used in `getMetadataId()` to generate the metadata identifier (`ip:port:cluster`). However, it was omitted from `equals()` and `hashCode()`:

```java
// Before: cluster missing
return port == that.port && healthy == that.healthy
    && Objects.equals(ip, that.ip) && Objects.equals(extendDatum, that.extendDatum);
```

In `DistroClientDataProcessor`, `equals()` is used to detect instance changes during cluster sync:
```java
if (!instancePublishInfo.equals(client.getInstancePublishInfo(singleton))) {
```

Without `cluster` in equals, changing an instance's cluster will not be detected as a change, causing data inconsistency between Nacos cluster nodes.

## Fix

Added `cluster` to both `equals()` and `hashCode()`:

```java
// After: cluster included
return port == that.port && healthy == that.healthy
    && Objects.equals(ip, that.ip) && Objects.equals(cluster, that.cluster)
    && Objects.equals(extendDatum, that.extendDatum);
```

## Tests Added

Created `InstancePublishInfoTest` with 8 test cases:

| Change Point | Test |
|-------------|------|
| cluster in equals | `testEqualsWithSameCluster()` — same cluster returns equal |
| cluster in equals | `testNotEqualsWithDifferentCluster()` — different cluster returns not equal |
| null-safety | `testEqualsWithNullCluster()` — null clusters are handled safely |
| hashCode consistency | `testEqualsWithSameCluster()` — equal objects have equal hashCodes |
| existing behavior preserved | `testNotEqualsWithDifferentPort()`, `testNotEqualsWithDifferentHealthy()`, `testEqualsSameObject()`, `testNotEqualsNull()` |
| metadataId uses cluster | `testGetMetadataId()` — verifies cluster is part of metadata identity |

## Impact

Fixes Distro sync detection for cluster changes. `InstancePublishInfo` is used as map VALUES (not keys) in `AbstractClient.publishers`, so this change does not affect HashMap key behavior.

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/core/v2/pojo/InstancePublishInfo.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/core/v2/pojo/InstancePublishInfo.java
@@ -104,13 +104,13 @@ public class InstancePublishInfo implements Serializable {
             return false;
         }
         InstancePublishInfo that = (InstancePublishInfo) o;
-        return port == that.port && healthy == that.healthy && Objects.equals(ip, that.ip) && Objects
-                .equals(extendDatum, that.extendDatum);
+        return port == that.port && healthy == that.healthy && Objects.equals(ip, that.ip)
+                && Objects.equals(cluster, that.cluster) && Objects.equals(extendDatum, that.extendDatum);
     }
-    
+
     @Override
     public int hashCode() {
-        return Objects.hash(ip, port, extendDatum, healthy);
+        return Objects.hash(ip, port, healthy, cluster, extendDatum);
     }
     
     @Override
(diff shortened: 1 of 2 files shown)


KomachiSion:
Sorry, this is not a bug. we design one instance only allow specified to one cluster.

You can see that, you deploy one service instance, it only deploy one logic situation, such as zone, env, and other custom classifications.

So cluster can be calculate for same instance.

PR #14772, opened 2026-03-28
fix(console-ui-next): use customNamespaceId when creating namespace

WyRainBow:
Fix new UI namespace creation to send customNamespaceId (with backward-compatible mapping from namespaceId) so custom namespace IDs are preserved instead of being auto-generated as UUIDs.


KomachiSion:
Fixed by #14773

PR #14779, opened 2026-03-29
fix(common): use ConcurrentHashMap for fileMd5Map in TlsFileWatcher

daguimu:
## Problem

`TlsFileWatcher.fileMd5Map` is a plain `HashMap` that is accessed from multiple threads concurrently, which can lead to data races, missed TLS certificate change notifications, or `ConcurrentModificationException`.

## Root Cause

`fileMd5Map` is written by the calling thread in `addFileChangeListener()` (line 87) and read/written by the scheduled executor thread in the periodic file-check task (lines 114, 116). A `HashMap` is not thread-safe — concurrent reads and writes can corrupt its internal structure or produce inconsistent results.

## Fix

Replace `HashMap<String, String>` with `ConcurrentHashMap<String, String>` for `fileMd5Map`, matching the pattern already used for `watchFilesMap` in the same class.

## Tests Added

| Change Point | Test |
|---|---|
| `fileMd5Map` changed from HashMap to ConcurrentHashMap | `testFileMd5MapIsConcurrentHashMap()` — verifies the map instance is ConcurrentHashMap |
| Existing behavior preserved | All 6 existing tests continue to pass (listener registration, change detection, duplicate start guard) |

## Impact

Only affects `TlsFileWatcher` internal state management. No change to public API or behavior — the fix ensures existing thread-safety expectations are met.

### DIFF
--- a/common/src/main/java/com/alibaba/nacos/common/tls/TlsFileWatcher.java
+++ b/common/src/main/java/com/alibaba/nacos/common/tls/TlsFileWatcher.java
@@ -29,7 +29,6 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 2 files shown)


shiyiyue1102:

---

KomachiSion (dismissed):

KomachiSion (approved):

PR #14780, opened 2026-03-29
fix(config): make ClientIpWhiteList.isOpen volatile and null-safe

daguimu:
## Problem

`ClientIpWhiteList.isOpen` has two issues:

1. **Visibility bug**: The field is written by the config-load thread (in `load()`) and read by HTTP request threads (in `isEnableWhitelist()`). Without `volatile`, the Java Memory Model does not guarantee cross-thread visibility, so request threads may read a stale value indefinitely.

2. **NullPointerException risk**: `isOpen` is a `Boolean` wrapper, and `acl.getIsOpen()` can return `null` when the JSON field is absent. The `isEnableWhitelist()` method returns primitive `boolean`, so a `null` `isOpen` would cause auto-unboxing NPE.

## Root Cause

- `private static Boolean isOpen = false;` — not `volatile`, not thread-safe for cross-thread reads/writes.
- `isOpen = acl.getIsOpen();` — no null guard, yet downstream returns primitive `boolean`.

## Fix

- Change `Boolean isOpen` to `volatile boolean isOpen` for guaranteed visibility.
- Use `Boolean.TRUE.equals(acl.getIsOpen())` for null-safe assignment — treats `null` as `false`.

## Tests Added

| Change Point | Test |
|---|---|
| `isOpen` field is volatile | `testIsOpenFieldIsVolatile()` — reflection check on field modifiers |
| Valid JSON enables whitelist | `testLoadValidContentEnablesWhitelist()` — verifies isOpen=true and IP matching |
| Blank content disables whitelist | `testLoadBlankContentDisablesWhitelist()` — load then clear |
| Null isOpen does not throw NPE | `testLoadNullIsOpenDoesNotThrowNpe()` — JSON without isOpen field |
| isOpen=false works correctly | `testLoadWithIsOpenFalse()` — explicit false |
| Blank/null client IP validation | `testIsLegalClientThrowsOnBlankInput()` — boundary check |
| Invalid JSON does not crash | `testLoadInvalidJsonDoesNotCrash()` — malformed input |

## Impact

Only affects `ClientIpWhiteList` internal state. No API or behavioral change for correctly-formed inputs — the fix prevents stale reads and NPE on malformed ACL config.

### DIFF
--- a/config/src/main/java/com/alibaba/nacos/config/server/service/ClientIpWhiteList.java
+++ b/config/src/main/java/com/alibaba/nacos/config/server/service/ClientIpWhiteList.java
@@ -41,7 +41,7 @@ public class ClientIpWhiteList {
     private static final AtomicReference<List<String>> CLIENT_IP_WHITELIST = new AtomicReference<>(
         new ArrayList<>());
     
-    private static Boolean isOpen = false;
+    private static volatile boolean isOpen = false;
     
     /**
      * Judge whether specified client ip includes in the whitelist.
@@ -85,7 +85,7 @@ public class ClientIpWhiteList {
         DEFAULT_LOG.warn("[clientIpWhiteList] {}", content);
         try {
             AclInfo acl = JacksonUtils.toObj(content, AclInfo.class);
-            isOpen = acl.getIsOpen();
+            isOpen = Boolean.TRUE.equals(acl.getIsOpen());
             CLIENT_IP_WHITELIST.set(acl.getIps());
         } catch (Exception ioe) {
             DEFAULT_LOG.error("failed to load clientIpWhiteList, " + ioe.toString(), ioe);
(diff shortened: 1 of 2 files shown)


KomachiSion:
@daguimu Hi, 

we are welcome you contribute nacos community and do many PR help nacos to do enhancement and bugfix.

But I suggest you submit issue to describe issues before submit PRs.

KomachiSion:
And welcome provide an dingtalk number, which we can invite you join the nacos community contributor groups.

daguimu:
> And welcome provide an dingtalk number, which we can invite you join the nacos community contributor groups.
my dingtalk number is bazs8u9

shiyiyue1102:

---

KomachiSion (dismissed):

KomachiSion:
Sorry, nacos currently is doing spotless auto reformat code works so that your pr is conflict with this work.

Please solve conflict and use `mvn spotless:apply` after you solve conflicts

KomachiSion (approved):

PR #14795, opened 2026-03-30
Fix NPE when importing config without metadata file

Senrian:
## Bug Description

Issue: #14785

When importing a ZIP configuration file without the required metadata file (`.meta.yml` or `.metadata.yml`), the server throws a NullPointerException:

```
Cannot invoke "com.alibaba.nacos.config.server.utils.ZipUtils$ZipItem.getItemData()" because "metaDataItem" is null
```

## Root Cause

In `ConfigControllerV3.parseImportDataV2()`, the code directly calls `metaDataItem.getItemData()` without checking if `metaDataItem` is null.

## Fix

Add a null check before accessing `metaDataItem`:

```java
ZipUtils.ZipItem metaDataItem = unziped.getMetaDataItem();
Map<String, Object> failedData = new HashMap<>(4);

if (metaDataItem == null) {
    failedData.put("succCount", 0);
    return Result.failure(ErrorCode.METADATA_ILLEGAL, failedData);
}

String metaData = metaDataItem.getItemData();
```


KomachiSion:
I don't get the reason why add a new api to do metadata?

Senrian:
Hi @KomachiSion, this PR does not add a new API. It only adds a null check in the existing `parseImportDataV2()` method. When a ZIP file without `.meta.yml`/`.metadata.yml` is uploaded, `metaDataItem` becomes null and causes NPE at `metaDataItem.getItemData()`. The fix simply returns an error result with `ErrorCode.METADATA_ILLEGAL` instead of crashing. No new API is introduced.

KomachiSion:
Please change the basic branch to `develop`

KomachiSion:
> Hi @KomachiSion, this PR does not add a new API. It only adds a null check in the existing `parseImportDataV2()` method. When a ZIP file without `.meta.yml`/`.metadata.yml` is uploaded, `metaDataItem` becomes null and causes NPE at `metaDataItem.getItemData()`. The fix simply returns an error result with `ErrorCode.METADATA_ILLEGAL` instead of crashing. No new API is introduced.

Sorry I see the api PUT /metadata already in develop branch.  Please base on develop branch and fix again.

The PR I will closed.

Issue #14806, opened 2026-03-31
/v3/admin/cs/metrics/cluster still returns success (code=0) when a member node fails, causing partial failure to be hidden


CHEN666333-SVG:
Description
In a 2-node cluster, when requesting `/v3/admin/cs/metrics/cluster` and one member is unreachable, the API still returns HTTP 200 with business `code=0`.
However, server logs clearly show member fetch failure (`Connection refused`).
This makes clients believe aggregation succeeded completely, while the result is actually partial/incomplete.

Environment
- Nacos: 3.2.1-SNAPSHOT
- Java: 17.0.17
- OS: Windows
- Deployment: local 2-node cluster (127.0.0.1:8848, 127.0.0.1:8850)
- Auth: disabled (`nacos.core.auth.enabled=false`, `nacos.core.auth.admin.enabled=false`)
- Note: using 8848 + 8850 on same host to avoid local port conflicts

Steps to Reproduce
1. Start 2 nodes: 8848 and 8850.
2. Baseline request with both nodes alive:  
   `GET /nacos/v3/admin/cs/metrics/cluster?ip=127.0.0.1&dataId=test&groupName=DEFAULT_GROUP&namespaceId=public`
3. Stop node 8850.
4. Send the same request to node 8848 again.

Actual Result
- API still returns success:  
  `after-stop status=200 body={"code":0,"message":"success","data":{"code":0,"message":"success","data":{}}}`
- node1 log shows member fetch error:  
  `2026-03-31 22:17:22,711 ERROR Get config metrics error from member address=127.0.0.1:8850, ip=127.0.0.1,dataId=test,group=DEFAULT_GROUP,namespaceId=public,error={}`
- Connection refused exception is present:  
  `org.apache.hc.client5.http.HttpHostConnectException: Connect to http://127.0.0.1:8850 ... Connection refused`

Expected Result
When any member fails/timeouts during aggregation, the API should do at least one of:
1. Return non-success status/code.
2. Include explicit partial-failure indicators in response (e.g. `failedMembers` / `timeoutMembers` / `partial=true`).
3. Avoid returning pure success with empty/incomplete data and no failure signal.

Impact
Client cannot distinguish partial failure from full success.

Attachments
- `node1-metrics-failure-snippet.log` (primary evidence)
- `node2-restart-window-snippet.log` (optional context)

[node1-metrics-failure-snippet.log](https://github.com/user-attachments/files/26382997/node1-metrics-failure-snippet.log)
[...]

PR #14818, opened 2026-04-01
[ISSUE #14806] Improve cluster metrics aggregation completeness signal in v3 API

CHEN666333-SVG:
## What is the purpose of the change

This PR improves v3 cluster config metrics aggregation by adding an aggregate-level completeness signal while keeping node-level failure details hidden.

Previously, `/v3/admin/cs/metrics/cluster` could still return a successful response when aggregation was partial (for example timeout, interruption, callback error, or cancellation), but callers had no explicit way to distinguish complete vs partial merged results.

This change introduces a `complete` flag in the response data:
- `complete = true`: all member callbacks completed successfully
- `complete = false`: aggregation is partial/incomplete

It also fixes `InterruptedException` handling in the aggregation path by using logger-based handling and restoring thread interrupt status.

## Brief changelog

- Keep existing API philosophy: do not expose per-node failure details in response.
- Add aggregate-level completeness indicator `complete` in cluster metrics response.
- Keep merged metrics behavior unchanged.
- Improve interruption handling in cluster aggregation flow:
  - remove printStackTrace-style handling
  - log with logger
  - restore interrupt status
- Update unit tests for the new behavior:
  - add controller-level test to assert `complete=false` when partial failure occurs
  - keep existing callback-level behavior checks

## Verifying this change

- Unit test:
  - `MetricControllerV3Test`
  - Result: `Tests run: 4, Failures: 0, Errors: 0, Skipped: 0`
- Local manual verification:
  - Use local 2-node no-auth setup
  - Stop one node and call `/v3/admin/cs/metrics/cluster`
  - Response still returns merged data and includes `complete=false` for partial aggregation
  - No node-level failure details are returned

Fixes #14806

Follow this checklist to help us incorporate your contribution quickly and easily:

* [x] Make sure there is a Github issue filed for the change (usually before you start working on it). Trivial changes like typos do not require a Github issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.
* [x] Format the pull request title like `[ISSUE #123] Fix UnknownException when host config not exist`. Each commit in the pull request should have a meaningful subject line and body.
* [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
* [x] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/alibaba/nacos/tree/master/test).
[...]

PR #14827, opened 2026-04-01
fix: mark cluster metrics response as failed when member is unreachable (#14806)

Senrian:
## Bug Description

Issue: #14806

When calling `/v3/admin/cs/metrics/cluster` in a multi-node Nacos cluster, if one member node is unreachable, the API still returns HTTP 200 with business `code=0` (success), even though the result is incomplete/partial.

**Root Cause:** In `ClusterMetricsCallBack.onError()`, only the error is logged but `responseMap` is never updated to signal the failure. After `latch.await()` completes (3 seconds timeout), the endpoint returns `Result.success(responseMap)` with whatever was collected — which may be empty if all nodes failed.

## Fix

Add `responseMap.put("__error__", "failed")` in `onError()` so clients can distinguish between:
- Full success: has data, no `__error__` key
- Partial/total failure: `__error__=failed` present in response

```java
// Before:
public void onError(Throwable throwable) {
    Loggers.CORE.error("Get config metrics error from member...", throwable);
    latch.countDown();
}

// After:
public void onError(Throwable throwable) {
    Loggers.CORE.error("Get config metrics error from member...", throwable);
    responseMap.put("__error__", "failed");
    latch.countDown();
}
```

## Impact

Clients can now reliably detect when cluster metrics aggregation partially or completely failed due to member node issues.

### DIFF
--- a/config/src/main/java/com/alibaba/nacos/config/server/controller/v3/MetricsControllerV3.java
+++ b/config/src/main/java/com/alibaba/nacos/config/server/controller/v3/MetricsControllerV3.java
@@ -166,6 +166,7 @@ public class MetricsControllerV3 {
             Loggers.CORE.error(
                     "Get config metrics error from member address={}, ip={},dataId={},group={},namespaceId={},error={}",
                     member.getAddress(), ip, dataId, group, namespaceId, throwable);
+            responseMap.put("__error__", "failed");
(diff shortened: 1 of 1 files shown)


KomachiSion:
Fixed by #14818

PR #14927, opened 2026-04-14
[Bug] Fix TOCTOU race condition and thread leak in ClientWorker.ensureSyncExecutor()

LiyunZhang10:
## Issue Description
In `ClientWorker.java`, `multiTaskExecutor` was defined as a standard `HashMap` and accessed concurrently using a non-atomic check-then-act (`containsKey` + `put`) pattern. 

When multiple threads request the same `taskId` concurrently, each thread could evaluate `!containsKey(taskId)` as true, leading to multiple `ThreadPoolExecutor` instances being created. Since only the last one inserted into the map is referenced and eventually shut down, the others become permanent thread leaks, potentially causing OOM or CPU exhaustion.

Furthermore, iterating over `HashMap.values()` during `shutdown()` could throw `ConcurrentModificationException` if a concurrent `ensureSyncExecutor()` call occurred.

## Proposed Changes
- Changed `multiTaskExecutor` from `HashMap` to `ConcurrentHashMap`.
- Replaced the `containsKey` + `put` pattern with `computeIfAbsent()` to guarantee that the `ThreadPoolExecutor` is created exactly once per `taskId`, resolving both the thread leak and iteration hazards.
- Verified locally with 20-thread concurrency tests: 100% eliminated the ThreadPool leakage observed in the original code.

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/config/impl/ClientWorker.java
+++ b/client/src/main/java/com/alibaba/nacos/client/config/impl/ClientWorker.java
@@ -93,6 +93,7 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingQueue;
@@ -651,7 +652,7 @@ public class ClientWorker implements Closeable {
     
     public class ConfigRpcTransportClient extends ConfigTransportClient {
         
-        Map<String, ExecutorService> multiTaskExecutor = new HashMap<>();
+        Map<String, ExecutorService> multiTaskExecutor = new ConcurrentHashMap<>();
 
         private ExecutorService listenExecutor;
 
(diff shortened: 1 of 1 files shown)


client/src/main/java/com/alibaba/nacos/client/config/impl/ClientWorker.java
```diff
@@ -651,7 +651,7 @@ public boolean isHealthServer() {
     
     public class ConfigRpcTransportClient extends ConfigTransportClient {
         
-        Map<String, ExecutorService> multiTaskExecutor = new HashMap<>();
+        Map<String, ExecutorService> multiTaskExecutor = new java.util.concurrent.ConcurrentHashMap<>();
```
KomachiSion: Please don't use full package if no class conflict.

KomachiSion (changes requested):

client/src/main/java/com/alibaba/nacos/client/config/impl/ClientWorker.java
LiyunZhang10: Thanks for the review! Updated to use import java.util.concurrent.ConcurrentHashMap; and removed the full package name as suggested.

KomachiSion (approved):

Issue #14932, opened 2026-04-15
[ISSUE] Refactor MemoryMcpCacheIndex to remove read-path LRU contention and improve cache consistency


YutShang:
## Background

`McpCacheIndex` is the cache abstraction used by MCP-related query paths, and its default implementation is `MemoryMcpCacheIndex`.

The current implementation maintains:
- in-memory `name -> id` and `id -> data` indexes
- a hand-rolled LRU doubly linked list
- capacity eviction
- periodic cleanup

However, the current design has a concurrency and maintainability problem:
- cache read paths (`getMcpId`, `getMcpServerById`) also update LRU order
- this makes reads effectively carry write semantics
- the implementation currently relies on `ReadWriteLock` plus `synchronized` linked-list operations to keep the list consistent
- under high-QPS read scenarios, this can become a hotspot and reduces the benefit of concurrent reads

## Problem

There are three main concerns in the current implementation:

1. **Concurrency model mismatch**
   - read operations mutate the LRU list
   - multiple concurrent readers can still contend on internal mutation paths

2. **Performance hotspot**
   - `synchronized` protection around LRU node movement serializes hot reads
   - this can degrade throughput in MCP lookup scenarios

3. **Complexity and consistency risk**
   - the cache maintains two maps plus a custom linked list
   - expiration, eviction, and invalidation need to keep multiple structures consistent
   - removing by ID may require extra cleanup work to keep `name -> id` mappings correct

## Proposal

Replace the hand-written LRU implementation in `MemoryMcpCacheIndex` with a Caffeine-backed local cache while keeping the `McpCacheIndex` interface unchanged.

Recommended direction:
- keep `McpCacheIndex` API unchanged
- keep `MemoryMcpCacheIndex` as the default implementation class to minimize integration churn
- use Caffeine for:
  - maximum size based eviction
  - expire-after-write semantics
  - access tracking / eviction policy management
- keep `name -> id` mapping as a lightweight secondary index
- use cache removal callbacks to clean related name mappings and preserve index consistency
- keep existing configuration properties as much as possible to avoid operational impact

## Expected Benefits

- remove read-path manual LRU mutation and lock contention
- reduce custom concurrent data structure complexity
- improve maintainability and correctness
- keep changes localized to the `ai` module and minimize PR merge conflict risk
- preserve current business-facing interface and wiring

## Compatibility Notes

To reduce review risk, this refactor should:
- avoid changing `McpCacheIndex` consumers
- preserve current configuration names
[...]

PR #14933, opened 2026-04-15
[ISSUE #14932] Refactor MemoryMcpCacheIndex with Caffeine to remove read-path LRU contention

YutShang:
##Related to #14932

## Summary

This PR refactors `MemoryMcpCacheIndex` from a hand-rolled `ReadWriteLock + doubly linked list` LRU implementation to a Caffeine-backed local cache.

The goal is to remove read-path LRU mutation contention, simplify cache maintenance logic, and improve `name -> id` / `id -> data` index consistency while keeping the external `McpCacheIndex` contract unchanged.

## Background

The previous implementation updated LRU order on cache reads. That meant:
- read paths had write semantics
- the linked-list mutation path still needed synchronization
- hot lookup traffic could be serialized on internal LRU maintenance

This PR replaces that custom logic with Caffeine and keeps the change localized to the `ai` module.

## Main Changes

- replace custom linked-list LRU management in `MemoryMcpCacheIndex`
- use Caffeine for:
  - maximum size eviction
  - expire-after-write handling
  - eviction/removal callbacks
- keep the `McpCacheIndex` interface unchanged
- preserve the current bean wiring and configuration model
- maintain the secondary `name -> id` index and clean it through removal callbacks
- add focused regression tests for:
  - basic put/get behavior
  - name/id mapping consistency
  - expiration cleanup
  - eviction behavior
  - concurrent access stability
  - shutdown behavior

## Why this approach

Compared with only tightening the lock strategy, this refactor:
- removes the custom concurrent LRU structure entirely
- reduces the complexity of keeping map/list state consistent
- avoids turning cache reads into a synchronized hotspot
- keeps integration risk low because callers do not need to change

## Compatibility

- `McpCacheIndex` API is unchanged
- existing callers in `ai` module remain unchanged
- existing cache configuration properties remain available
- expiration semantics use `expireAfterWrite`, which stays close to the prior implementation behavior

## Verification

Executed:

```bash
mvn -pl ai "-Dtest=MemoryMcpCacheIndexCaffeineTest,CachedMcpServerIndexTest,McpCacheIndexPropertiesTest" test


KomachiSion:
From the discussion in issue, we need to discuss whether import new depend Caffeine, or use guava LRU which already depend.

PR close first.

PR #14951, opened 2026-04-17
fix: SocketChannel leak + exact IP match in hasMember()

srpatcha:
### 1. Use exact IP match instead of substring match in `hasMember()`
Prevents false positive membership checks.

### 2. Fix SocketChannel leak on successful health check (`TcpHealthCheckProcessor.java`)
In `PostProcessor.run()`, when a TCP health check succeeds (connection is established), `finishCheck()` is called but the `SocketChannel` is never closed. Added `key.cancel()` and `key.channel().close()` after successful checks to prevent resource leaks.

PR #14953, opened 2026-04-19
Fix #14951: hasMember() false positives due to StringUtils.contains() fuzzy match

Senrian:
### Problem
The `hasMember()` method in `ServerMemberManager.java` used `StringUtils.contains()` for fuzzy IP matching, causing false positives. For example:
- `hasMember("8.8")` would incorrectly return `true` when `"10.8.8.1:8848"` exists in the server list

### Solution
Replace the fuzzy `StringUtils.contains(entry.getKey(), address)` check with an exact IP extraction and match:

```java
String keyIp = entry.getKey().split(":")[0];
if (keyIp.equals(address)) {
    result = true;
    break;
}
```

This extracts the IP portion from the address key (before the colon) and compares it exactly against the provided address, ensuring only exact IP matches are found.

### Files Changed
- `core/src/main/java/com/alibaba/nacos/core/cluster/ServerMemberManager.java`

### Issue Reference
Fixes #14951

### DIFF
--- a/core/src/main/java/com/alibaba/nacos/core/cluster/ServerMemberManager.java
+++ b/core/src/main/java/com/alibaba/nacos/core/cluster/ServerMemberManager.java
@@ -294,9 +294,10 @@ public class ServerMemberManager implements NacosMemberManager {
             return true;
         }
         
-        // If only IP information is passed in, a fuzzy match is required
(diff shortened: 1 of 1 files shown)


KomachiSion:
CLA problem don't solved for a long time, close pr

PR #14960, opened 2026-04-20
[BUG] Fix cross-thread visibility and race conditions for ManagerListenerWrap inNotifying

LiyunZhang10:
This PR fixes a critical cross-thread visibility and race condition bug in the `ManagerListenerWrap` inner class within `CacheData.java`.
Currently, the `inNotifying` field is a plain `boolean`. It is read by the Nacos Client polling thread (`checkListenerMd5`) and written asynchronously by user-provided custom `Executor` threads inside the `NotifyTask`.
This causes two severe concurrency issues:
1. **Visibility Issue:** Because `inNotifying` is not `volatile`, JIT compiler optimizations may cache the value. As a result, the Nacos polling thread might never see the `inNotifying = false` update from the user's thread, causing the listener to be permanently skipped for future configuration change events.
2. **Atomicity Issue (Race Condition):** The `if (listenerWrap.inNotifying)` check and the subsequent setting to `true` have no synchronization. Multiple checking threads can enter simultaneously, leading to the same listener being invoked concurrently multiple times.
## Brief changelog
* Refactored the `inNotifying` field from a plain `boolean` to `final AtomicBoolean inNotifying`.
* Used `inNotifying.compareAndSet(false, true)` to atomically perform the status check and the lock acquisition at the entry of `safeNotifyListener`, preventing concurrent duplicate notifications.
* Removed the delayed `inNotifying = true` assignment inside `NotifyTask.run()` as it is now securely acquired upfront.
* Replaced `inNotifying = false` with `inNotifying.set(false)` in the `finally` block to correctly and visibly release the lock."

### DIFF
--- a/client/src/main/java/com/alibaba/nacos/client/config/impl/CacheData.java
+++ b/client/src/main/java/com/alibaba/nacos/client/config/impl/CacheData.java
@@ -419,7 +419,7 @@ public class CacheData {
     private void safeNotifyListener(final String dataId, final String group, final String content, final String type,
             final String md5, final String encryptedDataKey, final ManagerListenerWrap listenerWrap) {
         final Listener listener = listenerWrap.listener;
-        if (listenerWrap.inNotifying) {
+        if (!listenerWrap.inNotifying.compareAndSet(false, true)) {
             LOGGER.warn(
                     "[{}] [notify-currentSkip] dataId={}, group={},tenant={}, md5={}, listener={}, listener is not finish yet,will try next time.",
                     envName, dataId, group, tenant, md5, listener);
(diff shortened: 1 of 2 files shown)


KomachiSion (changes requested):
这个修改是错误的， 如果不是在运行过程中设置这个值， 可能会导致listenerWrap无法复原。

KomachiSion:
re-open for cla

LiyunZhang10:
> 这个修改是错误的， 如果不是在运行过程中设置这个值， 可能会导致listenerWrap无法复原。

您好，我现在在 catch 块中加上了兜底释放锁的逻辑 listenerWrap.inNotifying.set(false); 请再次 review，感谢！

KomachiSion (approved):

PR #14996, opened 2026-04-29
fix(auth): make AbstractCachedUserService userMap volatile

daguimu:
## Problem

`AbstractCachedUserService.userMap` is a non-volatile field that gets reassigned every 15 seconds by the `@Scheduled reload()` method. Readers on the auth hot path (`NacosUserServiceDirectImpl.loadUserByUsername`, `NacosUserServiceRemoteImpl.getUserFromDatabase`) call `getCachedUserMap()` to look up users for login and authorization, but without `volatile` they have no happens-before guarantee that they will observe the swap. The result is stale auth data: a user just created may not be findable, or a user just deleted may still be served from cache for arbitrarily long.

## Root Cause

`reload()` builds a fresh `ConcurrentHashMap` locally and publishes it via plain reference assignment:

```java
Map<String, User> map = new ConcurrentHashMap<>(16);
for (User user : users.getPageItems()) {
    map.put(user.getUsername(), user);
}
userMap = map; // not volatile — JMM does not guarantee visibility to readers
```

The two sister caches in the same module already follow the volatile pattern for exactly this reason:

- `AbstractCachedRoleService` — `roleSet`, `roleInfoMap`, `permissionInfoMap` are all `volatile`
- `CachedJwtTokenManager` — `tokenMap`, `userMap` are both `volatile`

`AbstractCachedUserService.userMap` was the only outlier.

## Fix

- Add `volatile` to `userMap` so the periodic reference swap in `reload()` is published with proper visibility to readers calling `getCachedUserMap()`.

## Tests Added

New `AbstractCachedUserServiceTest` covering the change point and surrounding reload semantics:

| Change point / behavior | Test |
|-------------------------|------|
| `userMap` field declared `volatile` | `testCachedUserMapFieldIsVolatile` — reflection check; verified to fail when `volatile` is removed and pass when present |
| `reload()` swaps in a fresh map populated from the source page | `testReloadReplacesCachedUserMap` |
| `reload()` keeps the existing cache when source returns `null` | `testReloadKeepsExistingMapWhenSourceReturnsNull` |
| `reload()` keeps the existing cache when source throws | `testReloadSwallowsErrorsAndKeepsExistingMap` |

The reflection assertion was confirmed to flip with the modifier — without `volatile` the test reports `expected: <true> but was: <false>` and the build fails. All 163 tests in `nacos-default-auth-plugin` pass with the fix applied; checkstyle is clean.

## Impact

- Scope: one-line modifier change on a single private field plus a new dedicated test class.
- Affects: any deployment whose auth plugin uses `AbstractCachedUserService` (`NacosUserServiceDirectImpl`, `NacosUserServiceRemoteImpl`). User cache freshness on multi-core hardware is now guaranteed by the JMM rather than left to incidental synchronization.
- Does not affect on-disk persistence, database access, public APIs, configuration, or behavior of the `roleSet` / token caches — those were already correctly declared `volatile`.

### DIFF
--- a/plugin-default-impl/nacos-default-auth-plugin/src/main/java/com/alibaba/nacos/plugin/auth/impl/users/AbstractCachedUserService.java
+++ b/plugin-default-impl/nacos-default-auth-plugin/src/main/java/com/alibaba/nacos/plugin/auth/impl/users/AbstractCachedUserService.java
@@ -33,7 +33,7 @@ import java.util.concurrent.ConcurrentHashMap;
  */
 public abstract class AbstractCachedUserService implements NacosUserService {
     
-    private Map<String, User> userMap = new ConcurrentHashMap<>();
+    private volatile Map<String, User> userMap = new ConcurrentHashMap<>();
     
     protected AbstractCachedUserService() {
     }
--- a/plugin-default-impl/nacos-default-auth-plugin/src/test/java/com/alibaba/nacos/plugin/auth/impl/users/AbstractCachedUserServiceTest.java
+++ b/plugin-default-impl/nacos-default-auth-plugin/src/test/java/com/alibaba/nacos/plugin/auth/impl/users/AbstractCachedUserServiceTest.java
@@ -1,0 +1,167 @@
+/*
+ * Copyright 1999-2025 Alibaba Group Holding Ltd.
+ *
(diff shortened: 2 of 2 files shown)


KomachiSion:
reopen for cla

shiyiyue1102:

---

KomachiSion (approved):

PR #14998, opened 2026-04-29
fix(auth): read cached map once in remote auth services to avoid TOCTOU

daguimu:
## Problem

Three look-up methods in the remote auth services consult the cached map twice in a row through two separate `getCached*()` accessor calls:

- `NacosUserServiceRemoteImpl.getUser(String)`
- `NacosRoleServiceRemoteImpl.getPermissions(String)`
- `NacosRoleServiceRemoteImpl.getRoles(String)`

Each one followed the same shape:

```java
if (getCachedUserMap().containsKey(username)) {
    return getCachedUserMap().get(username);
}
reload();
return getCachedUserMap().get(username);
```

The two cache accessors back onto fields published by `@Scheduled reload()` in `AbstractCachedUserService` / `AbstractCachedRoleService`, which rebuild a fresh `ConcurrentHashMap` and reassign the field every 15 s. The scheduled swap can interleave between the `containsKey` call and the matching `get` call, so the second accessor returns a value from a different snapshot than the one we just verified.

## Root Cause

A classic check-then-act / TOCTOU race: the predicate (`containsKey`) and the action (`get`) execute against two independent reads of a reload-replaceable field.

Failure mode in production:

1. Reader: `containsKey("alice")` → snapshot A says yes.
2. Scheduler: reload swaps in snapshot B (alice no longer present, e.g. just deleted).
3. Reader: `get("alice")` → returns `null`.
4. Caller now observes a `null` user/role/permission list immediately after the cache reported the key as present, leading either to NPEs in downstream code paths or to spurious "not found" results on resources that very much do exist.

The race is independent of, and not closed by, declaring the underlying field `volatile` — that only fixes visibility of each individual read, not atomicity across two reads.

## Fix

Read the map once into a local, then inspect the value with a null check before deciding whether to reload. `ConcurrentHashMap` forbids `null` values, so `containsKey == true` and `get != null` are equivalent — behaviour is preserved exactly. The reload-then-retry fallback at the end of each method is untouched.

```java
User cached = getCachedUserMap().get(username);
if (cached != null) {
    return cached;
}
reload();
return getCachedUserMap().get(username);
```

Three call sites, two files. No API or wire-format change.

## Tests Added

New per-site tests using a `CountingMap` extension of `ConcurrentHashMap` that increments separate counters in its overridden `get` and `containsKey` methods. After the fix each test observes exactly one `get` call and zero `containsKey` calls on a cache hit; before the fix the tests fail with `expected: <1> but was: <2>` because `ConcurrentHashMap.containsKey` is internally implemented as `get(key) != null` and the explicit `get` adds a second invocation.

| Change point | Test |
|--------------|------|
| `NacosUserServiceRemoteImpl.getUser` reads cached map once on hit | `NacosUserServiceRemoteImplTest#testGetUserReadsCachedMapOnceOnHit` |
| `NacosRoleServiceRemoteImpl.getPermissions(String)` reads cached map once on hit | `NacosRoleServiceRemoteImplTest#testGetPermissionsReadsCachedMapOnceOnHit` |
| `NacosRoleServiceRemoteImpl.getRoles(String)` reads cached map once on hit | `NacosRoleServiceRemoteImplTest#testGetRolesReadsCachedMapOnceOnHit` |

Verified by reverting the production change and re-running: all three new tests fail with the original code (`expected: <1> but was: <2>`) and pass once the fix is reapplied. `mvn -pl plugin-default-impl/nacos-default-auth-plugin test` reports 162 tests passing; checkstyle is clean.

## Impact

- Scope: 9 added / 6 removed lines across two production files plus two dedicated test classes.
- Affects: any deployment using the remote variants of the auth plugin (`NacosUserServiceRemoteImpl`, `NacosRoleServiceRemoteImpl`), and by extension `AbstractCheckedRoleService.hasPermission` / `hasGlobalAdminRole(...)` which call these methods on every authorization decision.
- Does not change the cache reload schedule, the reload trigger on miss, the result returned to callers in any consistent state, or the equivalent direct-DB implementations (which only access the cache once and are not affected).
- Independent of and complementary to PR #14996 (volatile publication of `userMap`): this PR closes the cross-read atomicity hole, that PR closes the per-read visibility hole. Reverting either fix leaves the other one's behaviour intact.

### DIFF
--- a/plugin-default-impl/nacos-default-auth-plugin/src/main/java/com/alibaba/nacos/plugin/auth/impl/roles/NacosRoleServiceRemoteImpl.java
+++ b/plugin-default-impl/nacos-default-auth-plugin/src/main/java/com/alibaba/nacos/plugin/auth/impl/roles/NacosRoleServiceRemoteImpl.java
@@ -91,8 +91,9 @@ public class NacosRoleServiceRemoteImpl extends AbstractCheckedRoleService imple
     
     @Override
     public List<PermissionInfo> getPermissions(String role) {
-        if (getCachedPermissionInfoMap().containsKey(role)) {
-            return getCachedPermissionInfoMap().get(role);
+        List<PermissionInfo> cached = getCachedPermissionInfoMap().get(role);
+        if (cached != null) {
+            return cached;
         }
         reload();
         return getCachedPermissionInfoMap().get(role);
@@ -114,8 +115,9 @@ public class NacosRoleServiceRemoteImpl extends AbstractCheckedRoleService imple
     
     @Override
     public List<RoleInfo> getRoles(String username) {
-        if (getCachedRoleInfoMap().containsKey(username)) {
-            return getCachedRoleInfoMap().get(username);
+        List<RoleInfo> cached = getCachedRoleInfoMap().get(username);
+        if (cached != null) {
+            return cached;
         }
         reload();
         return getCachedRoleInfoMap().get(username);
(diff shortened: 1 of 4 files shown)


shiyiyue1102:

---

KomachiSion (approved):

PR #15002, opened 2026-05-01
fix(auth): make NacosServerAuthConfig plugin map volatile and read once

daguimu:

`NacosServerAuthConfig.authPluginProperties` (`core/src/main/java/com/alibaba/nacos/core/auth/NacosServerAuthConfig.java`) is reassigned by `refreshPluginProperties` whenever the dynamic-config refresh path runs (triggered via `AbstractDynamicConfig.onEvent` on `ServerConfigChangeEvent`), while request-handling threads call `getAuthPluginProperties` concurrently. Two issues exist on the read path:

1. The field has no `volatile` modifier, so a reassignment performed on the refresh thread is not guaranteed to be visible to other threads (and the new map's writes inside `refreshPluginProperties` only happen-before publication if the field is `volatile`).
2. `getAuthPluginProperties` reads the field twice — once via `containsKey`, then via `get`. If a refresh swaps in a map missing the requested key between the two reads, `get` returns `null` and the method propagates `null` to callers instead of falling back to the empty-`Properties` branch as intended.

This mirrors the same race that was fixed in `ConfigChangeConfigs` in #14988; `AuthConfigs` in the auth plugin module has the same pattern and is being addressed in a separate PR.

## Brief changelog

- `NacosServerAuthConfig.authPluginProperties`: declared `volatile` so refresh-thread reassignments and the contents of the new map are guaranteed-visible to readers.
- `NacosServerAuthConfig.getAuthPluginProperties`: reads the field once, null-checks the lookup, and falls back to a fresh empty `Properties` whenever the key is absent — eliminating the check-then-act window.
- `NacosServerAuthConfigTest`: adds `testGetAuthPluginPropertiesNeverReturnsNullDuringConcurrentRefresh`, which spins up 8 reader threads against a refresher thread that alternates two environments (one with the probe key, one without) and asserts no reader ever observes a `null` result.

## Verifying this change

- Reverting the source change (single-`get` + null check restored to the original `containsKey` + `get` pattern) makes the new test fail with `getAuthPluginProperties returned null mid-refresh` (verified locally).
- After the fix, `mvn -pl core -am test -Dtest=NacosServerAuthConfigTest` and `mvn -pl core -am apache-rat:check checkstyle:check spotbugs:check -DskipTests` both pass.
- All existing tests in `NacosServerAuthConfigTest` continue to pass — the change preserves the documented contract that `getAuthPluginProperties` returns a non-null `Properties` (empty when the type is unknown).

### DIFF
--- a/core/src/main/java/com/alibaba/nacos/core/auth/NacosServerAuthConfig.java
+++ b/core/src/main/java/com/alibaba/nacos/core/auth/NacosServerAuthConfig.java
@@ -59,7 +59,7 @@ public class NacosServerAuthConfig extends AbstractDynamicConfig implements Naco
     
     private String serverIdentityValue;
     
-    private Map<String, Properties> authPluginProperties = new HashMap<>();
+    private volatile Map<String, Properties> authPluginProperties = new HashMap<>();
     
     public NacosServerAuthConfig() {
         super("NacosServerAuth");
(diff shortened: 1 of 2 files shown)


shiyiyue1102:

---

KomachiSion (approved):

PR #15064, opened 2026-05-07
[AI] Harden AgentSpecZipParser against Zip Slip / Zip Bomb / entry-count DoS

LiyunZhang10:
## Bug / Feature Description

`AgentSpecZipParser#unzipToEntries` is missing the three security checks that `SkillZipParser` already enforces for the Skill upload path. The Zip is parsed entirely in-memory, but without these checks a crafted upload can:

1. **Zip Slip / absolute path** — entries named `../../etc/passwd` or `/etc/shadow` are accepted and flow into downstream processing (no `SkillUtils.validatePathSafety(name)` call).
2. **Zip Bomb (memory DoS)** — a small zip (~KB) can expand to GBs because there is no running cap on total decompressed bytes. JVM can OOM and take down the Nacos process.
3. **Entry-count flooding** — a zip with millions of tiny entries exhausts heap/CPU because there is no cap on the number of entries.

For comparison, `SkillZipParser` already does all three checks (`ai/src/main/java/com/alibaba/nacos/ai/utils/SkillZipParser.java:243~278`).

## Fix

Mirror the `SkillZipParser` hardening in `AgentSpecZipParser` — minimal diff,no behavioural change for legitimate uploads:

- Add `SkillUtils.validatePathSafety(name)` on every entry (rejects `..` and absolute paths, throws `SecurityException`).
- New constant `MAX_TOTAL_UNCOMPRESSED_BYTES = 50 MB`; running `totalSize`  accumulator throws `IOException` as soon as the cap is crossed.
- New constant `MAX_ZIP_ENTRIES = 500`; running count throws `IOException` when exceeded. Limits chosen to match `SkillZipParser` exactly.

## Tests

- All 12 existing tests in `AgentSpecZipParserTest` / `AgentSpecZipParserPathsTest`  still pass — no functional regression for legitimate AgentSpec packages.
- New `AgentSpecZipParserSecurityTest` adds 4 cases that precisely reproduce  the three attack classes:
    - `testPathTraversalAttackShouldBeRejected`  (`../../etc/passwd`)
    - `testAbsolutePathEntryShouldBeRejected`    (`/etc/shadow`)
    - `testZipBombShouldBeRejected`              (110 × 1 MB zero-fill entries)
    - `testTooManyEntriesShouldBeRejected`       (600 entries)

### DIFF
--- a/ai/src/main/java/com/alibaba/nacos/ai/utils/AgentSpecZipParser.java
+++ b/ai/src/main/java/com/alibaba/nacos/ai/utils/AgentSpecZipParser.java
@@ -22,6 +22,7 @@ import com.alibaba.nacos.api.ai.model.agentspecs.AgentSpecResource;
 import com.alibaba.nacos.api.ai.model.agentspecs.AgentSpecUtils;
 import com.alibaba.nacos.api.ai.model.skills.SkillUtils;
 import com.alibaba.nacos.api.exception.api.NacosApiException;
+import com.alibaba.nacos.api.exception.runtime.NacosRuntimeException;
 import com.alibaba.nacos.api.model.v2.ErrorCode;
 import com.alibaba.nacos.common.utils.JacksonUtils;
 import com.alibaba.nacos.common.utils.StringUtils;
@@ -168,6 +169,11 @@ public class AgentSpecZipParser {
      *   <li>Enforces maximum number of entries ({@link #MAX_ZIP_ENTRIES})
      *       to prevent entry-count flooding attacks</li>
      * </ul>
+     *
+     * <p>Security-limit violations are reported as {@link NacosRuntimeException} (not {@link IOException})
+     * because they represent invalid user input rather than an underlying I/O failure. The caller
+     * {@link #parseAgentSpecFromZip(byte[], String)} translates them into a {@link NacosApiException}
+     * for the HTTP layer.
      */
     private static List<ZipEntryData> unzipToEntries(byte[] zipBytes) throws IOException {
         List<ZipEntryData> result = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


ai/src/main/java/com/alibaba/nacos/ai/utils/AgentSpecZipParser.java
```diff
@@ -159,12 +181,24 @@ private static List<ZipEntryData> unzipToEntries(byte[] zipBytes) throws IOExcep
                     continue;
                 }
                 String name = entry.getName();
+                // Security: reject path traversal (..) and absolute paths.
+                SkillUtils.validatePathSafety(name);
                 if (name != null && (name.startsWith("__MACOSX/") || name.contains("/__MACOSX/"))) {
                     continue;
                 }
+                if (result.size() >= MAX_ZIP_ENTRIES) {
+                    throw new IOException(
```
KomachiSion: Should we throw NacosRuntimeException？

KomachiSion:
@LiyunZhang10 help create an issue and link to this PR

ai/src/main/java/com/alibaba/nacos/ai/utils/AgentSpecZipParser.java
LiyunZhang10: Done.
However,in `/nacos/ai/src/main/java/com/alibaba/nacos/ai/utils/SkillZipParser.java` line 261、269 ,it throws IOException, should we fix that?

KomachiSion: Sure, you can do this in other PR.

BTW, I think the Max bytes and max entrires of zip should be configued.  By ENV or JVM properties.

LiyunZhang10: `SkillZipParser.java` fix at https://github.com/alibaba/nacos/pull/15069

LiyunZhang10: via JVM properties / env vars  with the current values as defaults.May I do that in a separate follow-up PR to keep this one focused?

KomachiSion (approved):

KomachiSion:
> May I do that in a separate follow-up PR to keep this one focused?

Please do this with a new one PR. Also do this for #15069

Issue #14981, opened 2026-04-24
[Bug] PostgreSQL: Concurrent publishConfig calls lead to UniqueIndexConflict instead of idempotent update


suwei999:
In a Nacos cluster environment using PostgreSQL as the database backend, concurrent calls to the publishConfig API for the same dataId, group, and tenant occasionally result in a 500 Internal Server Error. The root cause is a SQL exception: duplicate key value violates unique constraint "uk_configinfo_datagrouptenant".

Expected behavior
The publishConfig operation should be idempotent. If the configuration already exists, it should be updated gracefully. Under high concurrency or network retries, the server should handle the race condition and ensure that one of the requests succeeds in updating the record rather than throwing a UniqueIndexConflict exception to the client.

Actual behavior
Nacos server returns a 500 error during the race condition. The logs show that PostgreSQL rejects the INSERT statement because a record with the same unique key was committed by another thread just milliseconds prior, and the current thread's "check-then-insert" logic failed to catch it.

How to Reproduce

Set up a Nacos 2.x cluster with PostgreSQL 14+ as the storage backend.

Use a stress testing tool (like JMeter or a custom script) to call the /nacos/v1/cs/configs (POST) API simultaneously with the exact same dataId, group, and namespace.

Observe that a small percentage of requests will fail with a 500 error and a PSQLException in the server logs.

Environment

Nacos Version: 2.3.2 or 2.4.0

OS: Linux / Docker

DB: PostgreSQL 15.x

Deployment: Cluster mode

Possible Solution
For MySQL, Nacos leverages ON DUPLICATE KEY UPDATE. For PostgreSQL, the implementation should be reviewed to ensure the INSERT ... ON CONFLICT (data_id, group_id, tenant_id) DO UPDATE ... syntax is correctly utilized, or the service layer should implement a more robust retry/locking mechanism to handle this specific PG constraint violation.

shiyiyue1102:
Thanks for the detailed report and clear reproduction steps. The root cause is confirmed in the source: `ExternalConfigInfoPersistServiceImpl.insertOrUpdate()` uses a check-then-insert pattern (`findConfigInfoState` → `addConfigInfo`), which is susceptible to a TOCTOU race under concurrent writes. The PostgreSQL plugin (`ConfigInfoMapperByPostgresql`) inherits the base insert SQL without an `ON CONFLICT` clause, so the second concurrent INSERT hits the unique constraint on `(data_id, group_id, tenant_id)`.

[...]

PR #15071, opened 2026-05-07
[ISSUE #14981] fix(config): keep publishConfig idempotent when concurrent inserts race

daguimu:
## Problem

In a Nacos cluster with an external RDBMS, two concurrent `publishConfig` calls for the same `dataId`/`group`/`tenant` can lose a small fraction of requests with HTTP 500 and a unique-constraint violation in the server log. On PostgreSQL the symptom is "duplicate key value violates unique constraint `uk_configinfo_datagrouptenant`" (#14981); MySQL and Derby surface analogous integrity violations under the same window. The user expects `publishConfig` to remain idempotent.

## Root Cause

`ExternalConfigInfoPersistServiceImpl#insertOrUpdate` and `#insertOrUpdateCas` first probe `findConfigInfoState` and then either `addConfigInfo` (INSERT path) or `updateConfigInfo` / `updateConfigInfoCas` (UPDATE path). Two concurrent threads can both observe "no row" from the probe and both run the INSERT; one wins and commits, the other loses with `DataIntegrityViolationException`. The existing surrounding `catch (Exception)` re-throws after logging, so the loser's request returns 500.

```java
ConfigInfoStateWrapper configInfoState = findConfigInfoState(dataId, group, tenant);
if (configInfoState == null) {
    return addConfigInfo(...);   // ← race loser hits unique-constraint violation here
} else {
    return updateConfigInfo(...);
}
```

## Fix

Wrap the INSERT in a localized `try/catch` for `DataIntegrityViolationException` and fall through to the UPDATE path when it fires. The recovery delivers the same outcome the caller would have gotten if the winner's row had been visible at the probe.

`config/src/main/java/com/alibaba/nacos/config/server/service/repository/extrnal/ExternalConfigInfoPersistServiceImpl.java`:

- `insertOrUpdate`: on `DataIntegrityViolationException` from `addConfigInfo`, call `updateConfigInfo`.
- `insertOrUpdateCas`: on `DataIntegrityViolationException` from `addConfigInfo`, call `updateConfigInfoCas`. A stale `casMd5` still surfaces correctly — `updateConfigInfoCas` returns `ConfigOperateResult(success=false)`, and the controller layer turns that into the existing "Cas publish fail, server md5 may have changed" error rather than 500.

The fix is intentionally localized to the persistence layer's check-then-act window. No SQL or schema changes; no controller / service-layer changes.

## Tests Added

`config/src/test/java/com/alibaba/nacos/config/server/service/repository/extrnal/ExternalConfigInfoPersistServiceImplTest.java`:

| Change Point | Test |
|--------------|------|
| `insertOrUpdate` recovers via `updateConfigInfo` after losing the insert race | `testInsertOrUpdateRecoversFromConcurrentInsertRace` — mocks `findConfigInfoState` to return null, then makes the row INSERT throw `DataIntegrityViolationException`. Asserts no exception escapes, the INSERT was attempted exactly once, and the history record was written via the UPDATE path (`"U"` op type). |
| `insertOrUpdateCas` recovers via `updateConfigInfoCas` after losing the insert race | `testInsertOrUpdateCasRecoversFromConcurrentInsertRace` — same race shape on the CAS path, with the recovery UPDATE returning 0 affected rows (CAS check fails because the winner's md5 differs from `casMd5`). Asserts the result reports `success=false` instead of throwing. |

The full `ExternalConfigInfoPersistServiceImplTest` suite (40 tests including the two new ones) passes; the existing `testInsertOrUpdateOfException` still asserts that non-integrity errors (e.g. `CannotGetJdbcConnectionException`) still propagate, so the new catch is properly scoped to integrity violations only.

Verification:

```
mvn -pl config -am test -Dtest=ExternalConfigInfoPersistServiceImplTest -Dsurefire.failIfNoSpecifiedTests=false
# Tests run: 40, Failures: 0, Errors: 0, Skipped: 0
mvn -pl config -am -B compile apache-rat:check checkstyle:check spotbugs:check -DskipTests
# 0 Checkstyle violations, 0 SpotBugs findings
```

## Impact

- Behavior: concurrent `publishConfig` calls for the same key on PostgreSQL / MySQL / Derby no longer leak a 500 to the client when they lose the insert race; the loser silently recovers via UPDATE, preserving the documented idempotency of the publish API.
- Scope: limited to the External (cluster + external RDBMS) impl. The Embedded impl serializes writes through the Raft leader, so the same window does not occur there in practice and is left unchanged per the one-fix-per-PR policy.
- API: no public-API changes. The new `catch` is internal to the persistence service.
- Performance: no change in the success path. Only the rare race-loser case incurs the cost of a second statement (the UPDATE) instead of the previous behavior of throwing.
- Risk: bounded — `DataIntegrityViolationException` is Spring's narrow wrapper for unique-constraint violations from JDBC drivers; non-integrity failures (connection drops, etc.) still propagate as before.

### DIFF
--- a/config/src/main/java/com/alibaba/nacos/config/server/service/repository/extrnal/ExternalConfigInfoPersistServiceImpl.java
+++ b/config/src/main/java/com/alibaba/nacos/config/server/service/repository/extrnal/ExternalConfigInfoPersistServiceImpl.java
@@ -222,7 +222,14 @@ public class ExternalConfigInfoPersistServiceImpl implements ConfigInfoPersistSe
             ConfigInfoStateWrapper configInfoState = findConfigInfoState(configInfo.getDataId(), configInfo.getGroup(),
                     configInfo.getTenant());
             if (configInfoState == null) {
-                return addConfigInfo(srcIp, srcUser, configInfo, configAdvanceInfo);
+                try {
+                    return addConfigInfo(srcIp, srcUser, configInfo, configAdvanceInfo);
+                } catch (DataIntegrityViolationException raceWithConcurrentInsert) {
+                    // Another thread inserted the same dataId/group/tenant in the gap between the
+                    // findConfigInfoState above and our insert. Treat this as the update path and
+                    // keep publishConfig idempotent instead of surfacing a 500 to the caller.
+                    return updateConfigInfo(configInfo, srcIp, srcUser, configAdvanceInfo);
+                }
             } else {
                 return updateConfigInfo(configInfo, srcIp, srcUser, configAdvanceInfo);
             }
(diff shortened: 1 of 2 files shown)


KomachiSion:
这不是又回滚回老版本的问题了吗。

老版本就是因为不同的数据库抛出的异常各有不同，导致add的try catch可能catch不到冲突，从而导致更新失败。

后来修改成先查再改，即使此时出现了并发问题，更新失败了。 用户会收到错误提示，从而决定是重试还是刷新更新之后在 重试。也能避免一定的ABA问题。

PR #15186, opened 2026-05-19
fix(naming): close addPublisherIndexes race that re-emits ADD_SERVICE

daguimu:

Follow-up to #15182 (which I called out in that PR's description). The publisher-index registration path in `ClientServiceIndexesManager#addPublisherIndexes` decides the event type from a non-atomic `containsKey` check and publishes the event *before* the index entry is written:

```java
String serviceChangedType = INSTANCE_CHANGED;
if (!publisherIndexes.containsKey(service)) {
    serviceChangedType = ADD_SERVICE;
}
NotifyCenter.publishEvent(new ServiceChangedEvent(service, serviceChangedType, true));
publisherIndexes.computeIfAbsent(service, key -> new ConcurrentHashSet<>()).add(clientId);
```

Two concurrent first-time registrations of the same service can each see `containsKey` return `false` and each publish a redundant `ADD_SERVICE` event before the eventual `computeIfAbsent`. Downstream subscribers (fuzzy-watch notify, push planner, metrics) then react twice to what should be a single first-creation event. The ordering is also wrong on its own — a subscriber that queries the map on `ADD_SERVICE` is racing the writer.

## Brief changelog

`ClientServiceIndexesManager#addPublisherIndexes` now uses `computeIfAbsent`'s mapping function as the atomic first-insert signal, and publishes the event only after the entry is in the index:

```java
boolean[] firstInsert = {false};
publisherIndexes.computeIfAbsent(service, key -> {
    firstInsert[0] = true;
    return new ConcurrentHashSet<>();
}).add(clientId);
String serviceChangedType = firstInsert[0] ? ADD_SERVICE : INSTANCE_CHANGED;
NotifyCenter.publishEvent(new ServiceChangedEvent(service, serviceChangedType, true));
```

`ConcurrentHashMap`'s contract guarantees the mapping function runs at most once per absent key, so exactly one caller observes `firstInsert[0] == true`. The exact same `computeIfAbsent` atomicity is already relied on directly below in `removePublisherIndexes` (via `computeIfPresent`), so the idiom is consistent with the surrounding code.

A code comment above the call site spells out the two invariants the change restores (atomic first-insert detection + index-then-event ordering) so future readers don't accidentally re-order the two calls.

## Verifying this change

```bash
mvn -pl naming -am install -DskipTests
mvn -pl naming test -Dtest=ClientServiceIndexesManagerTest
# Tests run: 11, Failures: 0, Errors: 0, Skipped: 0
mvn -pl naming -B spotless:check checkstyle:check spotbugs:check -DskipTests
# 0 spotless / 0 checkstyle / 0 spotbugs violations
```

New test `testAddPublisherIndexesEmitsAddServiceOnFirstAndInstanceChangedOnRepeat` asserts the contract on a fresh `ClientServiceIndexesManager` (the shared `setUp()` pre-seeds the index, which would mask first-time behavior). It uses `MockedStatic<NotifyCenter>` to count events and verifies:

- first call → exactly one `ADD_SERVICE` event,
- the index entry exists before the event is published (the new ordering),
- subsequent calls → only `INSTANCE_CHANGED`, never another `ADD_SERVICE`,
- all clients end up in the index.

## Notes for reviewers

I deliberately did **not** add a multi-threaded test that reproduces the race itself. `MockedStatic` is thread-confined to the thread that created it, so a real concurrent test would have to stand up `NotifyCenter`'s actual publisher pipeline plus a synchronous bridge to count events — disproportionate test infrastructure for asserting an atomicity property the JDK already guarantees. The race argument relies entirely on `ConcurrentHashMap#computeIfAbsent` running its mapping function at most once for an absent key, which is the same guarantee the file already uses for the symmetric `removePublisherIndexes` flow.

This is the follow-up I called out in #15182:

> `ClientServiceIndexesManager#addPublisherIndexes` (line 136-145) carries a different race — concurrent first-time registrations of the same service can each see `containsKey` return `false` and each emit a redundant `ADD_SERVICE` event before the eventual `computeIfAbsent`. Fixing that requires reordering the `publishEvent` to follow the `computeIfAbsent`, which is a behavioral change worth its own PR and its own concurrent test.

* [x] Make sure there is a Github issue filed for the change (usually before you start working on it). Trivial changes like typos do not require a Github issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue. *(Direct follow-up to #14751 / #15182; same module, same anti-pattern family, same contributor.)*
* [x] Run `mvn -B clean package apache-rat:check spotbugs:check -DskipTests` to make sure basic checks pass. Run `mvn clean install` to make sure unit-test pass. Run `mvn clean test-compile failsafe:integration-test`  to make sure integration-test pass.

### DIFF
--- a/naming/src/main/java/com/alibaba/nacos/naming/core/v2/index/ClientServiceIndexesManager.java
+++ b/naming/src/main/java/com/alibaba/nacos/naming/core/v2/index/ClientServiceIndexesManager.java
@@ -134,14 +134,21 @@ public class ClientServiceIndexesManager extends SmartSubscriber {
     }
     
     private void addPublisherIndexes(Service service, String clientId) {
-        String serviceChangedType = Constants.ServiceChangedType.INSTANCE_CHANGED;
-        if (!publisherIndexes.containsKey(service)) {
-            // The only time the index needs to be updated is when the service is first created
-            serviceChangedType = Constants.ServiceChangedType.ADD_SERVICE;
-        }
-        NotifyCenter
-            .publishEvent(new ServiceEvent.ServiceChangedEvent(service, serviceChangedType, true));
-        publisherIndexes.computeIfAbsent(service, key -> new ConcurrentHashSet<>()).add(clientId);
+        // Detect the first registration atomically: only the thread whose mapping function actually
+        // ran on ConcurrentHashMap#computeIfAbsent sets firstInsert[0] = true, so concurrent
+        // first-time registrations of the same service no longer each see containsKey == false and
+        // each emit an ADD_SERVICE event. The event is also published only after the index entry
+        // exists, so subscribers that re-query the map on ADD_SERVICE no longer race the writer.
+        boolean[] firstInsert = {false};
+        publisherIndexes.computeIfAbsent(service, key -> {
+            firstInsert[0] = true;
+            return new ConcurrentHashSet<>();
+        }).add(clientId);
+        String serviceChangedType = firstInsert[0]
+            ? Constants.ServiceChangedType.ADD_SERVICE
+            : Constants.ServiceChangedType.INSTANCE_CHANGED;
+        NotifyCenter.publishEvent(
+            new ServiceEvent.ServiceChangedEvent(service, serviceChangedType, true));
     }
     
     private void removePublisherIndexes(Service service, String clientId) {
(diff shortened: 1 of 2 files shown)


KomachiSion (changes requested):
没有必要， computeIfAbsent 能够保证数据的一致性， 事件仅是一个触发器，不影响实际的数据同步。

PR #15344, opened 2026-06-09
Add SQL Server datasource support

think-gem:

Add SQL Server datasource support

## Brief changelog

Add SQL Server datasource support

## Verifying this change

- `mvn clean install`
- `mvn clean test`
- connect sql server ok


KomachiSion:
Please move to nacos-group/nacos-plugin first.

https://github.com/nacos-group/nacos-plugin/tree/develop/nacos-datasource-plugin-ext

think-gem:
> Please move to nacos-group/nacos-plugin first.
> 
> https://github.com/nacos-group/nacos-plugin/tree/develop/nacos-datasource-plugin-ext

Suggestion: Have you considered including this in the "default implementation of the plugin"? SQL Server has a large user base, but the version in nacos-group cannot be updated in a timely manner, which often leads to various issues. If it were incorporated, the AI model would synchronize updates and modifications, facilitating unified management.

PR #15406, opened 2026-06-21
[ISSUE #15359] Fix AI registry version deletion

### LINKED ISSUE #15359: [Bug] AI registry: deleting a version corrupts the version aggregate — A2A query NPE (500) / MCP server & endpoints disappear
Two related defects in AI registry version lifecycle handling (both in the config-aggregate based implementations of A2A and MCP). They likely explain real-world reports such as #15353 (MCP http endpoints disappearing at runtime → 503).

Bug 1 — A2A: querying an agent by version throws NPE (500) after the latest version is deleted
Reproduce

Publish an A2A agent with two versions v1, then v2 (so v2 is latest).
Delete the latest v2 while v1 remains: DELETE /v3/admin/ai/a2a?agentName=<name>&version=v2.
Query a remaining version: GET /v3/admin/ai/a2a?agentName=<name>&version=v1 → HTTP 500.
Root cause — In A2aServerOperationService, deleting the latest version while others remain sets latestPublishedVersion=null (and version=null) and persists, without electing a new latest. Then queryTargetVersion(...) does agentCardVersionInfo.getLatestPublishedVersion().equals(result.getVersion()) → null.equals(...) → NPE. Reachable via admin A2aAdminController#getAgentCard, the console, and gRPC QueryAgentCardRequestHandler.

Bug 2 — MCP: deleting a single version makes the whole server (and endpoints) disappear
Reproduce

Create an MCP server with two versions.
Delete one version: DELETE /v3/admin/ai/mcp?...&version=<v>.
The whole server then disappears from list/detail (MCP_SERVER_NOT_FOUND), even though another version still exists; the remaining version's spec config is orphaned.
Root cause — In McpServerOperationService#deleteMcpServer, inside the per-version loop, it deletes the entire version-aggregate config (formatServerVersionInfoDataId(mcpServerId) — no version), instead of removing just the deleted version from the aggregate and updating it (the way A2A updates its aggregate). After that, getMcpServerVersionInfo returns NOT_FOUND. This matches the symptom in #15353 (endpoints/server vanishing after a version-level operation), though that report could also involve ephemeral endpoint instances lost on restart — worth confirming the operation history there.

Proposed fix (need maintainers' direction)
A2A: on deletion, re-elect a new latest from the remaining versions (so latestPublishedVersion is never null while versions exist), and/or null-guard the comparison in queryTargetVersion.
MCP: when deleting a single version, remove that version from the aggregate and update it (like A2A), instead of deleting the whole aggregate.
Happy to submit a PR with regression tests once the direction is confirmed.

Environment: develop / 3.2.x, AI registry (A2A & MCP).

hutiefang76:
## Summary
- re-elect a remaining A2A agent version when deleting the latest version
- keep the MCP server version aggregate when deleting only one version, and publish the remaining version list
- delete the MCP version aggregate only when all versions are removed

## Root cause
Deleting a target version updated the per-version config but left aggregate metadata inconsistent. For A2A, deleting the latest version could leave latest/version empty while older versions still existed. For MCP, deleting one version removed the whole version aggregate, so the server and endpoints disappeared.

## Tests
- JAVA_HOME=$(/usr/libexec/java_home -v 17) ./mvnw -s .codex-empty-settings.xml -pl ai -am -Dtest=A2aServerOperationServiceTest#testDeleteAgentLatestVersionElectsRemainingVersion,McpServerOperationServiceTest#deleteMcpServerForTargetVersion -Dsurefire.failIfNoSpecifiedTests=false test
- JAVA_HOME=$(/usr/libexec/java_home -v 17) ./mvnw -s .codex-empty-settings.xml -pl ai -am -Dtest=A2aServerOperationServiceTest,McpServerOperationServiceTest -Dsurefire.failIfNoSpecifiedTests=false test
- JAVA_HOME=$(/usr/libexec/java_home -v 17) ./mvnw -s .codex-empty-settings.xml -pl ai -am -Dsurefire.failIfNoSpecifiedTests=false test

Fixes #15359

### DIFF
--- a/ai/src/main/java/com/alibaba/nacos/ai/service/McpServerOperationService.java
+++ b/ai/src/main/java/com/alibaba/nacos/ai/service/McpServerOperationService.java
@@ -624,7 +624,7 @@ public class McpServerOperationService {
                 namespaceId, null, null,
                 "nacos", null);
         }
-
+        
         String serverVersionDataId = McpConfigUtils.formatServerVersionInfoDataId(mcpServerId);
         if (deleteAllVersions) {
             configOperationService.deleteConfig(serverVersionDataId,
@@ -647,7 +647,7 @@ public class McpServerOperationService {
                 : AiResourceTraceService.OP_DELETE_RESOURCE,
             VisibilityHelper.resolveCurrentIdentity(), VisibilityHelper.resolveClientIp());
     }
-
+    
     private void electLatestMcpServerVersion(McpServerVersionInfo mcpServerVersionInfo) {
         List<ServerVersionDetail> versionDetails = mcpServerVersionInfo.getVersionDetails();
         if (CollectionUtils.isEmpty(versionDetails)) {
--- a/ai/src/main/java/com/alibaba/nacos/ai/service/McpServerOperationService.java
+++ b/ai/src/main/java/com/alibaba/nacos/ai/service/McpServerOperationService.java
@@ -609,6 +609,9 @@ public class McpServerOperationService {
                 .map(ServerVersionDetail::getVersion)
                 .collect(Collectors.toList());
         }
+        List<String> deletedVersions = new ArrayList<>(versionsNeedDelete);
+        boolean deleteAllVersions = mcpServerVersionInfo.getVersionDetails().stream()
+            .map(ServerVersionDetail::getVersion).allMatch(deletedVersions::contains);
         
         for (String versionNeedDelete : versionsNeedDelete) {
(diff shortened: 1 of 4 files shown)


KomachiSion:
I re-checked the linked issue and the patch. The problem is valid and the fix direction looks correct to me.

The current blocker is CI formatting only: `spotless:check` is failing on the changed files. Please run `mvn spotless:apply` and push the formatting update so CI can continue. If this is not fixed soon, we may need to let someone else take over the CI cleanup.

KomachiSion (approved):

PR #15509, opened 2026-07-15
[ISSUE #15468] Route duplicate-key classification through the datasource dialect SPI

### LINKED ISSUE #15468: Add datasource plugin support for duplicate-key exception classification
Background:
#14981 reports that concurrent publishConfig calls with PostgreSQL external storage may hit `uk_configinfo_datagrouptenant` unique-key conflicts. #15465 handles the safe Spring-standard `DuplicateKeyException` path and distinguishes non-CAS `insertOrUpdate` upsert semantics from `insertOrUpdateCas` conflict semantics. Related previous PRs: #15272, #15278.

Problem:
Spring's DAO exception hierarchy distinguishes `DuplicateKeyException` from the broader `DataIntegrityViolationException`. However, different databases, JDBC drivers, and Spring exception translators may expose unique-key conflicts differently. Some datasource plugins may get a precise `DuplicateKeyException`, while others may only see a `DataIntegrityViolationException` or a lower-level `SQLException` with vendor-specific SQLState/errorCode/message.

The common config persistence layer should not parse database-specific error codes or messages. That would leak datasource dialect knowledge into generic config logic and may accidentally swallow non-duplicate integrity failures.

Proposal:
Add a datasource-plugin-level capability for duplicate-key classification, for example:

- `boolean isDuplicateKey(Throwable throwable)`
- or an equivalent exception translator/handler interface in the datasource plugin SPI

The default implementation can keep the current safe behavior by detecting Spring `DuplicateKeyException`. Specific datasource plugins such as PostgreSQL, MySQL, Derby, Oracle, and other external datasource plugins can then implement their own classification based on driver behavior, SQLState, vendor error code, or dialect-specific exception structure.

Expected usage:
- Non-CAS `insertOrUpdate`: duplicate-key conflict during insert can fall back to update.
- CAS `insertOrUpdateCas`: duplicate-key conflict during insert should return `ConfigOperateResult(false)` and preserve CAS conflict semantics.
- Non-duplicate integrity failures must continue to be thrown.

Target:
This is a datasource plugin SPI/behavior extension and should be planned for at least 3.3 rather than a small patch release.

wushiyuanmaimob:

Resolves #15468. Make the datasource dialect SPI the single entry point for classifying duplicate unique-key conflicts, following the direction agreed with @KomachiSion in the issue. This removes the hardcoded Spring `DuplicateKeyException` translation from the config persistence layer and gives vendor dialects a clean place to add driver-specific detection (for example a PostgreSQL `SQLState 23505` that Spring's generic translation cannot classify).

## Brief changelog

- `DatabaseDialect#isDuplicateKeyException(Throwable)` default now walks the throwable cause chain and recognizes Spring's `DuplicateKeyException`, matched by class name so the datasource plugin modules keep their Spring-free dependency footprint. A raw vendor `SQLState` such as `23505` is deliberately **not** treated as a duplicate on its own, preserving the `#15465` rethrow contract.
- `ExternalConfigInfoPersistServiceImpl` now delegates all duplicate-key judgement to the active dialect, and only falls back to the inline Spring `DuplicateKeyException` check when no dialect can be resolved (e.g. before datasource plugins are loaded).
- Vendor dialects (PostgreSQL/MySQL/Derby/Oracle) may override the SPI default to additionally inspect the original driver exception, typically combining their check with `DatabaseDialect.super.isDuplicateKeyException(throwable)`.
- Updated the datasource dialect plugin spec (`specs/en` + `specs/zh-cn`).

## Verifying this change

- `AbstractDatabaseDialectTest` (plugin-base, Spring-free): the default rejects `RuntimeException` and a raw `SQLState 23505`, wrapped or not.
- `ExternalConfigInfoPersistServiceImplTest` (config, 93 tests): the dialect default recognizes a wrapped Spring `DuplicateKeyException`; the persist service delegates to the dialect; and the `#15465` rethrow contract for a `BadSqlGrammarException` wrapping `SQLState 23505` is preserved.
- Local checks all pass on the affected modules: `spotless:check`, `checkstyle:check`, `spotbugs:check`, `apache-rat:check`.

* [x] Make sure there is a Github issue filed for the change (#15468).
* [x] Format the pull request title like `[ISSUE #15468] ...`.
* [x] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist.
* [x] Run `mvn -B clean package apache-rat:check spotbugs:check -DskipTests` / module checks to make sure basic checks pass.

### DIFF
--- a/config/src/main/java/com/alibaba/nacos/config/server/service/repository/extrnal/ExternalConfigInfoPersistServiceImpl.java
+++ b/config/src/main/java/com/alibaba/nacos/config/server/service/repository/extrnal/ExternalConfigInfoPersistServiceImpl.java
@@ -49,6 +49,8 @@ import com.alibaba.nacos.plugin.datasource.constants.CommonConstant;
 import com.alibaba.nacos.plugin.datasource.constants.ContextConstant;
 import com.alibaba.nacos.plugin.datasource.constants.FieldConstant;
 import com.alibaba.nacos.plugin.datasource.constants.TableConstant;
+import com.alibaba.nacos.plugin.datasource.dialect.DatabaseDialect;
+import com.alibaba.nacos.plugin.datasource.manager.DatabaseDialectManager;
 import com.alibaba.nacos.plugin.datasource.mapper.ConfigInfoMapper;
 import com.alibaba.nacos.plugin.datasource.mapper.ConfigTagsRelationMapper;
 import com.alibaba.nacos.plugin.datasource.model.MapperContext;
@@ -310,6 +312,13 @@ public class ExternalConfigInfoPersistServiceImpl implements ConfigInfoPersistSe
     }
     
     private boolean isDuplicateKeyException(Throwable exception) {
+        DatabaseDialect dialect = resolveDatabaseDialect();
+        if (dialect != null) {
+            return dialect.isDuplicateKeyException(exception);
+        }
+        // Fallback when no dialect can be resolved (for example before datasource plugins are
+        // loaded): keep the database-agnostic Spring DuplicateKeyException classification, which is
+        // exactly what the active dialect applies as its default behavior.
         Throwable cause = exception;
         while (cause != null) {
             if (cause instanceof DuplicateKeyException) {
@@ -318,6 +327,28 @@ public class ExternalConfigInfoPersistServiceImpl implements ConfigInfoPersistSe
             cause = cause.getCause();
         }
         return false;
+    }
+    
+    /**
+     * Resolve the active datasource dialect for duplicate-key classification.
+     *
+     * <p>All duplicate-key judgement is delegated to {@link DatabaseDialect#isDuplicateKeyException},
+     * whose default already recognizes Spring {@link DuplicateKeyException} and which vendor dialects
+     * may extend with driver-specific detection. Returns {@code null} when no dialect can be resolved
+     * (for example before datasource plugins are loaded), so the caller falls back to the
+     * database-agnostic Spring classification.
+     *
+     * @return the active dialect, or {@code null} when it cannot be resolved
+     */
+    DatabaseDialect resolveDatabaseDialect() {
+        try {
+            return DatabaseDialectManager.getInstance()
+                .getDialect(dataSourceService.getDataSourceType());
+        } catch (IllegalStateException stateException) {
+            LogUtil.DEFAULT_LOG.warn("[duplicate-key] cannot resolve datasource dialect, fallback "
+                + "to spring-standard classification, msg: {}", stateException.getMessage());
+            return null;
+        }
     }
     
     @Override
--- a/config/src/test/java/com/alibaba/nacos/config/server/service/repository/extrnal/ExternalConfigInfoPersistServiceImplTest.java
+++ b/config/src/test/java/com/alibaba/nacos/config/server/service/repository/extrnal/ExternalConfigInfoPersistServiceImplTest.java
@@ -36,6 +36,8 @@ import com.alibaba.nacos.persistence.datasource.DynamicDataSource;
 import com.alibaba.nacos.api.model.Page;
 import com.alibaba.nacos.plugin.datasource.MapperManager;
 import com.alibaba.nacos.plugin.datasource.constants.TableConstant;
+import com.alibaba.nacos.plugin.datasource.dialect.DatabaseDialect;
+import com.alibaba.nacos.plugin.datasource.manager.DatabaseDialectManager;
 import com.alibaba.nacos.plugin.datasource.mapper.ConfigInfoMapper;
 import com.alibaba.nacos.sys.env.EnvUtil;
 import org.junit.jupiter.api.AfterEach;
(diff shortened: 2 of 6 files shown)


KomachiSion:
Hi @wushiyuanmaimob , would you mind provider your dingtalk number, I want to invite you into nacos contributor group.

KomachiSion (approved):

KomachiSion:
@wushiyuanmaimob should we add one more PR to override impl to make pg plugin can handle duplicateKey case?

wushiyuanmaimob:
Thanks again for merging, @KomachiSion  I'd be happy to join the contributor group.
My DingTalk ID is: wushiyuanwork

wushiyuanmaimob:
Yes, happy to take that follow-up. Before writing it I checked what the PG plugin actually does on the current `develop` baseline, and the result is a bit counter-intuitive, so I'd like to align on scope first.

**On the current baseline, PG already works through the SPI default.** With Spring Boot 4.0.x (Spring Framework 7), a real pgjdbc unique violation is already translated into Spring's `DuplicateKeyException`, so `DatabaseDialect#isDuplicateKeyException` returns `true` for PostgreSQL with no override. Running an actual `PSQLException(PSQLState.UNIQUE_VIOLATION)` through the default translator:

```
pgjdbc exception class : org.postgresql.util.PSQLException
SQLState               : 23505
Spring translates to   : org.springframework.dao.DuplicateKeyException
SPI default says dup?  : true
```

The chain that makes this work:

1. Nacos ships no `sql-error-codes.xml`, so `JdbcAccessor` installs `SQLExceptionSubclassTranslator`.
2. `PSQLException` extends `java.sql.SQLException` directly (not `SQLIntegrityConstraintViolationException`), so the JDBC-4 subclass match misses.
3. It falls back to `SQLStateSQLExceptionTranslator`, which since Spring 6 maps SQLState `23505` -> `DuplicateKeyException`.

So a PG override that only looks for `23505` would be dead code today - it can never change the verdict.

**However, the behaviour is Spring-version dependent**, which is the argument for still doing something. The same input on Spring 5.3 translates to `DataIntegrityViolationException`, i.e. not a duplicate. PG correctness currently rests on an implicit property of Spring's translation table rather than on anything the PG plugin itself asserts.

**One constraint on any override:** a naive "walk the cause chain for `23505`" implementation would conflict with the #15465 rethrow contract we pinned in this PR:

```java
assertFalse(dialect.isDuplicateKeyException(new BadSqlGrammarException("insert", "sql",
    new SQLException("duplicate key value violates unique constraint", "23505"))));
```

Chain-walking the SQLState would flip that to `true` and re-swallow the errors #15465 deliberately rethrows.

**Two options - happy to go either way:**

- **Option A (my preference, smaller):** no override; add a regression test pinning the real `PSQLException(23505)` -> `DuplicateKeyException` -> duplicate behaviour on the current baseline. If a Spring upgrade or a user-supplied `sql-error-codes.xml` ever changes the translation, CI catches it, and the override becomes a one-liner at that point.
- **Option B:** add the PG override as defence-in-depth, scoped to the *untranslated* driver exception only (a top-level `PSQLException` with SQLState `23505`), combined with `DatabaseDialect.super.isDuplicateKeyException(...)` so the `BadSqlGrammarException` contract stays intact. The PG module already has the `postgresql` driver on its classpath, so this needs no new dependency.

Which would you prefer? I can open it against #15468 or a new issue if you'd rather track it separately.

KomachiSion:
So in develop branch, which already upgrade to spring boot 4.0, the PG jdbc also can throw DuplicateKeyException and current code can handle it right?

wushiyuanmaimob:
Correct, with one clarification on the wording: pgjdbc itself never throws `DuplicateKeyException`. It throws a `PSQLException` carrying SQLState `23505`, and Spring's exception translation is what turns that into `DuplicateKeyException`. The full path on `develop` today:

1. pgjdbc throws `PSQLException`, SQLState `23505`.
2. `JdbcTemplate` translates it -> `DuplicateKeyException` (via `SQLExceptionSubclassTranslator` -> `SQLStateSQLExceptionTranslator`).
3. The `DatabaseDialect` default finds `DuplicateKeyException` in the cause chain -> returns `true`.
4. `ExternalConfigInfoPersistServiceImpl` acts on it: `insertOrUpdate` falls back to update, `insertOrUpdateCas` returns `ConfigOperateResult(false)`.

So yes: on `develop` with Boot 4.0, a concurrent-publish duplicate key on PostgreSQL is classified and handled correctly today, and the PG plugin needs no override.

The reason I'd still add the regression test is step 2. It is the only step PostgreSQL relies on that neither the PG plugin nor Nacos asserts anywhere - it is a property of Spring's SQLState table. That table has changed before, which is why the same input on Spring 5.3 yields a plain `DataIntegrityViolationException` rather than a duplicate. The test pins steps 1-3 end to end with a real `PSQLException`, so if a future Spring upgrade or a user-supplied `sql-error-codes.xml` changes the translation, CI fails there instead of PostgreSQL silently regressing to rethrowing on concurrent publish (#14981).

I have that test ready (test-only, no production change, no new dependency - the PG driver is already on the `config` test classpath). Shall I open it against #15468, or would you prefer a new issue?

KomachiSion:
Great, we can full close the #14981

Issue #11306, opened 2023-10-30
The latest version of the table `permissions` creates an error and the index is too long.


andotorg:
**Describe the bug**
The latest version of the table `permissions` creates an error and the index is too long.

**Expected behavior**
Create table correctly

**Actually behavior**
creates an error 

**How to Reproduce**
1. exec distribution/conf/mysql-schema.sql
2. show log `1071 - Specified key was too long; max key length is 767 bytes, Time: 0.008000s`

**Desktop (please complete the following information):**
 - OS: MacOS 10.15
 - Version nacos-server  2.2.3
 - Module  config

PR #15513, opened 2026-07-15
[ISSUE #15476] Add plugin-owned visibility grant API

### LINKED ISSUE #15476: [3.3] Add plugin-owned AI resource visibility authorization APIs
## Motivation

Nacos AI resources already have `owner` and `scope` metadata, and the visibility plugin contract is used by AI resource read/write paths. The current default visibility behavior is useful but still incomplete for team collaboration:

- global administrators can read/write all AI resources;
- the resource owner can read/write the resource;
- `PUBLIC` resources are readable by non-owners;
- `PRIVATE` resources are not visible to other users unless an explicit permission check succeeds.

The code and specs already reserve an explicit authorization model:

- `DefaultAiVisibilityService` checks the explicit resource permission `@@visibility/{namespaceId}/{resourceType}/{resourceName}` for single-resource access.
- `VisibilityService.adviseQuery(...)` returns `AuthorizedResources` so list/search queries can include explicitly authorized resources.
- `DefaultVisibilityAdvisorConverter` already has the hook to include `authorizedResourceNames`, but the default visibility service currently leaves the authorized resource list empty.
- The visibility plugin spec says visible resources may be owner-only, public-readable, or explicitly authorized, and that list/search APIs must apply visibility before count and pagination.
- The default auth plugin spec documents the visibility permission resource format and says the default implementation should support explicit visibility permission.

What is missing is a plugin-owned management API and implementation for granting a specific user read/write permission to a specific AI resource.

This should be completed in 3.3.

## Problem

Today, the default AI resource visibility model effectively behaves as owner/admin/public-read for most user workflows. There is no clear API for a resource owner or administrator to grant another user read or write permission on one AI resource.

This blocks common collaboration scenarios, for example:

- user `alice` owns a private `skill` and wants user `bob` to read it;
- user `alice` owns a private `agentspec` and wants user `bob` to update drafts or publish versions;
- a team wants to keep a resource private while granting a small set of users access;
- list/search APIs should show resources that the caller is explicitly authorized to read, not only public resources and resources owned by the caller.

A key design constraint: the authorization API should belong to the plugin implementation, not to the generic AI resource API.

Generic AI resource APIs should continue to manage resource metadata, versions, lifecycle, scope, labels, and content. They should not know how a specific visibility/auth plugin stores grants, roles, users, or external authorization state.

### 1. Add plugin-owned AI resource visibility authorization APIs

Add v3 APIs owned by the default auth/visibility plugin implementation. The exact path can be finalized in design, but it should live under the plugin API surface such as `/v3/auth/...`, not under `/v3/admin/ai/*`, `/v3/console/ai/*`, or generic AI resource controllers.

The API should support at least:

- grant read permission for a user on one AI resource;
- grant write permission for a user on one AI resource;
- revoke read/write permission;
- list current grants for one AI resource;
- optionally list AI resources explicitly granted to the current user or a target user, if needed by the console.

The model should include:

- `namespaceId`;
- `resourceType`, such as `mcp`, `prompt`, `skill`, or `agentspec`;
- `resourceName`;
- grantee identity, initially user-based;
- action: `r`, `w`, or a compatible `rw` representation;
- operator and audit-friendly metadata if available.

### 2. Keep the API boundary plugin-specific

The default `nacos` auth/visibility plugin can implement these APIs using its RBAC/permission storage, for example by writing permissions with resource:

```text
@@visibility/{namespaceId}/{resourceType}/{resourceName}
```

Other visibility/auth plugins should be free to implement their own API or report that grant management is unsupported. The generic AI resource service should only call the `VisibilityService` SPI for validation and query advice.

### 3. Complete runtime authorization behavior

After a grant is created:

- single-resource read should allow a user with explicit `r` permission;
- single-resource write should allow a user with explicit `w` permission;
- list/search/count should include explicitly authorized resources through `AuthorizedResources`;
- explicit read grants should not imply write permission;
- write permission should be at least as strict as read permission, as required by the visibility plugin spec;
- read-denied single-resource access should continue to hide existence when the API contract requires not found;
- write-denied operations should return access denied.

### 4. Define who can manage grants

The first-stage 3.3 behavior should be explicit and conservative. Suggested rule:

- global administrators can manage grants for any AI resource;
- the resource owner can manage grants for that resource;
- users with explicit write permission can modify the resource, but whether they can delegate grants should be a deliberate plugin policy, not an accidental side effect.

The selected rule should be documented in the default auth/visibility plugin spec.

### 5. Console integration

If the console exposes AI resource sharing or authorization management, it should call the plugin-owned API and handle "unsupported by current plugin" gracefully.

The console should not hard-code default plugin storage details into generic AI resource pages.

### 6. Specs and tests

Update the affected specs:

- visibility plugin spec;
- default auth plugin spec;
- AI resource model/lifecycle specs if they need to clarify the boundary between resource metadata and visibility grants;
- v3 API surface spec for the new plugin-owned API family.

Testing should cover:

- grant/revoke API validation and authorization;
- owner/admin grant management rules;
- explicit read and write checks in `DefaultAiVisibilityService`;
- list/search/count including explicitly authorized resources before pagination;
- no leakage of private resources without explicit permission;
- console/API behavior when the selected visibility plugin does not support grant management.

Because this adds HTTP API behavior, the API integration test scenario matrix and coverage registry should also be updated according to the repository API IT rules.

## Non-goals

- This issue should not move grant management into generic AI resource controllers.
- This issue should not require every custom visibility plugin to adopt the default plugin storage model.
- This issue should not make `PUBLIC` resources writable by all users.
- This issue should not replace the existing owner/scope visibility model.

## Acceptance Criteria

- The default plugin provides a clear v3 API to grant/revoke/list user read/write access for one AI resource.
- The API is implemented under the plugin-owned auth/visibility API surface.
- Explicitly granted users can read/write the target AI resource according to the granted action.
- List/search/count APIs include explicitly readable resources correctly.
- Owner/admin grant management rules are documented and tested.
- Generic AI resource APIs remain storage-agnostic and delegate visibility decisions to the visibility plugin.

Zhengcy05:
Special note: Since there are many design changes this time, I sincerely hope someone will carefully review this PR. Thx😘
fixes #15476 

This PR adds explicit visibility grant management to the default auth/visibility plugin so that a resource owner or admin can grant/revoke another user's read/write visibility for a single visibility-managed resource.

The API is exposed under the plugin-owned auth surface:

- `POST /v3/auth/visibility`
- `DELETE /v3/auth/visibility`

Both endpoints are `ADMIN_API`.

## Brief changelog

- Add default visibility grant/revoke APIs under `/v3/auth/visibility`.
- Store grants through the existing RBAC model using one deterministic internal visibility role per grantee user.
- Persist permissions as raw canonical visibility resources:
  `@@visibility/{namespaceId}/{resourceType}/{resourceName}` with action `r` or `rw`.
- Keep visibility generic: the grant model is not AI-specific, while current AI list/search/count flows consume the visibility plugin decisions.
- Remove the management-side grant-list API from this PR scope.
- Extend `permissions.resource` schema support to 512 characters and add upgrade SQL/docs for MySQL, Derby, PostgreSQL, and Oracle.
- Update English/Chinese auth and visibility specs, v3 API surface docs, OpenAPI scenario docs, and focused tests.

## Verifying this change

Ran:

- `mvn -pl plugin-default-impl/nacos-default-auth-plugin spotless:apply`
- `mvn -pl plugin-default-impl/nacos-default-auth-plugin spotless:check`
- `mvn -pl plugin-default-impl/nacos-default-auth-plugin -am -Dtest=DefaultVisibilityGrantServiceTest,VisibilityGrantControllerV3Test,NacosRoleServiceDirectImplTest,NacosRoleServiceRemoteImplTest -Dmaven.compiler.forceJavacCompilerUse=true -Dsurefire.failIfNoSpecifiedTests=false -Djacoco.skip=true test`

Focused tests cover:

- one dedicated visibility role reused for multiple resources of the same user
- different users receiving different dedicated roles
- idempotent repeated grant/revoke
- revoke after grantee user deletion
- orphan dedicated role without permissions
- `r` and `rw` action semantics
- runtime authorized resource names from actual dedicated-role permissions
- direct and remote permission cache invalidation
- maximum 512-character canonical resource grant/revoke
- `ApiType.ADMIN_API` metadata for grant/revoke endpoints

* [x] Make sure there is a Github issue filed for the change.
* [x] Format the pull request title like `[ISSUE #15476] ...`.
* [x] Write necessary unit-test to verify your logic correction.
* [ ] Run full pre-submission checks.


ai/src/main/java/com/alibaba/nacos/ai/service/visibility/AiVisibilityResourceLocator.java
```diff
@@ -0,0 +1,54 @@
...
+
+import static com.alibaba.nacos.api.common.Constants.DEFAULT_NAMESPACE_ID;
+
+/**
+ * AI implementation of {@link VisibilityResourceLocator}.
+ *
+ * @author Zhengcy05
+ */
+@Service
+public class AiVisibilityResourceLocator implements VisibilityResourceLocator {
```
KomachiSion, line 36: Visibility插件本身没有AI语意，只是我们现在只实现了AI相关的资源。

KomachiSion: 这个评论针对的是后续所有的实现，包括类名， API等信息。

KomachiSion (changes requested):

ai/src/main/java/com/alibaba/nacos/ai/service/visibility/AiVisibilityResourceLocator.java
Zhengcy05: 明白了🫡

KomachiSion:
@Zhengcy05  Thanks for the contribution. The overall direction—keeping visibility grant management inside the auth plugin and resolving domain resource metadata through an SPI—is reasonable.

However, I think the RBAC mapping, permission lookup, database compatibility, and authentication scope need to be revised before this feature can be merged. Requesting changes for the following items.

## 1. Create one dedicated visibility role per user

The current implementation creates one internal role for every resource/action pair and encodes the namespace, resource type, resource name, and action into the role name.

This causes several problems:

- the number of distinct roles grows with the resource count;
- normal resource names already exceed the existing `roles.role varchar(50)` limit;
- permission information is duplicated in the role name;
- list authorization later needs to decode resources from role names.

For this user-specific grant model, please create at most one dedicated internal visibility role for each user when the user receives their first visibility grant:

```text
identity
  -> dedicated visibility role
      -> permission(resource A, r)
      -> permission(resource B, rw)
      -> ...
```

The intended persistence model should be similar to:

```text
roles:
username -> dedicated visibility role

permissions:
dedicated visibility role
    -> @@visibility/{namespaceId}/{resourceType}/{resourceName} : r|rw
```

The dedicated role name must be deterministic, reserved for visibility grants, unique per user, and bounded to the existing 50-character column.

Please do not attach these permissions to an arbitrary existing role. Existing roles may be shared by multiple users, so adding a user-specific visibility permission to such a role would grant access to every user bound to that role.

Grant should:

1. resolve or create the user's dedicated visibility role;
2. ensure the user-to-role binding exists;
3. add the raw resource/action permission idempotently.

Revoke should delete the corresponding permission. The empty internal role may either be retained or cleaned up.

A partially created role without a permission must never grant visibility.

## 2. Keep the authorization chain as identity -> role -> permission

`findAuthorizedResourceNames` currently derives authorized resources directly from encoded role names.

A role binding only establishes identity-to-role membership. It must not be treated as a permission by itself. Otherwise, an orphaned or manually created role can expose a private resource in list/search results even when the backing permission does not exist.

Please derive visible resources from the actual permissions attached to the current user's dedicated visibility role:

```text
identity
  -> visibility role
  -> actual permissions
  -> filter namespace/resource type/action
  -> extract resource names
  -> build list/search/count condition
```

This keeps single-resource authorization and list/search/count authorization on the same permission source of truth.

The resource grant-list API also needs the reverse lookup:

```text
resource permission -> visibility role -> user
```

Please add an exact persistence query and the required indexes for this path. It should not scan all roles and permissions or depend on decoding role names.

Please also ensure that grant and revoke invalidate or refresh the relevant role/permission caches so that a successful API response has clearly defined visibility semantics.

## 3. Store the raw resource identifier and expand it to 512 characters

The permission should store the original resource identifier:

```text
@@visibility/{namespaceId}/{resourceType}/{resourceName}
```

Using the current AI resource field limits, the complete identifier can require approximately 431 characters:

```text
13 + 128 + 1 + 32 + 1 + 256 = 431
```

The current built-in schemas are inconsistent:

| Database | Current resource length |
| --- | ---: |
| MySQL | 128 |
| Oracle | 255 |
| Derby | 512 |
| PostgreSQL | 512 |

For this PR, expanding `permissions.resource` to 512 is an acceptable short-term solution.

Please update:

- the MySQL initialization schema;
- the Oracle initialization schema, using `VARCHAR2(512 CHAR)`;
- the corresponding schema tests;
- migration SQL for existing MySQL and Oracle deployments;
- the upgrade/release documentation.

The MySQL `permissions` table currently does not declare a charset, engine, or row format and therefore inherits the user's database defaults. The absence of an explicit `utf8mb4` declaration does not guarantee that the table uses a single-byte charset.

With `utf8mb4`, the current unique index has a maximum declared size of approximately:

```text
(role 50 + resource 512 + action 8) * 4 = 2280 bytes
```

This fits the 3072-byte InnoDB index limit only under a compatible configuration, such as:

```text
MySQL >= 5.7.9
innodb_page_size = 16KB
ROW_FORMAT = DYNAMIC
```

The field was previously reduced from 255 to 128 because the unique index exceeded the historical 767-byte limit: https://github.com/alibaba/nacos/issues/11306

Therefore, the migration documentation must include preflight checks such as:

```sql
SELECT VERSION();
SHOW VARIABLES LIKE 'innodb_page_size';
SHOW VARIABLES LIKE 'innodb_default_row_format';
SHOW CREATE TABLE permissions;
```

The MySQL migration should explicitly rebuild the table with a compatible row format while changing the column, for example:

```sql
ALTER TABLE permissions
    ROW_FORMAT=DYNAMIC,
    MODIFY COLUMN resource VARCHAR(512) NOT NULL;
```

The exact charset and collation should also be explicitly defined instead of silently inheriting the database default. Please confirm whether permission resource comparison is intended to be case-sensitive before selecting the collation.

Users whose MySQL configuration cannot support the resulting index must be warned before upgrading rather than encountering a runtime permission-insert failure.

## 4. Handle database-independent indexing as a follow-up

A database-independent SHA-256 index is the preferred long-term design, but it does not need to expand the scope of this PR.

Please create a follow-up issue/design for a model such as:

```text
role
resource          -- raw resource identifier
resource_hash     -- SHA-256 of the canonical UTF-8 resource
action
```

The unique index would become:

```text
UNIQUE(role, resource_hash, action)
```

To reduce database-type differences, `resource_hash` can be stored as a normalized lowercase hexadecimal `CHAR(64)` value calculated in Java rather than using database-specific binary types or database hash functions.

The raw resource must remain the final authorization value. After looking up by hash, the implementation should still compare the raw resource using exact equality.

The follow-up design should define the canonicalization rules for:

- default namespace normalization;
- resource type normalization;
- resource name case sensitivity;
- UTF-8 encoding;
- migration and backfill of existing permission rows.

## 5. Match visibility resources exactly in memory

Visibility grants target one exact resource.

The current default RBAC matcher treats stored permission resources as regular expressions. A resource name containing `.`, `[]`, `+`, `(`, or other regex metacharacters may therefore match a different resource.

Please keep the original raw resource string in storage, but use exact in-memory matching for the reserved `@@visibility/` resource prefix.

For example:

```java
if (permissionResource.startsWith(VISIBILITY_RESOURCE_PREFIX)) {
    matched = permissionResource.equals(targetResource);
} else {
    matched = existingWildcardMatch(permissionResource, targetResource);
}
```

The persisted resource must not be regex-escaped or otherwise transformed. Existing wildcard behavior for Config/Naming permissions should remain unchanged.

## 6. Explicitly select the authentication scope

The visibility grant endpoints rely on the service layer to resolve the authenticated `NacosUser` and allow only the resource owner or a global administrator.

However, the current `@Secured` annotations default to `ApiType.OPEN_API`.

With the distribution defaults:

```properties
nacos.core.auth.enabled=false
nacos.core.auth.admin.enabled=true
```

the open API filter skips identity validation, while `isAnyAuthEnabled()` still returns true. Consequently, `resolveCurrentUsername()` returns null and even owner/global-admin requests are rejected.

Please explicitly assign the intended authentication scope, likely `ApiType.ADMIN_API`, and verify that identity-only authentication still allows the service layer to distinguish:

- the resource owner;
- a global administrator;
- an authenticated non-owner;
- an unauthenticated caller.

This behavior must be covered by auth-enabled integration tests. The current auth-disabled IT cannot validate the most important authorization semantics of this API.

## 7. Keep Prompt visibility integration in a separate issue/PR

Prompt has not yet fully joined the standard visibility flow. Its read/detail/version/download and list paths do not consistently apply standard readable checks and pre-pagination visibility conditions.

Refactoring Prompt visibility is outside the scope of this PR and should be handled by a separate issue and PR.

However, this PR must not accept a Prompt visibility grant and return success when that permission cannot affect Prompt runtime behavior.

Please introduce or validate a supported-resource-type capability so that the grant API only accepts resource types whose single-resource and list/search/count paths already support the standard visibility flow.

Unsupported resource types should return a clear unsupported error until their visibility integration is complete. The currently supported resource types must also be documented in the visibility/auth specs.

The same rule should apply to any other AI resource type that has not yet joined the standard visibility flow.

Grant cleanup when a resource is deleted is not required in this PR and can remain a later lifecycle enhancement.

## 8. Align the API and specification text

Please use the same endpoint consistently in:

- controller mappings;
- PR description;
- English and Chinese specs;
- V3 API surface;
- API test coverage documents.

The implementation currently uses:

```text
/v3/auth/visibility
```

but some text still refers to `/v3/auth/ai/visibility` or `/ai/visibility`.

## 9. Required tests

In addition to the current API contract tests, please cover:

- one user receiving grants for multiple resources while retaining one dedicated visibility role;
- multiple users receiving grants for the same resource;
- idempotent repeated grant and revoke;
- a visibility role without a backing permission granting no access;
- list/search/count being derived from actual permissions rather than role names;
- read grants not implying write;
- `rw` grants implying read;
- exact matching for resource names containing regex metacharacters;
- maximum supported resource identifier length;
- MySQL schema creation and migration under the documented configuration;
- Oracle `VARCHAR2(512 CHAR)` schema behavior;
- rejection of resource types that have not joined the standard visibility flow;
- auth-enabled owner, global-admin, non-owner, and unauthenticated behavior;
- permission/cache visibility after successful grant and revoke.

Once these items are addressed, the plugin-owned API and locator SPI should provide a reasonable base for explicit visibility grants without coupling the generic resource controllers directly to the default auth implementation.

KomachiSion:
One more need todo after this PR:

## Follow-up proposal: use a SHA-256 resource index for database-independent permission storage

The current PR can use `resource VARCHAR(512)` as a short-term solution, with the required MySQL configuration and migration documentation.

For the long term, I suggest creating a separate issue/design to replace the raw resource string in permission indexes with a Java-computed SHA-256 digest. This would remove the index-length dependency on database charset, collation, page size, and database-specific index limits while retaining the original resource value for authorization.

### Proposed data model

The `permissions` table should continue storing the original resource identifier and add a fixed-length hash column:

```text
role
resource          -- original canonical resource identifier
resource_hash     -- SHA-256(resource), lowercase hexadecimal
action
```

A cross-database logical schema could use:

```sql
resource      VARCHAR(512) NOT NULL,
resource_hash CHAR(64) NOT NULL
```

Using lowercase hexadecimal `CHAR(64)` is preferable to database-specific binary types:

```text
MySQL       BINARY(32)
Oracle      RAW(32)
PostgreSQL  BYTEA
Derby       CHAR FOR BIT DATA
```

The hash must be calculated in Java, not with database-specific SQL functions.

### Canonical hash input

The implementation should first build the canonical raw resource identifier:

```text
@@visibility/{normalizedNamespaceId}/{normalizedResourceType}/{resourceName}
```

It should then calculate:

```text
resource_hash = lowercaseHex(
    SHA-256(canonicalResource.getBytes(StandardCharsets.UTF_8))
)
```

The canonicalization rules must be defined in the visibility/auth specification:

- a blank namespace is converted to the default namespace;
- resource type uses the same normalization as authorization;
- resource name preserves its defined case-sensitive identity;
- no additional Unicode normalization is applied unless the resource model explicitly requires it;
- UTF-8 is always used;
- `role` and `action` remain separate index fields and are not included in `resource_hash`.

The hash should always be calculated from the exact canonical string that is stored in `resource`.

### Indexes

The current unique index:

```text
UNIQUE(role, resource, action)
```

should eventually be replaced with:

```text
UNIQUE(role, resource_hash, action)
```

The grant-list API also needs efficient reverse lookup from a resource to users, so an additional index should be considered:

```text
INDEX(resource_hash, action, role)
```

This supports both directions:

```text
authorization:
identity -> role -> resource_hash/action permission


** Of cource if you has better way to solve this problem or better design **

**Comment by KomachiSion** [comment]
And @Zhengcy05 Would you like provide your dingtalk number?

**Comment by Zhengcy05** [comment]
Understood. I will focus on this PR first, which might take a bit longer. 🫡

**Comment by Zhengcy05** [comment]
> And @Zhengcy05 Would you like provide your dingtalk number?

My DingTalk account : 1ex-76mml35v5b

**Comment by KomachiSion** [comment]
Thanks for the update. I re-reviewed the current implementation.

The overall design direction is now acceptable:

- Each grantee uses a deterministic, dedicated visibility role.
- The authorization chain remains `identity -> role -> permission`.
- The real canonical resource string is stored in `permissions.resource`.
- One user can reuse the same dedicated role for multiple visibility grants.
- Runtime authorized resource names are derived from actual permissions.
- The short-term resource length is extended to 512 characters, while a hash-based storage design can be considered separately.

Before merging, please address the following items.

## 1. Mark the new visibility APIs as `ADMIN_API`

PR [#15563](https://github.com/alibaba/nacos/pull/15563) has been merged and establishes that the existing `/v3/auth/*` management APIs use `ApiType.ADMIN_API`.

Therefore, the new visibility grant APIs introduced by this PR must follow the same convention. Please add `apiType = ApiType.ADMIN_API` to the `@Secured` annotations of the remaining grant and revoke endpoints:

```java
@Secured(
    resource = AuthConstants.VISIBILITY_RESOURCE,
    action = ActionTypes.WRITE,
    apiType = ApiType.ADMIN_API,
    tags = Constants.Tag.ONLY_IDENTITY
)
```

Please also import:

```java
import com.alibaba.nacos.api.common.ApiType;
```

This is part of this PR because it defines the security scope of the newly introduced APIs. The broader global-admin identity handling and the behavior of historical Auth APIs can still be consolidated separately.

Please add a focused controller test that verifies the `@Secured` metadata of both the POST and DELETE endpoints, especially `apiType = ApiType.ADMIN_API`.

## 2. Remove the management-side grant-list capability

For the initial version, this PR only needs to provide:

- Grant visibility authorization.
- Revoke visibility authorization.

The management-side `GET /v3/auth/visibility/list` API is not required and should be removed. Revocation does not depend on grant-list because the dedicated role can be derived from the grantee and the target permission is identified by resource and action.

Please remove the list-only implementation, including:

- The `/list` controller endpoint.
- `VisibilityGrantService.list`.
- `VisibilityGrantInfo`.
- `DefaultVisibilityGrantService.list`.
- Persistence queries such as `getPermissionsByResource`.
- The resource-filter extension added to `PermissionControllerV3`.
- Reverse indexes introduced only for grant-list queries, such as `idx_permission_resource` and `idx_role_user`.
- List-specific unit tests, OpenAPI IT cases and specification content.

Please retain the runtime authorization query, such as `findAuthorizedResourceNames`, and its use by AI resource list/search/count operations. Runtime resource filtering is required by the visibility feature and is different from exposing a management-side grant-list API.

## 3. Complete the 512-character schema and upgrade delivery

The short-term solution of extending `permissions.resource` to 512 characters is acceptable.

Please ensure that:

- New-install schemas for all supported datasources used by this feature accept the same 512-character resource contract.
- The raw canonical resource string remains stored in the database; do not persist a translated or escaped representation.
- Exact upgrade SQL is provided in the version upgrade documentation.
- The MySQL upgrade documentation includes the required index-length/storage-mode preflight and explains that users must configure a compatible mode before applying the migration.
- Upgrade SQL is not only placed under module `META-INF` resources, because those files are not directly available to operators in the final distribution.

If the upgrade SQL files must remain in this repository, please also copy them into the final distribution, for example through `nacos-default-plugin-all/pom.xml` into `distribution/conf`. Otherwise, the upgrade documentation should be the authoritative delivery location.

After removing grant-list, list-only reverse-index migrations should also be removed.

A hash-based resource-key design can be handled as a separate long-term change because it requires a larger persistence and compatibility design.

## 4. Align the API contract, PR description and specifications

The implementation uses:

```text
/v3/auth/visibility
```

but the PR description/specification still contains `/v3/auth/ai/visibility` in some places.

Please consistently use `/v3/auth/visibility` and update:

- The PR description.
- English and Chinese Auth/visibility specifications.
- English and Chinese V3 API surface documents.
- OpenAPI test scenario and coverage documents.

After removing grant-list, the public management API surface should contain only:

- `POST /v3/auth/visibility`
- `DELETE /v3/auth/visibility`

The documents should also state that these endpoints are `ADMIN_API`.

## 5. Complete focused tests for the new authorization model

Please update the OpenAPI scenario from grant/list/revoke to grant/revoke, and cover the important authorization invariants:

- Multiple resources granted to the same user reuse one dedicated visibility role.
- Different users receive different dedicated roles.
- Repeated grant and revoke operations are idempotent.
- Revoke can clean up a grant for a grantee whose user record has already been deleted.
- An orphan dedicated role without permissions is handled correctly.
- `r` and `rw` action semantics are preserved.
- Runtime authorized resource names come from the actual permissions associated with the dedicated role.
- Permission cache invalidation works for both direct and remote role-service paths.
- A canonical resource at the supported maximum length can be granted and revoked.
- Both visibility controller methods declare `ApiType.ADMIN_API`.

The broader global-admin identity propagation behavior does not need to be redesigned in this PR.

## 6. Resolve the merge conflict

The PR is currently reported as conflicting with the latest `develop`. Please rebase or merge the latest `develop`, resolve the conflict, and rerun the affected checks.

To keep the scope clear, this PR does not need to fix:

- Historical permission-regex behavior.
- Historical dangling role or permission data.
- Broader global-admin identity handling across existing Auth APIs.
- Prompt integration with the standard visibility flow.
- Automatic grant cleanup when an AI resource is deleted.
- The long-term hash-based permission resource schema.

Those can be handled in dedicated follow-up changes.

And One more thing, your pr is confilict with other develop, please rebase and solve conflicts.

Zhengcy05:
This is a checklist for myself to mark my progress
- [x] 1. Mark the new visibility APIs as ADMIN_API
- [x] 2. Remove the management-side grant-list capability
- [x] 3. Complete the 512-character schema and upgrade delivery
- [x] 4. Align the API contract, PR description and specifications （I have checked and changed all the old path names.）
- [x] 5. Complete focused tests for the new authorization model
- [x] 6. Resolve the merge conflict

Zhengcy05:
> Thanks for the update. I re-reviewed the current implementation.
> 
> The overall design direction is now acceptable:
> 
> * Each grantee uses a deterministic, dedicated visibility role.
> * The authorization chain remains `identity -> role -> permission`.
> * The real canonical resource string is stored in `permissions.resource`.
> * One user can reuse the same dedicated role for multiple visibility grants.
> * Runtime authorized resource names are derived from actual permissions.
> * The short-term resource length is extended to 512 characters, while a hash-based storage design can be considered separately.
> 
> Before merging, please address the following items.
> 
> ## 1. Mark the new visibility APIs as `ADMIN_API`
> PR [#15563](https://github.com/alibaba/nacos/pull/15563) has been merged and establishes that the existing `/v3/auth/*` management APIs use `ApiType.ADMIN_API`.
> 
> Therefore, the new visibility grant APIs introduced by this PR must follow the same convention. Please add `apiType = ApiType.ADMIN_API` to the `@Secured` annotations of the remaining grant and revoke endpoints:
> 
> ```java
> @Secured(
>     resource = AuthConstants.VISIBILITY_RESOURCE,
>     action = ActionTypes.WRITE,
>     apiType = ApiType.ADMIN_API,
>     tags = Constants.Tag.ONLY_IDENTITY
> )
> ```
> 
> Please also import:
> 
> ```java
> import com.alibaba.nacos.api.common.ApiType;
> ```
> 
> This is part of this PR because it defines the security scope of the newly introduced APIs. The broader global-admin identity handling and the behavior of historical Auth APIs can still be consolidated separately.
> 
> Please add a focused controller test that verifies the `@Secured` metadata of both the POST and DELETE endpoints, especially `apiType = ApiType.ADMIN_API`.
> 
> ## 2. Remove the management-side grant-list capability
> For the initial version, this PR only needs to provide:
> 
> * Grant visibility authorization.
> * Revoke visibility authorization.
> 
> The management-side `GET /v3/auth/visibility/list` API is not required and should be removed. Revocation does not depend on grant-list because the dedicated role can be derived from the grantee and the target permission is identified by resource and action.
> 
> Please remove the list-only implementation, including:
> 
> * The `/list` controller endpoint.
> * `VisibilityGrantService.list`.
> * `VisibilityGrantInfo`.
> * `DefaultVisibilityGrantService.list`.
> * Persistence queries such as `getPermissionsByResource`.
> * The resource-filter extension added to `PermissionControllerV3`.
> * Reverse indexes introduced only for grant-list queries, such as `idx_permission_resource` and `idx_role_user`.
> * List-specific unit tests, OpenAPI IT cases and specification content.
> 
> Please retain the runtime authorization query, such as `findAuthorizedResourceNames`, and its use by AI resource list/search/count operations. Runtime resource filtering is required by the visibility feature and is different from exposing a management-side grant-list API.
> 
> ## 3. Complete the 512-character schema and upgrade delivery
> The short-term solution of extending `permissions.resource` to 512 characters is acceptable.
> 
> Please ensure that:
> 
> * New-install schemas for all supported datasources used by this feature accept the same 512-character resource contract.
> * The raw canonical resource string remains stored in the database; do not persist a translated or escaped representation.
> * Exact upgrade SQL is provided in the version upgrade documentation.
> * The MySQL upgrade documentation includes the required index-length/storage-mode preflight and explains that users must configure a compatible mode before applying the migration.
> * Upgrade SQL is not only placed under module `META-INF` resources, because those files are not directly available to operators in the final distribution.
> 
> If the upgrade SQL files must remain in this repository, please also copy them into the final distribution, for example through `nacos-default-plugin-all/pom.xml` into `distribution/conf`. Otherwise, the upgrade documentation should be the authoritative delivery location.
> 
> After removing grant-list, list-only reverse-index migrations should also be removed.
> 
> A hash-based resource-key design can be handled as a separate long-term change because it requires a larger persistence and compatibility design.
> 
> ## 4. Align the API contract, PR description and specifications
> The implementation uses:
> 
> ```
> /v3/auth/visibility
> ```
> 
> but the PR description/specification still contains `/v3/auth/ai/visibility` in some places.
> 
> Please consistently use `/v3/auth/visibility` and update:
> 
> * The PR description.
> * English and Chinese Auth/visibility specifications.
> * English and Chinese V3 API surface documents.
> * OpenAPI test scenario and coverage documents.
> 
> After removing grant-list, the public management API surface should contain only:
> 
> * `POST /v3/auth/visibility`
> * `DELETE /v3/auth/visibility`
> 
> The documents should also state that these endpoints are `ADMIN_API`.
> 
> ## 5. Complete focused tests for the new authorization model
> Please update the OpenAPI scenario from grant/list/revoke to grant/revoke, and cover the important authorization invariants:
> 
> * Multiple resources granted to the same user reuse one dedicated visibility role.
> * Different users receive different dedicated roles.
> * Repeated grant and revoke operations are idempotent.
> * Revoke can clean up a grant for a grantee whose user record has already been deleted.
> * An orphan dedicated role without permissions is handled correctly.
> * `r` and `rw` action semantics are preserved.
> * Runtime authorized resource names come from the actual permissions associated with the dedicated role.
> * Permission cache invalidation works for both direct and remote role-service paths.
> * A canonical resource at the supported maximum length can be granted and revoked.
> * Both visibility controller methods declare `ApiType.ADMIN_API`.
> 
> The broader global-admin identity propagation behavior does not need to be redesigned in this PR.
> 
> ## 6. Resolve the merge conflict
> The PR is currently reported as conflicting with the latest `develop`. Please rebase or merge the latest `develop`, resolve the conflict, and rerun the affected checks.
> 
> To keep the scope clear, this PR does not need to fix:
> 
> * Historical permission-regex behavior.
> * Historical dangling role or permission data.
> * Broader global-admin identity handling across existing Auth APIs.
> * Prompt integration with the standard visibility flow.
> * Automatic grant cleanup when an AI resource is deleted.
> * The long-term hash-based permission resource schema.
> 
> Those can be handled in dedicated follow-up changes.
> 
> And One more thing, your pr is confilict with other develop, please rebase and solve conflicts.

Thank you very much for your careful review. I have made all the corrections and made a separate commit for each point.😚

KomachiSion:
Thanks for the update. The main visibility authorization design now matches the agreed direction. The authentication-filter consolidation can be handled separately later.

There are three remaining items to address in this PR:

1. Please remove the grant-list-only indexes still present in:

persistence/src/test/resources/META-INF/derby-schema.sql

The indexes are:

CREATE INDEX idx_role_user
    ON roles(role, username);

CREATE INDEX idx_permission_resource
    ON permissions(resource, action, role);

The current implementation queries roles by username and permissions by role, which are already covered by the existing unique indexes. These two indexes are leftovers from the removed management-side grant-list capability and also make the Derby test schema inconsistent with the production schemas.

2. Please correct the remaining old visibility API paths in:

- specs/en/http-api/v3-api-surface.md
- specs/zh-cn/http-api/v3-api-surface.md

They still refer to /ai/visibility. The actual API path is /v3/auth/visibility.

3. Please revert the role-wide cache reload changes in:

- NacosRoleServiceDirectImpl.deleteRole(String role)
- NacosRoleServiceRemoteImpl.deleteRole(String role)

The visibility grant flow does not call role-wide deletion. It only uses the user-specific role binding operation during rollback. These changes modify historical role-deletion behavior and introduce a synchronous full cache reload outside the scope of this PR.

Please retain the granular cache invalidations required by this feature:

- invalidate the affected user's role cache after adding or deleting a user-role binding;
- invalidate the affected role's permission cache after adding or deleting a permission.

I locally verified the focused auth/visibility tests, database schema tests, OpenAPI test compilation, and upgrade SQL packaging. After the three items above are addressed, the PR should be ready for another review.

KomachiSion (dismissed):

PR #15542, opened 2026-07-21
[ISSUE #15541] Add Agentic Resource Discovery support

### LINKED ISSUE #15541: Support Agentic Resource Discovery in Nacos AI Registry
**Is your feature request related to a problem? Please describe.**

Nacos AI Registry currently manages AI resources such as Skills, Prompts, and MCP servers, but clients do not have a protocol-aligned way to discover, search, explore, and retrieve those resources. Existing resources also need to become searchable without being republished after ARD is enabled.

Add Agentic Resource Discovery (ARD) support to the Nacos AI Registry, including:

- ARD search, list, explore, catalog, artifact, and host-level well-known endpoints.
- Visibility and namespace authorization filtering for ARD responses.
- Persistent entry and chunk indexes for Skill, Prompt, and MCP resources.
- Hybrid keyword and vector retrieval with extensible vector storage and an optional PostgreSQL/pgvector implementation.
- Optional LLM-assisted index text enhancement while retaining a local deterministic embedding fallback.
- Latest-online-version lifecycle handling and cluster-safe startup backfill for existing resources.
- Database schemas, configuration, and focused tests for the new capability.

Clients could use the existing resource-specific Nacos APIs and implement discovery independently, but that duplicates filtering, ranking, pagination, artifact construction, and indexing logic in every client. Keeping ARD in the registry provides one consistent discovery contract and lets deployments choose whether to enable vector or LLM-backed enhancements.

**Additional context**

The initial implementation focuses on registry-local and private-network discovery. Public federation across external registries is outside this issue.

saynoword:

Fixes #15541

Add Agentic Resource Discovery (ARD) support to the Nacos AI Registry so clients can discover, search, explore, and retrieve Skill, Prompt, and MCP resources through a consistent protocol. ARD is disabled by default and can be enabled explicitly without imposing ARD database, pgvector, or LLM runtime requirements on existing Nacos deployments.

## Brief changelog

- Add ARD search, list, explore, catalog, artifact, and host-level well-known endpoints with namespace authorization and visibility filtering.
- Keep external ARD models, protocol constants, endpoint paths, and response assembly in `ai-registry-adaptor`; internal index models and vector SPI retain only canonical resource/search fields.
- Encapsulate AI resource storage metadata parsing and file routing behind the `ai` module so the protocol adaptor does not depend on concrete storage implementations.
- Add persistent entry and chunk indexes, hybrid keyword/vector retrieval, ranking, pagination, and latest-online-version lifecycle handling.
- Add an extensible vector index SPI and a default PostgreSQL/pgvector implementation that can use dedicated vector storage.
- Discover vector plugins through a shared SPI registry so unified plugin loading works before or without the ARD Spring child context.
- Add optional OpenAI-compatible LLM index enhancement and local deterministic hashing embeddings.
- Add cluster-safe startup backfill for existing Skill, Prompt, and MCP resources.
- Add the default-off `nacos.ai.ard.enabled` global switch and conditionally load ARD endpoints, indexing, backfill, LLM enhancement, and vector routing.
- Register ARD endpoints with the standard Nacos authentication method cache and bind anonymous access to the existing `anonymous.ai.enabled` setting.
- Add datasource schemas, configuration, and focused unit tests for ARD behavior; derived protocol fields are no longer persisted in ARD index tables.

## Verifying this change

- `mvn -pl ai -Dtest=ArdEntryBuilderTest,ArdChunkBuilderTest,ArdIndexBuildServiceImplTest,AiResourceArdIndexContentLoaderTest,AiResourceFileReaderTest test` (19 tests passed)
- `mvn -pl ai-registry-adaptor test` (130 tests passed)
- AI vector registry, provider, and router focused tests passed (9 tests); the PostgreSQL implementation also passed a real SPI discovery integration test.
- `mvn -pl ai-registry-adaptor -am -Dmaven.test.skip=true install` (30 reactor modules passed)
- `mvn -pl plugin/ai,ai,ai-registry-adaptor,plugin-default-impl/nacos-default-ai-vector-plugin -DskipTests spotless:check`
- `mvn -pl plugin/ai,ai,ai-registry-adaptor,plugin-default-impl/nacos-default-ai-vector-plugin -DskipTests -Dcheckstyle.consoleOutput=true checkstyle:check` (0 violations)
- Runtime authentication verification on the local deployment: all six ARD endpoints returned 403 without or with an invalid token and 200 with a valid token while `anonymous.ai.enabled=false`.

* [ ] Run `mvn -B clean package apache-rat:check spotbugs:check -DskipTests` to make sure basic checks pass. Run `mvn clean install` to make sure unit-test pass. Run `mvn clean test-compile failsafe:integration-test` to make sure integration-test pass.


KomachiSion (changes requested):
1. model不要定义在api里， 因为这个不是nacos自己通信用的api， 可以直接放在protocol-adapter的module里。
2. plugin的加载方式最近有一定的修改， 建议按照新的plugin加载方式实现。

KomachiSion:
Thanks for adding ARD support. The overall direction is reasonable, especially keeping external ARD models in `ai-registry-adaptor` and defining the vector SPI under `plugin/ai`. However, I found several architectural and compatibility issues that should be addressed before merging:

1. **The PostgreSQL main schema currently requires pgvector even when ARD is disabled.**

   `pg-schema.sql` unconditionally executes `CREATE EXTENSION IF NOT EXISTS vector` and creates `ai_resource_ard_embedding_pg`. A fresh Nacos deployment may therefore fail when pgvector is not installed or the database user cannot create extensions, even with `nacos.ai.ard.enabled=false`.

   Please keep only the ARD entry/chunk tables in the main datasource schema. The pgvector extension and embedding table should be owned by `nacos-default-ai-vector-plugin` and initialized through its dedicated optional schema.

2. **Several response models do not match the current ARD specification.**

   - `ArdListResponse` serializes entries as `results`, while the ARD OpenAPI requires `items`.
   - Search result `source` is set to `nacos-local`, while the specification requires a URI.
   - The generated `trustManifest` does not include the required `identity`.
   - ARD controllers use `@NacosApi`, so errors are serialized by `NacosApiExceptionHandler` as Nacos `Result<T>`, while ARD requires an `{errorCode, message}` response.

   Please pin a specific upstream ARD Schema/OpenAPI version in `ai-registry-adaptor` and add schema validation or official conformance tests. ARD endpoints should also have protocol-specific exception handling instead of using the Nacos v3 response envelope.

3. **The generated Skill artifact URL points to the wrong web context by default.**

   `ArdSearchServiceImpl` builds Skill URLs using `/v3/client/ai/skills` and makes them absolute with the current ARD adaptor request origin. However, the ARD adaptor normally runs on port 9080, while `SkillClientController` is served by the main Nacos server context, normally on port 8848. Without a gateway routing both paths, the generated URL returns 404.

   Other resource types already use `/v3/ai/ard/artifacts`. Please either serve the complete Skill ZIP through the adaptor-owned artifact endpoint or introduce a separate and explicit main-server API base URL. This should be covered by an integration test using the actual two web contexts.

4. **The search domain logic is placed in the protocol adaptor.**

   `ArdSearchServiceImpl` directly owns repository access, keyword/vector recall, ranking, latest-version validation, visibility checks, filtering, and pagination. This is more than protocol adaptation and conflicts with the intended module boundary.

   It also causes a correctness issue: candidates are capped before visibility and latest-version filtering, so valid visible resources can be omitted when earlier candidates are stale or invisible.

   Please move recall, ranking, visibility query advice, current-version validation, and cursor pagination into an `ai`-module discovery service. The adaptor should only validate ARD requests and translate canonical search results into ARD DTOs.

5. **Index maintenance has no reliable consistency or retry mechanism.**

   `JdbcArdIndexRepository.replaceEntry` performs delete, entry insert, and chunk inserts without a transaction, and the vector index is updated separately. Skill, Prompt, and MCP operation services catch indexing failures and only log warnings. The canonical resource write may therefore succeed while the ARD index remains missing, partial, or stale until a later restart backfill.

   Please make the relational replacement transactional and define an explicit eventual-consistency mechanism for index/vector failures, such as idempotent lifecycle events with durable retry or periodic reconciliation. Centralizing this in an ARD index consumer would also avoid scattering best-effort rebuild/delete calls across Skill, Prompt, and MCP services.

In summary, the placement of protocol DTOs/controllers and the vector SPI/default implementation is generally correct, but the search application service should move into `ai`, and the pgvector schema should move alongside the default vector plugin. I also recommend adding an ARD-specific specification, an AI vector plugin specification, official protocol conformance tests, and end-to-end tests for the separate adaptor web context.

KomachiSion:
Thanks for the update. The previous changes have substantially improved the ARD integration, especially the protocol-specific response handling, adaptor-owned artifact routing, durable index tasks, discovery ownership in the AI module, and pgvector schema isolation.

After another architecture and API compliance review, I think the following issues should still be addressed.

## 1. Support all standard ARD federation values

`ArdSearchServiceImpl.validateAndBuildContext` currently treats an omitted `federation` as `none` and rejects explicit `auto` or `referrals` values with HTTP 400.

These are all valid values in the pinned ARD OpenAPI. A standard client may explicitly send:

```json
{
  "query": {
    "text": "weather tools"
  },
  "federation": "auto"
}
```

Even when Nacos has no upstream registries configured, it should not reject a valid protocol value.

Suggested behavior:

- `none`: search only the local Nacos index.
- `auto`: fall back to local search when no upstream registry is configured.
- `referrals`: return local results with an empty `referrals` array when no referrals are configured.

This preserves standard client compatibility without requiring a complete federation implementation in this PR.

## 2. Align the `GET /agents` filter parser with the pinned ARD syntax

The pinned ARD API documents expressions such as:

```text
type = 'application/mcp-server-card+json' AND updatedAfter > '2026-01-01T00:00:00Z'
```

The current `normalizeListFilter` implementation:

- splits only on `;` or `&`;
- recognizes only `=` and `:`;
- does not recognize `AND`;
- does not support comparison operators such as `>`;
- is not quote-aware.

As a result, a valid ARD filter may be silently interpreted as one incorrect value and return an empty or incorrect result instead of `INVALID_ARGUMENT`.

Please either:

1. implement the filter syntax defined by the pinned ARD version, including quoting, logical composition, and timestamp comparison; or
2. return `501 Not Implemented` for this optional endpoint until the grammar is supported.

A standard expression should not be silently misinterpreted.

## 3. Encode broad Nacos resource names into schema-valid ARD identifiers

It is reasonable for Nacos to support a broader resource-name character set than ARD. However, the external ARD `identifier` still needs to satisfy the pinned catalog schema.

`buildIdentifier` currently concatenates the raw namespace and resource name into the URN. Names already used by this PR's tests, such as:

```text
avatar prompt
mcp/avatar server
```

produce identifiers containing spaces or `/`, which do not match the ARD identifier pattern.

Please keep the broad internal Nacos resource-name support, but introduce a stable and collision-safe encoding for each external URN segment.

Simple replacement with `_` should be avoided because different resource names could produce the same identifier. Please also add catalog schema tests covering spaces, slashes, Unicode, and other valid Nacos resource names.

## 4. Do not calculate Explore facets from only the first 1,000 results

The current Explore implementation requests one discovery page with a maximum of 1,000 candidates and calculates facets in the adaptor from that page.

When more than 1,000 eligible resources match the query:

- facet counts are incomplete;
- bucket ordering may be incorrect;
- `otherCount` may be incorrect.

This is an API correctness issue rather than only a performance concern.

Please move facet aggregation into the canonical AI Resource Discovery layer, where it can aggregate the complete eligible matched set. Alternatively, iterate through the full matched set with an internal cursor before calculating the response. The adaptor should only translate the canonical aggregation result into the ARD Explore DTO.

## 5. Clarify whether the index is generic AI Resource Discovery infrastructure or ARD-private state

The new relational index currently consists of:

- `ai_resource_ard_entry`: resource-level search entries;
- `ai_resource_ard_chunk`: keyword-search chunks and vector-generation input;
- `ai_resource_ard_index_task`: durable index maintenance tasks.

The search and index-maintenance logic is now located in the `ai` module and described as protocol-neutral/canonical discovery. Architecturally, it could be reused by both ARD and a future native Nacos AI Resource Search API:

```text
ARD API -----------\
                    -> AI Resource Discovery -> relational index + optional vector index
Nacos Search API --/
```

However, the implementation is still strongly ARD-bound:

- the tables, models, repositories, and configuration keys use `ard` naming;
- `AiResourceDiscoveryService` is conditional on `nacos.ai.ard.enabled`;
- the service returns the `ArdEntry` persistence model;
- the current only caller is the ARD adaptor.

This would make a future native Nacos Search API depend on enabling ARD and on ARD-named persistence types.

Please make the ownership explicit in this PR. The preferred direction is to treat this as generic AI Resource Discovery infrastructure:

- use protocol-neutral discovery/index domain models and repository names;
- return a canonical discovery DTO instead of exposing the persistence entity;
- separate the generic discovery-index switch from the ARD protocol switch;
- let the AI module own indexing, recall, visibility/version checks, pagination, and aggregation;
- let `ai-registry-adaptor` own only ARD request validation and DTO/error/artifact mapping;
- keep the generic relational index available across supported databases;
- keep only the PostgreSQL embedding table and pgvector extension in the optional vector implementation schema.

If the index is intentionally ARD-private, the specification and service naming should no longer describe it as a protocol-neutral canonical discovery capability.

## Non-blocking follow-up: adaptor HTTP integration and protocol conformance tests

The current `ai-registry-adaptor` module does not yet have a shared standalone HTTP integration-test framework. This does not need to block this PR, but please create a follow-up issue so ARD, MCP Registry, and Skill Registry can use one common test infrastructure.

Suggested issue title:

```text
[AI Registry Adaptor] Add standalone HTTP integration test and protocol conformance framework
```

Suggested scope:

- start the real adaptor web context on its independent port;
- exercise APIs through real HTTP serialization and routing;
- cover ARD, MCP Registry, and Skill Registry protocols;
- run the pinned official ARD conformance runner or an equivalent test matrix;
- verify request validation, protocol error bodies, pagination, artifact routing, and separate main-server/adaptor ports;
- integrate the suite into CI.

KomachiSion (approved):

Issue #15625, opened 2026-07-31
[BUG] ConfigCacheService.getGrayLastModifiedTs NPE risk & ConfigDiskServiceFactory missing volatile


cyforkk:
## Describe the bug

Two issues found in the `config` module:

### 1. Null-pointer dereference risk in `ConfigCacheService.getGrayLastModifiedTs`

**File:** `config/src/main/java/com/alibaba/nacos/config/server/service/ConfigCacheService.java` (line 494)

The method calls `item.getConfigCacheGray()` without null-checking `item`, even though `item` comes from `CACHE.get(groupKey)` which can return `null`.

Every sibling method in the same class correctly guards `item` (e.g. `getContentGrayMd5` line 487, `getGrayRule` line 506). This is a clear copy-paste oversight.

### 2. Broken double-checked locking in `ConfigDiskServiceFactory`

**File:** `config/src/main/java/com/alibaba/nacos/config/server/service/dump/disk/ConfigDiskServiceFactory.java` (line 26)

The singleton uses double-checked locking but the field is not declared `volatile`. Without `volatile`, the Java Memory Model permits a thread to see a non-null reference to an object whose constructor has not finished running (broken DCL anti-pattern since Java 5).

## Expected behavior

1. `getGrayLastModifiedTs` should return `0` when `item` is null (consistent with siblings).
2. `ConfigDiskServiceFactory.configDiskService` should be `volatile` to ensure safe publication.

## Fix

Minimal targeted fixes matching existing patterns in the codebase. Will submit a PR.

PR #15626, opened 2026-07-31
[ISSUE #15625] Fix NPE risk in getGrayLastModifiedTs and missing volatile in ConfigDiskServiceFactory

cyforkk:

Fixes two bugs in the `config` module:

1. **NPE risk in `ConfigCacheService.getGrayLastModifiedTs`** — The method calls `item.getConfigCacheGray()` without null-checking `item`, even though `item` comes from `CACHE.get(groupKey)` which can return `null`. Every sibling method in the same class correctly guards `item` (`getContentGrayMd5`, `getGrayRule`, `getLastModifiedTs`), making this a clear copy-patch oversight. Now returns `0` consistently when `item` is null.

2. **Broken double-checked locking in `ConfigDiskServiceFactory`** — The singleton field `configDiskService` was not declared `volatile`. Without `volatile`, the Java Memory Model permits a thread to see a non-null reference to an object whose constructor has not finished running (textbook broken DCL anti-pattern since Java 5). Added `volatile` to ensure safe publication.

## Brief changelog

- `ConfigCacheService.java`: added `item == null ||` guard in `getGrayLastModifiedTs`.
- `ConfigDiskServiceFactory.java`: added `volatile` to `configDiskService` field.

## Verifying this change

- `mvn compile -pl config -am` ✅ BUILD SUCCESS
- `mvn test -pl config -am -Dtest="ConfigCacheServiceTest,ConfigDiskServiceFactoryTest"` ✅ 38 tests passed, 0 failures
- `mvn spotless:check -pl config` ✅ formatting compliant

Both fixes are minimal, localized, and match existing patterns in the codebase.

Assisted-by: Claude Code

### DIFF
--- a/config/src/main/java/com/alibaba/nacos/config/server/service/ConfigCacheService.java
+++ b/config/src/main/java/com/alibaba/nacos/config/server/service/ConfigCacheService.java
@@ -493,7 +493,8 @@ public class ConfigCacheService {
     
     public static long getGrayLastModifiedTs(String groupKey, String grayName) {
         CacheItem item = CACHE.get(groupKey);
-        if (item.getConfigCacheGray() == null || !item.getConfigCacheGray().containsKey(grayName)) {
+        if (item == null || item.getConfigCacheGray() == null
+            || !item.getConfigCacheGray().containsKey(grayName)) {
             return 0;
         }
         ConfigCache configCacheGray = item.getConfigCacheGray().get(grayName);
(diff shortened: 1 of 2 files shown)


KomachiSion:
Please sign cla first.

cyforkk:
ok，bro

KomachiSion (approved):

PR #15630, opened 2026-07-31
[ISSUE #15625] Add volatile to StsCredentialHolder.stsCredential for thread-safe publication

cyforkk:

Fixes a thread-safety issue in `StsCredentialHolder`:

The `stsCredential` field is read and written concurrently by multiple threads (via `NamingResourceInjector`, `ConfigResourceInjector`, `AiResourceInjector`, `LockResourceInjector`) **without synchronization and without `volatile`**.

Without `volatile`, the Java Memory Model permits:
- A thread to observe a **partially-constructed** `StsCredential` (non-null reference but stale/default field values)
- A thread to see a stale cached value indefinitely (no happens-before relationship between the write and subsequent reads)

This is the textbook "unsafe publication" race condition.

**Key observation**: The sibling class `Credentials.java` (same package) correctly marks all its fields `volatile` (lines 26, 28, 30) — this was simply forgotten in `StsCredentialHolder`.

## Brief changelog

- `StsCredentialHolder.java`: added `volatile` to `stsCredential` field.

## Verifying this change

- `mvn compile -pl client-basic -am` ✅ BUILD SUCCESS
- `mvn test -pl client-basic -am -Dtest="StsCredentialHolderTest"` ✅ 5 tests passed, 0 failures
- `mvn spotless:check -pl client-basic` ✅ formatting compliant

The fix is a single-word addition with zero API/behavior change — purely a concurrency correctness improvement.

Assisted-by: Claude Code

### DIFF
--- a/client-basic/src/main/java/com/alibaba/nacos/client/auth/ram/identify/StsCredentialHolder.java
+++ b/client-basic/src/main/java/com/alibaba/nacos/client/auth/ram/identify/StsCredentialHolder.java
@@ -38,7 +38,7 @@ public class StsCredentialHolder {
     
     private static final StsCredentialHolder INSTANCE = new StsCredentialHolder();
     
-    private StsCredential stsCredential;
+    private volatile StsCredential stsCredential;
     
     private StsCredentialHolder() {
     }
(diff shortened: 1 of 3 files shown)


KomachiSion:
duplicate with #15631

PR #14442, opened 2026-02-03
fix(#14402) fix derby and mysql like with escape

chlch:
## What is the purpose of the change

  fix "like '%_%'" this situation

## Brief changelog

mainly fix derby datasource

PR #15711, opened 2026-08-14
[ISSUE #15710] Declare the LIKE escape clause for dialects without a default escape character

### LINKED ISSUE #15710: Fuzzy search with an underscore returns no result on Derby and Oracle: LIKE escape clause is never declared
**Describe the bug**

Fuzzy search on a term containing `_` returns no result on Derby and Oracle, because the generated `LIKE` predicate escapes `_` with a backslash but never declares an `ESCAPE` clause.

`generateLikeArgument` (config, namespace, AI resource and auth persist services) rewrites `_` to `\_` — this was introduced by #14442 to fix #14402, where `_` was interpreted as a single-character wildcard. That escaping only works on a database whose `LIKE` treats the backslash as the **default** escape character. MySQL and PostgreSQL do; Derby and Oracle do not — for them the backslash has no special meaning unless `ESCAPE` is declared, so `%nacos\_test%` matches the literal string `nacos\_test` instead of `nacos_test`.

#14442 added `ESCAPE '\'` to the Derby mappers it touched, but two gaps remain:

1. **Derby** (the default standalone datasource): `ConfigInfoMapper#findAllConfigInfo4Export` is an interface `default` that Derby does not override, so it still emits a bare `data_id LIKE ?`.
2. **Oracle**: no `LIKE` site declares the escape clause at all — `ConfigInfoMapper#findConfigInfoLike4PageCountRows`, `ConfigTagsRelationMapper#findConfigInfoLike4PageCountRows` and `ConfigInfoMapper#findAllConfigInfo4Export` are inherited defaults, and `ConfigInfoMapperByOracle#findConfigInfoLike4PageFetchRows` / `ConfigTagsRelationMapperByOracle#findConfigInfoLike4PageFetchRows` build the predicate as raw SQL. `AiResourceMapper` name/bizTags search is affected through the same shared default.

The failure is silent: the SQL stays syntactically valid and the parameter list stays consistent, the query simply matches nothing.

**Expected behavior**

Searching a config whose dataId contains `_` (for example `nacos_test`) returns that config on every supported datasource, and exporting configs filtered by such a dataId exports them.

**Actually behavior**

On Derby and Oracle the result set is empty.

**How to Reproduce**

1. Start a standalone Nacos server (Derby is the default datasource).
2. Publish a config with dataId `nacos_test`.
3. Open the console config list and search by dataId `nacos_test`, then export configs filtered by dataId `nacos_test`.
4. The export returns nothing; on Oracle the list query returns nothing as well.

The underlying semantics can be reproduced directly against Derby 10.14.2.0 (`jdbc:derby:memory:probe;create=true`) with rows `nacos_test`, `nacosXtest`, `nacos\_test` and the bound argument `%nacos\_test%`:

```
no ESCAPE    SELECT data_id FROM t WHERE data_id LIKE ?                 -> matches: [nacos\_test]
with ESCAPE  SELECT data_id FROM t WHERE data_id LIKE ? ESCAPE '\'      -> matches: [nacos_test]
```

Without the clause the intended row `nacos_test` is not matched at all.

 - OS: All
 - Version: current `develop` (3.3.0-SNAPSHOT), also affects the released 3.x line since #14442
 - Module: config / plugin (datasource dialect)
 - SDK: original

**Additional context**

The escape clause cannot be hardcoded into the shared mapper defaults: MySQL rejects `ESCAPE '\'` because the backslash escapes the closing quote of the string literal, so the clause has to be reported per dialect. This suggests exposing it on the datasource dialect mapper SPI (next to the existing `getFunction` hook) and returning it only for the dialects that need it, leaving the MySQL and PostgreSQL SQL unchanged.

Not covered by this report, but affected by the same root cause and worth a follow-up: the namespace queries in `core` and the user/role/permission fuzzy queries in the default auth plugin build raw SQL that is not routed through the dialect mappers, so they cannot declare the clause today.

wushiyuanmaimob:
Fixes #15710

Fuzzy search on a term containing `_` silently returns nothing on Derby and Oracle.

#14442 (fixing #14402) made `generateLikeArgument` rewrite `_` to `\_` so the wildcard is matched literally, and added `ESCAPE '\'` to the Derby mappers it touched. That escaping is only self-sufficient on a database whose `LIKE` treats the backslash as the **default** escape character. MySQL and PostgreSQL do; Derby and Oracle do not, so for them `%nacos\_test%` matches the literal text `nacos\_test` rather than `nacos_test`, and the query returns an empty result without any error.

Two gaps were left behind:

* **Derby** (the default standalone datasource) — `ConfigInfoMapper#findAllConfigInfo4Export` is an interface `default` that Derby does not override, so config export filtered by a dataId containing `_` exports nothing.
* **Oracle** — no `LIKE` site declares the clause: `ConfigInfoMapper#findConfigInfoLike4PageCountRows`, `ConfigTagsRelationMapper#findConfigInfoLike4PageCountRows` and `ConfigInfoMapper#findAllConfigInfo4Export` are inherited defaults, while `ConfigInfoMapperByOracle#findConfigInfoLike4PageFetchRows` and `ConfigTagsRelationMapperByOracle#findConfigInfoLike4PageFetchRows` build the predicate as raw SQL. `AiResourceMapper` name/bizTags search reaches Oracle through the same shared default.

The clause cannot simply be hardcoded into the shared defaults: MySQL rejects `ESCAPE '\'`, because the backslash escapes the closing quote of the string literal. It has to be reported per dialect, which is what this change does.

## Brief changelog

* `Mapper` — new `LIKE_ESCAPE_CLAUSE` constant and a `default String getLikeEscapeClause()` hook returning an empty string. It sits next to the existing `getFunction` dialect hook and, being a `default` method, is source compatible with out-of-tree dialect plugins.
* `WhereBuilder` — new `like(field, parameter, escapeClause)` overload; `like(field, parameter)` and `likeWithEscape(field, parameter)` now delegate to it and emit byte-identical SQL.
* Shared mapper defaults bound to escaped parameters now append the clause: `ConfigInfoMapper#findConfigInfoLike4PageCountRows`, `ConfigInfoMapper#findAllConfigInfo4Export`, `ConfigTagsRelationMapper#findConfigInfoLike4PageCountRows`, `AiResourceMapper#appendSingleAndCondition`.
* `AbstractMapperByDerby` and `AbstractMapperByOracle` override the hook, which covers every mapper of those dialects at once.
* `ConfigInfoMapperByOracle#findConfigInfoLike4PageFetchRows` and `ConfigTagsRelationMapperByOracle#findConfigInfoLike4PageFetchRows` append the clause to their own predicates.
* `AiResourceMapperByDerby#appendSingleAndCondition` is removed — it existed only to swap in `likeWithEscape`, which the shared default now does through the hook.
* `datasource-dialect-plugin-spec.md` (en + zh-cn) documents the rule.

MySQL and PostgreSQL resolve the hook to an empty string, so their generated SQL is unchanged — their existing exact-SQL assertions still pass untouched and act as the guard for that.

## Verifying this change

Derby 10.14.2.0 (`jdbc:derby:memory:probe;create=true`), rows `nacos_test`, `nacosXtest`, `nacos\_test`, bound argument `%nacos\_test%`:

```
no ESCAPE    SELECT data_id FROM t WHERE data_id LIKE ?              -> matches: [nacos\_test]
with ESCAPE  SELECT data_id FROM t WHERE data_id LIKE ? ESCAPE '\'   -> matches: [nacos_test]
```

Tests were written first and confirmed red against the unmodified code (1 failure in `ConfigInfoMapperByDerbyTest`, 5 in the Oracle mapper tests), then green after the change. `OracleMapperCoverageTest` had pinned the two defective statements, so its expectations are updated in the same commit.

Local runs, all green:

* `plugin/datasource` 39 tests, `nacos-datasource-plugin-derby` 77, `nacos-datasource-plugin-oracle` 81, `nacos-datasource-plugin-mysql` 78, `nacos-datasource-plugin-postgresql` 27, `nacos-datasource-plugin-base` 17
* `ai` 2093 tests; `config` 1420 tests (the single `Md5ComparatorDelegateTest` error is a local JDK 21 restriction on writing a `static final` field and reproduces identically on unmodified `develop`)
* `mvn spotless:apply` then `spotless:check checkstyle:check spotbugs:check apache-rat:check` on the three touched modules: 0 checkstyle violations, RAT unapproved 0

No API, HTTP endpoint or Java SDK contract changes, so `test/openapi-test` and `test/java-sdk-test` coverage is unaffected.

## Follow-up (not in this PR)

The namespace queries in `core` and the user/role/permission fuzzy queries in the default auth plugin call `generateLikeArgument` too, but build raw SQL that is not routed through the dialect mappers, so they cannot declare the clause today. Happy to open a separate issue for that once the direction here is agreed.

* [x] Run `mvn -B clean package apache-rat:check spotbugs:check -DskipTests` to make sure basic checks pass. Run `mvn clean install` to make sure unit-test pass. Run `mvn clean test-compile failsafe:integration-test`  to make sure integration-test pass.


KomachiSion (changes requested):
Thanks for the fix. The overall direction is correct, but there is one regression that should be addressed before merging.

`generateLikeArgument` currently escapes `_` but does not escape an existing backslash in the user input. After this PR appends `ESCAPE '\'`, a search value such as `C:\path` contains the invalid escape sequence `\p`:

- Oracle throws `ORA-01424`.
- Derby throws `SQLSTATE 22025`.

This can affect fuzzy searches on fields that may legally contain backslashes, especially config content and possibly AI resource tags.

Could we escape the escape character itself before escaping `_`?

1. `\` → `\\`
2. `_` → `\_`
3. `*` → `%`

The Config and AI implementations of `generateLikeArgument` should remain consistent.

It would also be helpful to add tests for:

- `nacos_test`
- `C:\path`
- a trailing `\`
- `a\_b`

Ideally, at least one Derby test should execute the generated SQL with a bound parameter. The current assertions mainly verify that the SQL contains the `ESCAPE` clause, so they cannot detect invalid parameter escape sequences.

KomachiSion (approved):

PR #15719, opened 2026-08-17
[ISSUE #15718] Escape every LIKE predicate in the embedded role search

### LINKED ISSUE #15718: Role search returns no result when the username filter contains an underscore
## Describe the bug

On a standalone Nacos server (embedded Derby storage, the default), searching roles with **both** the username filter and the role filter returns no result as soon as the username contains an underscore. Filtering by only one of the two works.

## Expected behavior

`_` is a literal character in a username, so searching `na_cos` should return the matching role rows, exactly as it does when the role filter is left empty.

## Actual behavior

The query returns zero rows.

## How to reproduce

1. Start Nacos in standalone mode (embedded Derby) with auth enabled.
2. Create a user `na_cos` and bind it to a role `ro_le`.
3. Open Console -> Permissions -> Roles.
4. Search with username `na_cos` **and** role `ro_le` -> no result.
5. Search with username `na_cos` only -> the row is returned.

## Root cause

`EmbeddedRolePersistServiceImpl#findRolesLike4Page` appends the escape clause once, after both `LIKE` predicates have been built:

```java
if (StringUtils.isNotBlank(username)) {
    where.append(" AND username LIKE ? ");
    params.add(generateLikeArgument(username));
}
if (StringUtils.isNotBlank(role)) {
    where.append(" AND role LIKE ? ");
    params.add(generateLikeArgument(role));
}
if (CollectionUtils.isNotEmpty(params)) {
    where.append(SQL_DERBY_ESCAPE_BACK_SLASH_FOR_LIKE);
}
```

which produces:

```sql
WHERE 1 = 1  AND username LIKE ?  AND role LIKE ?  ESCAPE '\'
```

In SQL, `ESCAPE` qualifies only the single `LIKE` predicate it immediately follows. The clause therefore applies to `role` alone, and the `username` predicate is left without one. Since `generateLikeArgument` has already rewritten `_` into `\_`, Derby matches the backslash literally and the predicate can never be satisfied.

When only one filter is set the single predicate happens to be the last one, so the clause lands correctly. That is why the defect only shows up when both filters are combined.

## Evidence

Executed against Derby 10.14.2.0 with rows `('na_cos', 'ro_le')` and `('naXcos', 'roXle')`, binding `%na\_cos%` and `%ro\_le%`:

| SQL | Result |
|-----|--------|
| `... AND username LIKE ? AND role LIKE ? ESCAPE '\'` | `rows=0` |
| `... AND username LIKE ? ESCAPE '\' AND role LIKE ? ESCAPE '\'` | `rows=1` (`na_cos/ro_le`) |

## Scope

Only `findRolesLike4Page` combines two fuzzy predicates. The user and permission searches in the same module build a single `LIKE` predicate each and already place the clause correctly, so they are unaffected.

## Environment

- Version: `develop` (3.3.0-SNAPSHOT)
- Module: auth (`plugin-default-impl/nacos-default-auth-plugin`)
- Storage: embedded (Derby)

wushiyuanmaimob:
Fixes #15718

On a standalone server (embedded Derby, the default storage), the console role search returns nothing when the username filter and the role filter are combined and the username contains an underscore.

`EmbeddedRolePersistServiceImpl#findRolesLike4Page` appended `ESCAPE '\'` once, after both `LIKE` predicates had been built:

```sql
WHERE 1 = 1  AND username LIKE ?  AND role LIKE ?  ESCAPE '\'
```

`ESCAPE` qualifies only the single `LIKE` predicate it immediately follows, so the clause applied to the role filter alone and the username filter was left without one. `generateLikeArgument` had already rewritten `_` into `\_`, so Derby matched the backslash literally and the predicate could never be satisfied.

When only one filter is set, that single predicate happens to be the last one and the clause lands correctly, which is why the defect only surfaces with both filters combined.

Only `findRolesLike4Page` builds two fuzzy predicates. The user and permission searches in the same module build a single `LIKE` each and already place the clause correctly, so they were not changed.

## Brief changelog

- `EmbeddedRolePersistServiceImpl#findRolesLike4Page` appends the escape clause to each `LIKE` predicate, the same way the user and permission searches in this module already do. The now-unused `CollectionUtils` import is dropped.
- New `RoleSearchLikeEscapeDerbyTest` executes the SQL the service generates against a real in-memory Derby database with the bound parameters the service produces. Asserting the SQL text alone cannot prove which predicate the clause qualifies, so the query is actually run.
- `EmbeddedRolePersistServiceImplTest` gains an exact-SQL assertion for the two-filter case.
- `derby` is added as a `test` scope dependency of the module so the execution test can run. It is already part of the shipped plugin bundle, so no new runtime dependency is introduced.
- The default auth plugin spec (en + zh-cn) documents the escaping contract: the clause qualifies the predicate it immediately follows and must be repeated after every fuzzy filter.

## Verifying this change

Both new tests were confirmed RED before the fix:

- `EmbeddedRolePersistServiceImplTest#testFindRolesLike4PageEscapesEveryLikePredicate`
  `expected: <... AND username LIKE ? ESCAPE '\' AND role LIKE ? ESCAPE '\'> but was: <... AND username LIKE ? AND role LIKE ? ESCAPE '\'>`
- `RoleSearchLikeEscapeDerbyTest#testFindRolesLike4PageMatchesUnderscoreWhenBothFiltersAreUsed`
  `expected: <1> but was: <0>` on real Derby 10.14.2.0

The second Derby case (`username` filter only) passed before and after, which pins the defect to the multi-predicate path.

Independent probe against Derby 10.14.2.0, rows `('na_cos', 'ro_le')` and `('naXcos', 'roXle')`, binding `%na\_cos%` and `%ro\_le%`:

| SQL | Result |
|-----|--------|
| `... AND username LIKE ? AND role LIKE ? ESCAPE '\'` | `rows=0` |
| `... AND username LIKE ? ESCAPE '\' AND role LIKE ? ESCAPE '\'` | `rows=1` (`na_cos/ro_le`) |

Gates on `plugin-default-impl/nacos-default-auth-plugin`: 392 tests pass, `spotless:check`, `checkstyle:check` (0 violations), `spotbugs:check` and `apache-rat:check` (Unapproved: 0) all clean.

## Follow-up

The external (non-embedded) counterparts escape `_` but never declare the clause at all, and `findUserLikeUsername` / `findRolesLikeRoleName` bind `%value%` without escaping it, so the same keyword behaves differently between the dropdown search and the paged search. That needs a dialect seam in this plugin rather than a hardcoded clause, so it is deliberately left out of this PR. Happy to follow up if you agree on the direction.


KomachiSion:
Thanks for the fix. I noticed that `findUserLikeUsername` and `findRolesLikeRoleName` still bind `%value%` directly without escaping `_`, while the paged search now treats `_` as a literal character.

Could you clarify why these list/search paths are not fixed together in this PR? Keeping them unchanged means the same keyword may produce different results between the dropdown search and the paged search. Would it be better to unify the behavior in this PR, or at least create a follow-up issue to track it?

KomachiSion:
@wushiyuanmaimob Hi, Would you mind provider your dingtalk number? we would like to invite you into nacos contributors group.

KomachiSion (approved):

PR #15758, opened 2026-08-23
[ISSUE #15747] Write cluster.conf atomically

### LINKED ISSUE #15747: Concurrent cluster.conf writes during Kubernetes Pod restart corrupt Raft membership
## Description

In a Nacos 2.x cluster deployed on Kubernetes, recreating a Pod may corrupt `cluster.conf` and cause healthy Raft peers to be removed.

The resulting Raft membership becomes inconsistent across Nacos nodes. Service discovery then returns different instance lists depending on which Nacos node handles the request, causing service invocation failures.

We encountered this issue in both production and development environments, and it caused a production incident.

## Observed behavior

After recreating `nacos-1`, the following corrupted `cluster.conf` was found on `nacos-0`:

```text
#2026-08-21T00:33:53.703
nacos-0.nacos-hs.naconacos-1.nacos-hs.nacos.svc.cluster.local:8848
nacos-0.nacos-hs.nacos.svc.cluster.local:8848
nacos-2.nacos-hs.nacos.svc.cluster.local:8848
```

The addresses of `nacos-0` and `nacos-1` were merged, and `nacos-0` was duplicated. Therefore, `nacos-1` was not present as a valid member entry.

The Raft configuration then changed from three members to two. The removed node repeatedly logged:

```text
WARN Node <...> can't do preVote as it is not in conf
```

In the same incident, the ephemeral service `pay-account` had instance `172.16.35.111:8080` on two Nacos nodes, but no instance on the third node.

As a result, clients received different service discovery results depending on the Nacos node they connected to, and service invocations failed when the instance list was missing or incorrect.

## Steps to reproduce

Environment:

- Nacos 2.5.1
- Three-node Kubernetes StatefulSet
- Kubernetes headless Service
- `nacos/nacos-peer-finder-plugin:1.1`
- MySQL

Steps:

1. Confirm that all three nodes are present in `cluster.conf`.
2. Recreate one Nacos Pod:

   ```bash
   kubectl delete pod nacos-1 -n nacos
   ```

3. Wait for the Pod to restart and the headless Service endpoints to change.
4. Check `cluster.conf`, Raft membership, and ephemeral service instances on all nodes.

We reproduced the issue twice in development. Because it is a timing race, it may not occur on every restart.

## Root cause

`cluster.conf` has two independent non-atomic writers:

1. Nacos runtime:

   ```text
   MemberUtil.syncToFile()
     -> EnvUtil.writeClusterConf()
     -> DiskUtils.writeFile(file, bytes, false)
   ```

2. peer-finder `on-change.sh`:

   ```bash
   echo ... > "${CLUSTER_CONF}"
   ```

A Pod restart changes the headless Service endpoints and triggers peer-finder `on-change.sh`. At the same time, Nacos may call `syncToFile()` because its member list changes.

Both writers truncate and rewrite the same file. If their writes overlap, `cluster.conf` can be corrupted. Nacos then uses the invalid member list for dynamic Raft membership changes, which may remove healthy peers.

## Expected behavior

A Pod restart should not corrupt `cluster.conf`, make Raft membership inconsistent, return incorrect service discovery data, or cause service invocation failures.

## Suggested fix

1. Make `writeClusterConf()` atomic by writing to a temporary file in the same directory and replacing the target with `Files.move(..., ATOMIC_MOVE)`.
2. Avoid multiple runtime writers in Kubernetes: use peer-finder only for startup initialization, and let Nacos manage runtime member changes.

Zhengcy05:
Fixes #15747 

This PR prevents Nacos from truncating the published `cluster.conf` while constructing a new cluster member snapshot.

Previously, `EnvUtil.writeClusterConf` wrote directly to `cluster.conf`. When an external writer such as Kubernetes peer-finder updated the same file concurrently, the writes could overlap and expose partial or interleaved content.

The Nacos-side write path now creates a complete snapshot in a temporary file located in the same directory and then replaces `cluster.conf`. It requests an atomic move when supported by the filesystem and falls back to a regular replacement otherwise.

## Brief changelog

- Write cluster member snapshots to a same-directory temporary file before publishing them.
- Use an atomic move when supported and clean up temporary files after success or failure.
- Add regression coverage for normal replacement, concurrent writes through an old file handle, and temporary-file cleanup after publication failure.
- Document the atomic publication requirement in the English and Chinese cluster membership specifications.

## Verifying this change

- `mvn -pl sys spotless:check`
- `mvn -pl sys -Dtest=EnvUtilTest test`
  - 41 tests passed with no failures or errors.
- `mvn -pl sys test`
  - All `sys` module tests passed.
- `mvn -B clean compile apache-rat:check checkstyle:check spotbugs:check spotless:check -DskipTests`
  - All 62 Maven modules passed.

* [x] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [[test module](https://github.com/alibaba/nacos/tree/master/test)](https://github.com/alibaba/nacos/tree/master/test).
* [ ] Run `mvn -B clean package apache-rat:check spotbugs:check -DskipTests` to make sure basic checks pass. Run `mvn clean install` to make sure unit-test pass. Run `mvn clean test-compile failsafe:integration-test` to make sure integration-test pass.

### DIFF
--- a/specs/en/design/foundation-cluster-membership-spec.md
+++ b/specs/en/design/foundation-cluster-membership-spec.md
@@ -115,6 +115,10 @@ Change rules:
   full member-list change;
 - topology changes are written to `cluster.conf` for compatibility by
   `MemberUtil.syncToFile`;
+- compatibility persistence must write a complete member-list snapshot to a
+  temporary file in the same directory before replacing `cluster.conf`, using
+  an atomic move when the filesystem supports it; it must not truncate the
+  published file while constructing a new snapshot;
 - `MembersChangeEvent` is published only when the effective topology changes or
   basic member metadata changes.
 
--- a/specs/zh-cn/design/foundation-cluster-membership-spec.md
+++ b/specs/zh-cn/design/foundation-cluster-membership-spec.md
@@ -102,6 +102,8 @@ Lookup 选择规则：
 - 当同一地址仍然存在于新列表时，应保留已有 member 的元数据和能力；
 - `memberJoin` 和 `memberLeave` 是便捷操作，最终都解析为完整 member-list 变更；
 - topology 变化会通过 `MemberUtil.syncToFile` 写入 `cluster.conf` 以保持兼容；
+- 兼容落盘必须先在同一目录的临时文件中写完完整 member-list 快照，再替换 `cluster.conf`；文件系统支持时
+  必须使用原子移动，构造新快照期间不得截断已经发布的文件；
 - 只有有效 topology 变化或 member 基础元数据变化时，才发布 `MembersChangeEvent`。
 
 `ServerMemberManager.update(Member)` 用于更新已存在 member 的元数据/状态。它不得新增未知 member。
--- a/sys/src/main/java/com/alibaba/nacos/sys/env/EnvUtil.java
+++ b/sys/src/main/java/com/alibaba/nacos/sys/env/EnvUtil.java
@@ -21,7 +21,6 @@ import com.alibaba.nacos.common.utils.IoUtils;
 import com.alibaba.nacos.common.utils.StringUtils;
 import com.alibaba.nacos.common.utils.ThreadUtils;
 import com.alibaba.nacos.plugin.environment.CustomEnvironmentPluginManager;
-import com.alibaba.nacos.sys.utils.DiskUtils;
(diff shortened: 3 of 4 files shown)


KomachiSion:
discuss in issue first, thanks.

PR #15768, opened 2026-08-24
[ISSUE #15763] Limit compressed IPv6 mixed addresses to five explicit blocks

### LINKED ISSUE #15763: [BUG] isIpv6MixedAddress accepts IPv6 mixed addresses with more than six 16-bit blocks
**Describe the bug**

`InetAddressValidator.isIpv6MixedAddress` accepts IPv6 mixed addresses whose IPv6
part contains more than six 16-bit blocks.

In the IPv6 mixed notation the IPv6 part can hold at most 96 bits (six 16-bit
blocks), because the trailing IPv4 part occupies the remaining 32 bits of the
128-bit address. `IPV6_MIXED_COMPRESSED_REGEX` uses a `*` quantifier for the
group after `::`:

```java
private static final Pattern IPV6_MIXED_COMPRESSED_REGEX = Pattern.compile(
    "^" + "(" + "(?:[0-9A-Fa-f]{1,4}" + "(?::[0-9A-Fa-f]{1,4})*)?" + ")" + "::" + "("
        + "(?:[0-9A-Fa-f]{1,4}:"
        + "(?:[0-9A-Fa-f]{1,4}:)*)?" + ")" + "$");
```

so there is no upper bound on the number of blocks, and the address is reported
as a valid IPv6 address.

**Expected behavior**

`isIpv6MixedAddress("::1:2:3:4:5:6:7:8:9:172.12.55.18")` returns `false`, since
the address is longer than 128 bits. `java.net.InetAddress.getByName` also
rejects it.

**Actually behavior**

It returns `true`, and `isIpv6Address` therefore also returns `true`.

**How to Reproduce**

```java
// all three return true today, expected false
InetAddressValidator.isIpv6MixedAddress("::1:2:3:4:5:6:7:172.12.55.18");
InetAddressValidator.isIpv6MixedAddress("::1:2:3:4:5:6:7:8:9:172.12.55.18");
InetAddressValidator.isIpv6MixedAddress("::1:2:3:4:5:6:7:8:9:a:b:c:172.12.55.18");

// valid addresses, correctly true, listed for comparison
InetAddressValidator.isIpv6MixedAddress("0:0:0:0:0:0:172.12.55.18");
InetAddressValidator.isIpv6MixedAddress("::172.12.55.18");
```

Comparing the same inputs against `InetAddress.getByName`, the JDK rejects all
three of the first group.

 - OS: macOS 26.5.2 (arm64)
 - Version: `develop` branch at commit `fe25052`, nacos 3.3.0-SNAPSHOT
 - Module: common (`InetAddressValidator`)
 - SDK: original

**Additional context**

`isIpv6Address` is reached from `InternetAddressUtil.isIpv6` and
`InternetAddressUtil.isIp`, which are used by
`HealthCheckAddressValidator`, `McpEndpointAccessValidator` and
`ParamUtils`.

This is a validation strictness issue rather than an exploitable one: in
`McpEndpointAccessValidator` the address is afterwards passed to
`InetAddress.getByName`, which rejects it, so the over-permissive result is
caught by the later parsing step. The two layers just disagree about what is a
valid address.

The uncompressed branch (`IPV6_MIXED_UNCOMPRESSED_REGEX`) is already bounded to
six blocks and is not affected.

I have a patch and a regression test ready and would be happy to open a PR for
this.

GerardGao:

Fixes #15763. Follow-up to #15764, addressing the remaining boundary case raised
in review.

#15764 bounded the IPv6 part of a mixed address to six blocks. That is correct
for the uncompressed form, but a compressed IPv6 part may carry at most five
explicit blocks, because `::` stands for at least one omitted 16-bit block.
`::1:2:3:4:5:6:172.12.55.18` was therefore still accepted.

## Brief changelog

- Apply the six-block limit only to the uncompressed form, which
  `IPV6_MIXED_UNCOMPRESSED_REGEX` already enforces, and limit a compressed IPv6
  part to `MAX_COMPRESSED_MIXED_IPV6_BLOCKS` (5) explicit blocks.
- `InetAddressValidatorTest`: add `isIpv6MixedAddressBlockLimitDependsOnCompression`.

This also covers the same case with `::` in any position, since the root cause is
the total block count rather than where `::` appears. Generating 102 inputs by
varying the block count and the position of `::` and comparing the result with
`java.net.InetAddress`, the current `develop` accepts seven invalid addresses:

```
::1:2:3:4:5:6:172.12.55.18
1:2:3:4:5:6::172.12.55.18
1::1:2:3:4:5:172.12.55.18
1:2::1:2:3:4:172.12.55.18
1:2:3::1:2:3:172.12.55.18
1:2:3:4::1:2:172.12.55.18
1:2:3:4:5::1:172.12.55.18
```

After this change all 102 inputs agree with the JDK, with no address newly
rejected.

For the record, my note in #15764 about `{0,5}` causing a regression was wrong:
`::1:2:3:4:5:6:` is genuinely invalid, so rejecting it was the correct behaviour
and I had compared it against a wrong expectation. Checking the block count in
the method still seems clearer than encoding it in the regex, but the reasoning I
gave for it was not.

## Verifying this change

- `isIpv6MixedAddressBlockLimitDependsOnCompression` fails before this change
  (`expected: <false> but was: <true>`) and passes after it.
- `mvn -pl common -am test`: 956 tests, 0 failures.
- `mvn -B clean compile apache-rat:check checkstyle:check spotless:check -DskipTests`:
  BUILD SUCCESS.
- Cross-checked the 102 generated inputs against `java.net.InetAddress.getByName`.

Valid forms still pass: `1:2:3:4:5:6:172.12.55.18` (six explicit, uncompressed),
`0:0:0:0:0:0:172.12.55.18`, `::172.12.55.18`, `::1:2:3:4:5:172.12.55.18`,
`1:2:3:4:5::172.12.55.18` and `1:2:3::4:5:172.12.55.18`.

### DIFF
--- a/common/src/main/java/com/alibaba/nacos/common/utils/InetAddressValidator.java
+++ b/common/src/main/java/com/alibaba/nacos/common/utils/InetAddressValidator.java
@@ -42,6 +42,12 @@ public class InetAddressValidator {
      */
     private static final int MAX_MIXED_IPV6_BLOCKS = 6;
     
+    /**
+     * A compressed IPv6 part may hold at most five explicit blocks, because "::" stands for at
+     * least one omitted 16-bit block.
+     */
+    private static final int MAX_COMPRESSED_MIXED_IPV6_BLOCKS = MAX_MIXED_IPV6_BLOCKS - 1;
+    
     private static final Pattern IPV4_PATTERN = Pattern
         .compile("^" + "(25[0-5]|2[0-4]\\d|[0-1]?\\d?\\d)"
             + "(\\.(25[0-5]|2[0-4]\\d|[0-1]?\\d?\\d)){3}" + "$");
@@ -135,8 +141,9 @@ public class InetAddressValidator {
             IPV6_MIXED_UNCOMPRESSED_REGEX.matcher(ipV6Part).matches();
         boolean ipV6CompressedDetected = IPV6_MIXED_COMPRESSED_REGEX.matcher(ipV6Part).matches();
         
-        return ipv4PartValid && (ipV6UncompressedDetected || ipV6CompressedDetected)
-            && countHexBlocks(ipV6Part) <= MAX_MIXED_IPV6_BLOCKS;
+        return ipv4PartValid && (ipV6UncompressedDetected
+            || (ipV6CompressedDetected
+                && countHexBlocks(ipV6Part) <= MAX_COMPRESSED_MIXED_IPV6_BLOCKS));
     }
     
     /**
--- a/common/src/test/java/com/alibaba/nacos/common/utils/InetAddressValidatorTest.java
+++ b/common/src/test/java/com/alibaba/nacos/common/utils/InetAddressValidatorTest.java
@@ -57,6 +57,22 @@ class InetAddressValidatorTest {
     }
     
     @Test
+    void isIpv6MixedAddressBlockLimitDependsOnCompression() {
+        // Without "::" all six blocks are written explicitly.
+        assertTrue(InetAddressValidator.isIpv6MixedAddress("1:2:3:4:5:6:172.12.55.18"));
+        // "::" stands for at least one omitted block, so a compressed IPv6 part carries at
+        // most five explicit blocks, wherever "::" appears.
+        assertTrue(InetAddressValidator.isIpv6MixedAddress("::1:2:3:4:5:172.12.55.18"));
+        assertTrue(InetAddressValidator.isIpv6MixedAddress("1:2:3:4:5::172.12.55.18"));
+        assertFalse(InetAddressValidator.isIpv6MixedAddress("::1:2:3:4:5:6:172.12.55.18"));
+        assertFalse(InetAddressValidator.isIpv6MixedAddress("1:2:3:4:5:6::172.12.55.18"));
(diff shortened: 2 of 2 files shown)


nacos-community (approved):
### Summary
LGTM. Correctly fixes the remaining boundary case from #15764: the six-block limit now applies to the uncompressed mixed form only (already enforced by `IPV6_MIXED_UNCOMPRESSED_REGEX`), while a compressed IPv6 part is limited to five explicit blocks because `::` must stand for at least one omitted block. Verified against the current `InetAddressValidator` implementation: `countHexBlocks` counts non-empty blocks only, so inputs like `::1:2:3:4:5:6:172.12.55.18` (6 explicit + `::` + IPv4 tail ≥ 9 blocks) are now correctly rejected, while valid forms such as `1:2:3:4:5::172.12.55.18` and `::172.12.55.18` (handled by the `::` special case) are unaffected. The added test cases match the JDK `InetAddress` behavior described in the PR, and the change stays Java 8 compatible. Thanks for the thorough 102-input cross-check and the correction of the earlier `{0,5}` reasoning.

---

KomachiSion (approved):

PR #15784, opened 2026-08-28
[ISSUE #15477] Defer persistent health checks until service metadata is ready

### LINKED ISSUE #15477: [Bug]3.2.2版本集群模式下创建 SSE HTTP-to-MCP Server 时，持久化实例为不健康状态
**问题描述**
在 **Nacos 3.2.2**集群的 MCP 管理页面中创建 MCP Server，配置如下：

- 协议类型：SSE
- 服务类型：HTTP 转 MCP 服务
- 后端服务：选择“新建服务”
- 后端地址：填写一个固定且可正常访问的地址

使用完全相同的后端地址和配置重复创建 MCP Server 时，Nacos 自动创建的 mcp-endpoints 持久化服务实例健康状态不稳定：

- 有时创建后为“健康”
- 有时创建后为“不健康”

该现象在后端地址、端口、协议及服务可用性均未变化的情况下出现。
根据当前实现，Direct Endpoint 创建的服务集群健康检查类型应为 NONE，新注册实例的默认健康状态应为 true。因此，实例健康状态不应依赖后端地址的 TCP/HTTP 可达性，也不应随机变成不健康。

Image

Zhengcy05:

This PR fixes a startup race that can incorrectly mark persistent MCP endpoint instances as unhealthy.

Persistent instances and service metadata are recovered by two independent Raft groups. During persistent client snapshot recovery, `HealthCheckTaskV2` may start before the local `naming_service_metadata` group has completed recovery. Missing metadata is then interpreted as the default TCP health checker, and a failed check may persist an incorrect unhealthy state.

This change adds an interceptor that defers persistent instance health checks until the service metadata Raft group is ready. It only applies to `HealthCheckTaskV2` and preserves the existing fallback behavior after metadata recovery completes.

Fixes #15477.

## Brief changelog

- Add `ServiceMetadataReadyInterceptor` to gate persistent health checks on service metadata readiness.
- Register the interceptor through the existing health-check SPI.
- Add unit tests for unavailable, ready, and failed readiness checks.
- Keep ephemeral heartbeat health checks unaffected.
- Document the recovery readiness requirement in the English and Chinese naming specifications.

## Verifying this change

The following checks passed:

- `mvn -pl naming spotless:apply`
- `mvn -pl naming spotless:check`
- `mvn -pl naming -Dtest=ServiceMetadataReadyInterceptorTest,HealthCheckTaskInterceptWrapperTest,HealthCheckTaskV2Test test`
  - 21 tests passed.
- `mvn -B clean compile apache-rat:check checkstyle:check spotbugs:check spotless:check -DskipTests`
  - All 61 modules passed.

- [x] Make sure there is a Github issue filed for the change (usually before you start working on it). Trivial changes like typos do not require a Github issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.
- [x] Format the pull request title like `[ISSUE #123] Fix UnknownException when host config not exist`. Each commit in the pull request should have a meaningful subject line and body.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [[test module](https://github.com/alibaba/nacos/tree/master/test)](https://github.com/alibaba/nacos/tree/master/test).
- [ ] Run `mvn spotless:apply` to format your code. Run `mvn -B clean compile apache-rat:check checkstyle:check spotbugs:check spotless:check -DskipTests` to make sure basic checks pass. Run `mvn clean install` to make sure unit-test pass. Run `mvn clean test-compile failsafe:integration-test`  to make sure integration-test pass.


KomachiSion (changes requested):
Thanks for the update. I traced the startup and snapshot recovery paths again.

The current readiness check based on `LEADER_META_DATA` is not sufficient. It only shows that a leader is known for the Raft group; it does not mean that this node has completed loading the local snapshot or that the local naming state is ready.

Could you please adjust the implementation as follows?

1. Keep the current `HealthCheckTaskV2` type filter. In the production path, ephemeral clients use `ClientBeatCheckTaskV2`, while persistent clients use `HealthCheckTaskV2`, so an additional client-type check is unnecessary.

2. Add a monotonic `snapshotLoaded` state for both related CP processors:
   - `PersistentClientOperationServiceImpl` / `naming_persistent_service_v2`
   - `ServiceMetadataProcessor` / `naming_service_metadata`

   The state should initially be `false` and become `true` only after the corresponding `onSnapshotLoad()` completes successfully. Once it becomes `true`, it should not be reset during the process lifetime.

3. Please expose the state from the two Spring-managed processors rather than keeping it only in a temporary `SnapshotOperation` instance. `loadSnapshotOperate()` is called more than once and currently creates a new operation each time, so storing readiness only in one returned operation can cause the interceptor to observe a different instance.

4. The interceptor should quickly skip the health check while either snapshot is not loaded. Please log every skipped task and include at least:
   - task id
   - persistent client snapshot state
   - service metadata snapshot state
   - application startup state

5. To avoid permanently disabling health checks when no local snapshot exists, use `ApplicationUtils.isStarted()` as a fail-open condition. Once both snapshots are loaded, or the application has completed startup, set a monotonic `healthCheckReleased` flag and never close the gate again.

   Since `ApplicationUtils.started` is written by the startup thread and read by health-check threads, please make it `volatile` or replace it with an `AtomicBoolean`.

Please also add tests covering:

- either snapshot is not loaded before application startup: intercepted;
- both snapshots are loaded: passed;
- no snapshot is loaded but application startup has completed: passed;
- once released, the gate never closes again;
- snapshot load failure does not mark the snapshot as loaded;
- ephemeral heartbeat tasks remain unaffected.

This change addresses the local startup snapshot-ordering race described by this PR. `ApplicationUtils.isStarted()` should be documented as a fail-open fallback, not as a general guarantee that both Raft groups have caught up to their leaders after startup.

naming/src/main/java/com/alibaba/nacos/naming/healthcheck/interceptor/ServiceMetadataReadyInterceptor.java
```diff
@@ -0,0 +1,83 @@
...
+        boolean applicationStarted = ApplicationUtils.isStarted();
+        if (applicationStarted) {
+            healthCheckReleased = true;
+            return false;
+        }
+        try {
+            PersistentClientOperationServiceImpl persistentClientProcessor =
+                ApplicationUtils.getBean(PersistentClientOperationServiceImpl.class);
+            ServiceMetadataProcessor serviceMetadataProcessor =
+                ApplicationUtils.getBean(ServiceMetadataProcessor.class);
```
KomachiSion: ApplicationUtils.getBean

frequently will cost much cpu resource. 

Maybe it can be cache it after the first time get.

naming/src/main/java/com/alibaba/nacos/naming/healthcheck/interceptor/ServiceMetadataReadyInterceptor.java
```diff
@@ -0,0 +1,83 @@
...
+            return false;
+        }
+        boolean applicationStarted = ApplicationUtils.isStarted();
+        if (applicationStarted) {
+            healthCheckReleased = true;
+            return false;
+        }
+        try {
+            PersistentClientOperationServiceImpl persistentClientProcessor =
+                ApplicationUtils.getBean(PersistentClientOperationServiceImpl.class);
```
KomachiSion: ApplicationUtils.getBean

frequently will cost much cpu resource.

Maybe it can be cache it after the first time get.

KomachiSion (changes requested):
These is only one problem for PR.

naming/src/main/java/com/alibaba/nacos/naming/healthcheck/interceptor/ServiceMetadataReadyInterceptor.java
Zhengcy05: Done, I've switched to lazy loading.

naming/src/main/java/com/alibaba/nacos/naming/healthcheck/interceptor/ServiceMetadataReadyInterceptor.java
Zhengcy05: Done, I've switched to lazy loading.

KomachiSion (approved):

Zhengcy05:
By the way, Does the codecov must be 100%, or is it sufficient to reach a certain threshold.

KomachiSion:
> By the way, Does the codecov must be 100%, or is it sufficient to reach a certain threshold.

Best to improve the coverage to 100%

PR #15787, opened 2026-08-29
Add admin authentication to server state admin API

Pyolar:

GET `/nacos/v3/admin/core/state` returns the server runtime state (version, startup mode, auth switches, quotas, etc.) without any authentication. Other endpoints under `/v3/admin/*` are protected by the AuthAdminFilter when `nacos.core.auth.admin.enabled=true` (default), but `ServerStateController#serverState` has no `@Secured` annotation, so the auth filter skips it and the state fingerprint is readable by anyone who can reach the server port.

## Brief changelog

- Add `@Secured` (ADMIN_API, read) to `ServerStateController#serverState`, consistent with other core admin controllers
- Keep `/liveness` and `/readiness` anonymous: the client SDK health check (`NamingHttpClientProxy`) and Kubernetes probes depend on them

## Verifying this change

- Extended `SecuredMetadataTest`: `serverState` must be secured with ADMIN_API; `liveness`/`readiness` must stay annotation-free.
- Manually verified on a local 3.2.4 standalone build: anonymous `GET /nacos/v3/admin/core/state` returns 403 after the change (200 before), `/liveness` stays 200, and the console login page (`/v3/console/server/state`) is unaffected.
- `mvn -B package -DskipTests` passes.

### DIFF
--- a/core/src/main/java/com/alibaba/nacos/core/controller/v3/ServerStateController.java
+++ b/core/src/main/java/com/alibaba/nacos/core/controller/v3/ServerStateController.java
@@ -18,11 +18,15 @@ package com.alibaba.nacos.core.controller.v3;
 
 import com.alibaba.nacos.api.annotation.Since;
 import com.alibaba.nacos.api.annotation.NacosApi;
+import com.alibaba.nacos.api.common.ApiType;
 import com.alibaba.nacos.api.exception.NacosException;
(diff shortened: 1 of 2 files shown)


KomachiSion:
The API is defined as `public` API, please see the API document.

If you want to add auth, please submit issue to discussion first.

PR #15789, opened 2026-08-30
Add network boundary check to legacy MCP URL import

Pyolar:

The legacy MCP import path (`McpExternalDataAdaptor`) fetches user provided URLs with its own HTTP client. Unlike the unified import plugins, whose `DefaultImportHttpClient` enforces a network boundary (scheme check, private/local target rejection), this legacy fetch runs without such a check, so an imported URL pointing at loopback/private/link-local targets is fetched by the server when `nacos.ai.resource.import.allow-user-url` is enabled. The client also follows redirects, so a public URL can lead the fetch to a blocked target.

## Brief changelog

- Add `checkUserEndpoint` to `AiResourceImportSecurityGuard`: resolves the URL host and rejects local/private targets (loopback, site-local, link-local, multicast, unique-local IPv6), consistent with `DefaultImportHttpClient` semantics
- Call the check before any fetch in `McpExternalDataAdaptor#fetchUrlPage`
- Stop following redirects (`Redirect.NEVER`) in the adaptor's HTTP client

## Verifying this change

- Extended `AiResourceImportSecurityGuardTest` with user-endpoint cases (local/private/link-local/multicast/unique-local rejection, public targets accepted, non-http scheme rejected). Existing `McpExternalDataAdaptorTest` and `McpLegacyImportAdapterTest` pass unchanged.
- `mvn -B test -pl ai -am` passes.

### DIFF
--- a/ai/src/main/java/com/alibaba/nacos/ai/importer/security/AiResourceImportSecurityGuard.java
+++ b/ai/src/main/java/com/alibaba/nacos/ai/importer/security/AiResourceImportSecurityGuard.java
@@ -19,9 +19,15 @@ package com.alibaba.nacos.ai.importer.security;
 import com.alibaba.nacos.api.exception.NacosException;
 import com.alibaba.nacos.api.exception.api.NacosApiException;
 import com.alibaba.nacos.api.model.v2.ErrorCode;
+import com.alibaba.nacos.common.utils.InternetAddressUtil;
 import com.alibaba.nacos.common.utils.StringUtils;
 import com.alibaba.nacos.plugin.ai.importer.model.AiResourceImportArtifact;
 import org.springframework.stereotype.Service;
+
+import java.net.InetAddress;
+import java.net.URI;
+import java.net.UnknownHostException;
+import java.util.Locale;
 
 /**
  * Central guard for import artifacts crossing the plugin boundary.
@@ -31,6 +37,14 @@ import org.springframework.stereotype.Service;
  */
 @Service
 public class AiResourceImportSecurityGuard {
+    
+    private static final String HTTPS_SCHEME = "https";
+    
+    private static final String HTTP_SCHEME = "http";
+    
+    private static final String LOCALHOST = "localhost";
+    
+    private static final String LOCALHOST_SUFFIX = ".localhost";
     
     /**
      * Check artifact type and size before validation or import.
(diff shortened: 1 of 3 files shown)


KomachiSion:
Already add policy for mcp tool import feature.

https://github.com/alibaba/nacos/pull/15749

Is it enough?

Pyolar:
@KomachiSion Thanks for pointing this out. I checked #15749 (and its develop counterpart #15746).

Those changes protect the Console MCP **tool import** path through `ConsoleMcpController` / `McpEndpointAccessValidator`.

This PR targets a different, deprecated-but-still-available MCP **server import** path: `McpServerImportService -> McpExternalDataAdaptor#fetchUrlPage`. That adaptor uses its own `HttpClient` and does not go through the policy introduced by #15746/#15749, so a user-provided registry URL can still reach private/local targets through this legacy path.

The legacy path is planned for removal in 3.4.0, but while it remains available I think it still needs the same outbound network boundary protection.

I can also reuse/inject the existing security component more cleanly if you prefer.

KomachiSion:
> @KomachiSion Thanks for pointing this out. I checked #15749 (and its develop counterpart #15746).
> 
> Those changes protect the Console MCP **tool import** path through `ConsoleMcpController` / `McpEndpointAccessValidator`.
> 
> This PR targets a different, deprecated-but-still-available MCP **server import** path: `McpServerImportService -> McpExternalDataAdaptor#fetchUrlPage`. That adaptor uses its own `HttpClient` and does not go through the policy introduced by #15746/#15749, so a user-provided registry URL can still reach private/local targets through this legacy path.
> 
> The legacy path is planned for removal in 3.4.0, but while it remains available I think it still needs the same outbound network boundary protection.
> 
> I can also reuse/inject the existing security component more cleanly if you prefer.

Thanks for your contribution, After 3.3 version, the old way to import mcpserver is disabled default, and use new way to import mcp server, I think it's enough for this problem. Most of users don't need enabled the old risk import way to do and no need do more enhancement for this removal parts. How do you think about?

Pyolar:
> > @KomachiSion Thanks for pointing this out. I checked #15749 (and its develop counterpart #15746).
> > Those changes protect the Console MCP **tool import** path through `ConsoleMcpController` / `McpEndpointAccessValidator`.
> > This PR targets a different, deprecated-but-still-available MCP **server import** path: `McpServerImportService -> McpExternalDataAdaptor#fetchUrlPage`. That adaptor uses its own `HttpClient` and does not go through the policy introduced by #15746/#15749, so a user-provided registry URL can still reach private/local targets through this legacy path.
> > The legacy path is planned for removal in 3.4.0, but while it remains available I think it still needs the same outbound network boundary protection.
> > I can also reuse/inject the existing security component more cleanly if you prefer.
> 
> Thanks for your contribution, After 3.3 version, the old way to import mcpserver is disabled default, and use new way to import mcp server, I think it's enough for this problem. Most of users don't need enabled the old risk import way to do and no need do more enhancement for this removal parts. How do you think about?

Thanks for the clarification. That makes sense.

Since the legacy MCP server import path is disabled by default after 3.3, requires explicit operator opt-in to allow user-provided URLs, and is already planned for removal in 3.4.0, I agree that adding further hardening to this deprecated path may not be necessary.

I'll close this PR. Thanks for reviewing and explaining the migration plan.

PR #15212, opened 2026-05-21
[ISSUE #8354] Make prometheus client optional in nacos-client SDK

EvanYao826:
## What does this PR do?

Fixes [#8354](https://github.com/alibaba/nacos/issues/8354): Make the nacos-client SDK lighter by making the prometheus dependency optional.

## Problem

`nacos-client` has a hard dependency on `io.prometheus:simpleclient`. Users who don't need metrics monitoring still pull in this dependency, which adds to the SDK footprint and can cause version conflicts.

## Solution

1. Mark `io.prometheus:simpleclient` as `<optional>true</optional>` in `client/pom.xml`
2. Refactor `MetricsMonitor` to detect prometheus availability at runtime via `Class.forName()`. When prometheus is not on the classpath, all monitoring operations become no-ops.
3. Extract prometheus API calls into `PrometheusMetricsHelper` (only loaded when prometheus is available)
4. Replace direct `Histogram.Child` usage in `MetricsHttpAgent` with a `MetricsTimer` interface
5. Update all callers (`ClientWorker`, `ServiceInfoHolder`, `NamingGrpcClientProxy`) to use the new simplified API

## Changes

| File | Change |
|------|--------|
| `client/pom.xml` | Add `<optional>true</optional>` to prometheus dependency |
| `MetricsMonitor.java` | Rewrite with runtime prometheus detection and no-op fallback |
| `PrometheusMetricsHelper.java` | New file: isolates all prometheus direct API calls |
| `MetricsHttpAgent.java` | Use `MetricsTimer` instead of direct `Histogram.Child` |
| `ClientWorker.java` | Use `recordListenConfigCount()` |
| `ServiceInfoHolder.java` | Use `recordServiceInfoMapSize()` |
| `NamingGrpcClientProxy.java` | Use `recordNamingRequestFailed()` |

## Backward Compatibility

- Users who already include prometheus in their classpath see no behavior change
- Users without prometheus get no-op monitoring instead of `ClassNotFoundException`

Assisted-by: Hermes Agent

PR #15793, opened 2026-08-31
[ISSUE #8354] Remove Prometheus dependency from nacos-client via Micrometer

### LINKED ISSUE #8354: make nacos-client sdk remove prometheus dependency and metrics with other way.
**Is your feature request related to a problem? Please describe.**

提一个实际应用中碰到的一个小问题，nacos-client依赖了一个较低的prometheus版本，而且非可选，作为核心客户端sdk做到尽可能轻量级的依赖，或者一个Default实现，JMX等，如果需要prometheus，则由adapter或者另一个依赖完成，对sdk的依赖规范来讲，要多一次依赖去除和更新版本的引入，无法精简，期待一个精简的nacos-SDK。

**Additional context**

neoLsH:

Fixes #8354: make `nacos-client` lightweight by removing the mandatory `io.prometheus:simpleclient` dependency, per the plugin direction discussed in the issue. Supersedes #15212, which has been stalled since 2026-05-28 with failing CI and now conflicts with `develop` (maintainer offered reassignment on 2026-06-26).

## Brief changelog

- **Remove** `io.prometheus:simpleclient` from `client/pom.xml`.
- **Add** `NacosClientMetricsProvider` SPI (gauge / request-observation / failed-request-counter) and a `NoopClientMetricsProvider` default. Implementations are discovered via `NacosServiceLoader` (`META-INF/services`); without an adapter artifact on the classpath, recording is a no-op.
- **Refactor** `MetricsMonitor` into a thin static facade with semantic methods (`recordServiceInfoMapSize`, `recordListenConfigCount`, `observeConfigRequest`, `observeNamingRequest`, `recordNamingRequestFailed`); no prometheus types remain in the client.
- **Fix along the way**: `MetricsHttpAgent` previously recorded `httpPost`/`httpDelete` under the `GET` method label; it now records the actual method.
- The existing `enableClientMetrics` property continues to gate all recording sites.

Users who want Prometheus metrics can supply an adapter artifact implementing the SPI (the adapter itself can live in a separate module/repository to keep the SDK lean — happy to contribute it as a follow-up).

## Verifying this change

- New `MetricsMonitorTest` (8 cases): SPI loading, delegation of all five semantic methods, no-op provider safety
- Updated `ClientWorkerTest`, `ServiceInfoHolderTest` to the new static API; `MetricsHttpAgentTest` unaffected
- Full `client` module tests: 1216/1216 pass; `api` and `client-basic` unaffected
- `mvn -pl client -am apache-rat:check checkstyle:check spotless:check spotbugs:check` passes
- Integration tests do not consume these client metrics (verified by repo-wide search), so no coverage regression

Fixes #8354

* [x] Make sure there is a Github issue filed for the change
* [x] Format the pull request title like `[ISSUE #123] ...`
* [x] Write a pull request description that is detailed enough
* [x] Write necessary unit-test to verify your logic correction
* [x] Run `mvn spotless:apply` and basic checks (`apache-rat:check checkstyle:check spotbugs:check spotless:check`) pass


KomachiSion (changes requested):
Thanks for working on making `nacos-client` lighter. However, I don't think the current SPI-based implementation is complete enough to merge yet.

The main blocker is that this change removes the built-in Prometheus implementation and falls back to `NoopClientMetricsProvider`, but does not provide a production metrics adapter. Adding `io.prometheus:simpleclient` back to the application classpath will not restore metrics, because `simpleclient` does not implement or register `NacosClientMetricsProvider`. Existing client metrics would therefore silently disappear after upgrading.

I suggest using Micrometer instead of introducing another metrics SPI:

- `nacos-client` already has a non-optional dependency on `micrometer-core`.
- `FailoverReactor` already records client metrics through `Metrics.globalRegistry`.
- Therefore, migrating these metrics to Micrometer adds no new base dependency or footprint.
- Without a concrete registry, the global composite registry safely behaves as a no-op.
- Once the application registers a `PrometheusMeterRegistry`, subsequent Nacos client metrics are collected automatically.

For Prometheus export, users should add and register either:

- `micrometer-registry-prometheus` for the new Prometheus client; or
- `micrometer-registry-prometheus-simpleclient` for the legacy client.

Adding only the underlying `simpleclient` artifact is not sufficient. If compatibility with plain `simpleclient` is required, an official bridge/adapter must be delivered together with this change rather than left as a follow-up.

Please also ensure that metrics failures can never affect client requests. In the current implementation, provider loading and recording exceptions may escape, and calls from `MetricsHttpAgent` are executed in `finally`, so a provider exception can replace a successful result or mask the original HTTP exception.

The revised implementation should include tests for:

1. No registry configured: client behavior remains unaffected.
2. A registry added after meters are created: subsequent values are collected.
3. `PrometheusMeterRegistry.scrape()` contains the expected metric names, tags, units, and buckets.
4. Provider/registry failures do not affect business requests.
5. GET, POST, and DELETE request method labels are recorded correctly.

Special attention is needed when replacing the existing Prometheus `Histogram` with a Micrometer `Timer` or `DistributionSummary`, because metric names, time units, and bucket definitions may change and break existing dashboards.

KomachiSion:
@neoLsH Would you mind provide your dingtalk number, we would like invite you into nacos contributors group.

neoLsH:
> @neoLsH Would you mind provide your dingtalk number, we would like invite you into nacos contributors group.您能否提供您的钉钉号码？我们想邀请您加入 Nacos 贡献者群。

好的，我在钉钉私聊您～

neoLsH:
Thanks for the detailed review — you were right on all three points, and I have reworked the PR to use Micrometer instead of the SPI. The branch is also merged with the latest `develop`, which now includes the AI watch metrics from #15795.

## Micrometer instead of a new SPI

`MetricsMonitor` now records everything on `Metrics.globalRegistry`, so there is no new base dependency and no second metrics system: `micrometer-core` is already a non-optional dependency of `nacos-client` and `FailoverReactor` already publishes through the same registry. The `NacosClientMetricsProvider` / `NoopClientMetricsProvider` pair and its `META-INF/services` file are gone.

Metrics come back automatically once the application registers a concrete registry (`micrometer-registry-prometheus` or `micrometer-registry-prometheus-simpleclient`), and Spring Boot users get this for free because auto-configured registries are added to the global registry by default.

## Failures can never affect client requests

Every recording path in `MetricsMonitor` swallows `Throwable` now, so the calls made from the `finally` block of `MetricsHttpAgent` can neither replace a successful result nor mask the original HTTP exception. Repeated failures are logged once to avoid flooding the client log.

## Compatibility, measured on a real scrape

I dumped `/metrics` with both registries to check the wire format rather than guessing:

| Meter | Before | After |
| --- | --- | --- |
| `nacos_monitor` gauge | `nacos_monitor{module,name}` | identical |
| `nacos_client_naming_request_failed_total` | same name | identical (Micrometer does not append a second `_total`) |
| `nacos_client_ai_watch_events_total` | same name | identical |
| `nacos_client_request` histogram | `_bucket` / `_count` / `_sum` | `nacos_client_request_seconds_{bucket,count,sum}` plus a `_seconds_max` gauge |

So only the request timer is renamed. The Prometheus default bucket boundaries are kept as service level objectives, and `le="+Inf"` is present, so `histogram_quantile()` keeps working.

There is one thing worth pointing out about the units. The previous code called `histogram.observe(System.currentTimeMillis() - start)`, i.e. it fed a millisecond value into a histogram whose buckets are in seconds, so every observation ended up in `+Inf` and the bucket data was never usable. The timer records the elapsed time properly, so `_sum` changes from milliseconds to seconds and the buckets become meaningful. Dashboards that relied on `_sum`/`_count` can be kept working with a recording rule:

```yaml
- record: nacos_client_request_count
  expr: nacos_client_request_seconds_count
- record: nacos_client_request_sum
  expr: nacos_client_request_seconds_sum * 1000
```

I am happy to switch the timer to a `DistributionSummary` in milliseconds instead if you prefer to keep the old metric name at the cost of the non standard unit — it is a change in one method.

## Adapting to #15795

`AgentWatchClientMetrics` needs `inc()` / `dec()` and, more importantly, `get()` on its meters, which a plain Micrometer counter cannot provide without a registry. `MetricsMonitor` therefore keeps the gauge values and the AI watch event counts in process and exposes semantic methods (`incrementAgentWatchIntentCount()`, `getAgentWatchEventCount(event, result)`, ...). Only the ten call sites inside `AgentWatchClientMetrics` changed; `AgentWatchClientMetricsTest` and `AgentWatchManagerTest` pass unmodified. The read back is deliberately limited to the low cardinality AI watch meters, so the five label naming failure counter does not pay for double bookkeeping.

## Tests

Covering the five cases you asked for:

1. no registry configured — recording is a no-op and nothing throws
2. a registry added after the meters exist — subsequent values are collected
3. `PrometheusMeterRegistry.scrape()` — asserts names, tags, bucket boundaries including `+Inf`, and `timer.totalTime(MILLISECONDS) == 12` for the unit
4. a registry that throws — recording calls stay silent and the simulated `finally` path still returns the business result
5. `GET`, `POST` and `DELETE` labels in `MetricsHttpAgentTest`

The method label fix is included as well: `httpPost` and `httpDelete` used to be recorded as `GET`.

Verification: `nacos-client` 1320 tests pass, and `mvn -pl client apache-rat:check checkstyle:check spotless:check spotbugs:check` is clean.

KomachiSion:
Thanks for the update. The code changes look good, and the documentation follow-up does not need to block this PR.

After this PR is merged, please submit a documentation PR to the `next` documentation in [nacos-group/nacos-group.github.io](https://github.com/nacos-group/nacos-group.github.io), covering:

1. How to enable client metrics with `micrometer-registry-prometheus` or `micrometer-registry-prometheus-simpleclient`.
2. How non-Spring applications should register the registry with `Metrics.globalRegistry`.
3. That adding `io.prometheus:simpleclient` alone is no longer sufficient.
4. The metric migration from `nacos_client_request_{bucket,count,sum}` to `nacos_client_request_seconds_{bucket,count,sum}`.
5. The `_sum` unit change from milliseconds to seconds, including PromQL or recording-rule migration examples for existing dashboards and alerts.

Please link the documentation PR back here once it is created.

neoLsH:
The documentation PR is up: nacos-group/nacos-group.github.io#1147.

It adds a "Java SDK Metrics" page to `docs/next` in both English and Chinese, covering the five points from the review:

1. Enabling client metrics with `micrometer-registry-prometheus` or `micrometer-registry-prometheus-simpleclient`, for both Spring Boot and plain Java applications (`Metrics.globalRegistry` registration).
2. Why a bare `io.prometheus:simpleclient` dependency is no longer sufficient, and that the meters silently disappear if that is all the application has.
3. The metric rename from `nacos_client_request_{bucket,count,sum}` to `nacos_client_request_seconds_{bucket,count,sum}`.
4. The `_sum` unit change from milliseconds to seconds, with recording-rule and PromQL migration examples for existing dashboards.
5. A full metric reference (names, tags, bucket boundaries) plus the `enableClientMetrics` switch and troubleshooting entries.

The page also notes the failure-isolation guarantee you asked about in the first review (recording failures can never affect client requests).

Separately: the latest `ci` run failed on `MetricsMonitorTest.testAgentWatchEventCounterIsExportedAndReadableByLabels` — a test-isolation issue rather than a product bug. The agent watch event counts are process-wide statics, so when `AgentWatchManagerTest` records a `listener_callback`/`failed` event earlier in the same JVM, the assertion expecting `0.0` for that label pair fails. I pushed 2993630b4, which captures the failed-result baseline the same way the success baseline was already captured; the test now passes regardless of execution order.

neoLsH:
The last Integration Test run (attempt 2 on 2993630b) failed during Derby schema initialization on the self-hosted runner (`SQLTimeoutException: Login timeout exceeded` → server never became healthy → all IT steps skipped), so it is unrelated to the change, which only touches `MetricsMonitorTest`.

I have pushed b20913b5a, which adds the missing coverage for the `recordAgentWatchEvent` failure path flagged by codecov: the throwing-registry test now also records one ai watch event and asserts the in-process counter still increases while the registry is broken. The gauge registration catch stays uncovered on purpose — gauges are registered once during class initialization, as documented in the test.

The workflows for the new commit are awaiting approval — could you approve them when you have a moment? Thanks.

client/src/main/java/com/alibaba/nacos/client/monitor/MetricsMonitor.java
```diff
@@ -16,72 +16,215 @@
...
+    
+    private static final String MODULE_NAMING = "naming";
+    
+    private static final String MODULE_CONFIG = "config";
+    
+    private static final String MODULE_AI = "ai";
+    
+    private static final String NACOS_MONITOR = "nacos_monitor";
+    
+    private static final String NACOS_CLIENT_REQUEST = "nacos_client_request";
```
nacos-community, line 61: The `nacos_client_request` metric is now recorded as a Micrometer `Timer`, which the Prometheus registry exports as `nacos_client_request_seconds_*` (e.g., `nacos_client_request_seconds_bucket`). The previous Prometheus histogram exported `nacos_client_request_*` without the `_seconds` unit suffix, so existing dashboards and alerts that scrape this metric will break after upgrading. Document this as a breaking change in the release notes and consider whether to provide a meter filter or naming workaround to preserve the legacy name.

client/src/main/java/com/alibaba/nacos/client/monitor/MetricsMonitor.java
```diff
@@ -16,72 +16,215 @@
...
+ *
+ * <p>Metric names and tags keep the historical conventions: gauge {@code nacos_monitor}, timer
+ * {@code nacos_client_request}, counters {@code nacos_client_naming_request_failed_total} and
+ * {@code nacos_client_ai_watch_events_total}.
+ *
+ * <p>Any failure of the metrics system is swallowed, so it can never affect client requests.
  *
  * @author Nacos
  */
 public class MetricsMonitor {
```
nacos-community, line 49: `MetricsMonitor` removes the public Prometheus-typed accessor methods (`getServiceInfoMapSizeMonitor`, `getConfigRequestMonitor`, `getNamingRequestFailedMonitor`, etc.) and makes the constructor private. This is a source and binary compatibility break for any external code that imported this client class. The change is necessary to drop the Prometheus dependency, but it should be clearly called out in the migration guide.

client/src/main/java/com/alibaba/nacos/client/monitor/MetricsMonitor.java
```diff
@@ -16,72 +16,215 @@
...
-    public static Histogram.Child getConfigRequestMonitor(String method, String url, String code) {
-        return NACOS_CLIENT_REQUEST_HISTOGRAM.labels("config", method, url, code);
+    private static void observeRequest(String module, String method, String url, String code,
+        long elapsedMillis) {
+        try {
+            Timer.builder(NACOS_CLIENT_REQUEST).tag(TAG_MODULE, module).tag(TAG_METHOD, method)
+                .tag(TAG_URL, url).tag(TAG_CODE, code)
+                .serviceLevelObjectives(REQUEST_DURATION_BUCKETS).register(Metrics.globalRegistry)
+                .record(elapsedMillis, TimeUnit.MILLISECONDS);
+        } catch (Throwable t) {
```
nacos-community, line 203: Recording paths catch and swallow `Throwable`, including JVM `Error`s such as `OutOfMemoryError`. While this guarantees metrics can never affect client requests, it can also mask serious runtime problems. Consider catching `Exception` instead and letting `Error`s propagate, or document this deliberate resilience choice explicitly.

client/src/main/java/com/alibaba/nacos/client/monitor/MetricsMonitor.java
```diff
@@ -16,72 +16,215 @@
...
+    public static double getAgentWatchEventCount(String event, String result) {
+        return aiWatchEventCount(event, result).get();
     }
     
-    public static Histogram.Child getConfigRequestMonitor(String method, String url, String code) {
-        return NACOS_CLIENT_REQUEST_HISTOGRAM.labels("config", method, url, code);
+    private static void observeRequest(String module, String method, String url, String code,
+        long elapsedMillis) {
+        try {
+            Timer.builder(NACOS_CLIENT_REQUEST).tag(TAG_MODULE, module).tag(TAG_METHOD, method)
```
nacos-community, line 199: Every call to `observeRequest` allocates a new `Timer.Builder` and performs a `MeterRegistry` lookup. Because this is on the request hot path, consider caching frequently-used timers (and similarly counters) by their tag tuple if profiling shows allocation pressure.

nacos-community (commented):
This PR removes the mandatory `io.prometheus:simpleclient` dependency from `nacos-client` by re-implementing `MetricsMonitor` as a thin Micrometer facade, so metrics recording is a no-op when no `MeterRegistry` is on the classpath. The refactor is correct, thread-safe, well-tested (no-registry / late-registry / Prometheus-scrape / failure cases) and preserves Java 8 compatibility for the client module. Two compatibility notes (anchored inline) are worth capturing in the migration guide / release notes before merge; neither blocks the code change.

### Suggestions
- Record the Prometheus metric-name change `nacos_client_request` -> `nacos_client_request_seconds_*` (Micrometer appends the `_seconds` unit suffix to timers) in the release notes, since existing dashboards/alerts scraping the old name will break.
- Call out the `MetricsMonitor` public API break (removed Prometheus-typed accessors, private constructor) in the migration guide for anyone extending the client.

---

neoLsH:
The `ci` run on b20913b5a failed on **spotless format violations** (the ai watch counter assertions added in that commit exceeded the line length limit) — not a test failure; the Integration Test suite itself passed this time.

I have pushed 14d2d079c, which only reformats those lines via `mvn spotless:apply` — no logic change. Verified locally: `mvn -pl client spotless:check` passes and `MetricsMonitorTest` stays green (6/6).

The new run is awaiting workflow approval again — thanks in advance.

neoLsH:
Thanks for the review. Replying to the inline notes:

**Metric name change (`nacos_client_request` → `nacos_client_request_seconds_*`)** — covered in the documentation PR opened alongside this change (nacos-group/nacos-group.github.io#1147): section 4 "Migrating from pre-3.3" spells out the rename and ships PromQL / recording-rule migration examples. A `MeterFilter` to restore the legacy name was considered but intentionally not documented — the direction from the review of this PR is to accept the corrected name and migrate dashboards, which also matches the unit fix: the old series compared millisecond observations against second-based bucket boundaries. A release-notes entry sounds right; leaving that to maintainers at merge time.

**`MetricsMonitor` API break** — intentional and reviewed: the handle-based accessors returning `ClientGauge` were the direction requested during review of this PR. The doc PR targets metric users (scrape / queries) and deliberately does not cover internal Java API migration; happy to extend it if maintainers want a developer-facing migration note.

**`catch (Throwable)` on recording paths** — this is the documented resilience choice (class javadoc: "Any failure of the metrics system is swallowed, so it can never affect client requests"), and it is not fully silent: `logMetricsFailure` emits one warn with the full stack trace, then suppresses repeats to avoid log storms. I would rather keep `Throwable` than switch to `Exception`: inside the catch there is no meaningful recovery to attempt for an `Error` either, and letting it propagate would reintroduce exactly the coupling (metrics breaking requests) this design removes. Happy to change if maintainers prefer `Exception`.

**Per-call `Timer.builder` + registry lookup** — deferring until profiling shows a need: Micrometer caches meters by id inside the registry (repeated `register` returns the existing instance), so the per-call cost is builder allocation plus the id lookup, and client request volume is far below where this is measurable. If it ever shows up in a profile, caching timers by tag tuple would be a contained follow-up.

nacos-community (commented):
Follow-up on the new commit: I compared `(previous head)` with `14d2d079c` — the incremental change is limited to `client/src/test/java/com/alibaba/nacos/client/monitor/MetricsMonitorTest.java` (+6/−3) and consists solely of line wrapping applied by `mvn spotless:apply` (no logic change), matching the description above. The CI failure on the previous head was a spotless formatting violation, not a test failure.

My earlier review comments on this PR remain applicable; no new findings from this formatting commit.

---

client/src/main/java/com/alibaba/nacos/client/monitor/MetricsMonitor.java
neoLsH: This one is intentional — the rename comes together with the unit fix. The old series was comparing millisecond observations against second-based bucket boundaries, so keeping the legacy name would preserve a broken metric. Migration is covered in the doc PR opened alongside this change (nacos-group/nacos-group.github.io#1147), with PromQL and recording-rule examples. A release-notes entry definitely makes sense at merge time.

client/src/main/java/com/alibaba/nacos/client/monitor/MetricsMonitor.java
neoLsH: Deferred until profiling shows a need — the registry already caches meters by id, so a repeated register returns the existing instance and the per-call cost is just the builder allocation plus the lookup. Client request volume is far below where that becomes measurable. If it ever shows up in a profile, caching timers by tag tuple would be a small follow-up.

client/src/main/java/com/alibaba/nacos/client/monitor/MetricsMonitor.java
neoLsH: This is the documented resilience choice — see the class javadoc: any metrics failure is swallowed so it can never affect client requests. It's not fully silent though: the first failure logs a warn with the full stack trace, and repeats are suppressed to avoid log storms. Inside this catch there's nothing meaningful we could do for an Error either, and letting it propagate would bring back exactly the coupling this design removes. Narrowing to Exception is fine with me if maintainers prefer that.

client/src/main/java/com/alibaba/nacos/client/monitor/MetricsMonitor.java
neoLsH: Intentional as well — the handle-based ClientGauge accessors are the direction requested during the review of this PR; the old one-method-per-operation API was exactly what this refactor removed. The doc PR targets metric users (scrape / queries) rather than Java API migration, but I'm happy to add a short developer-facing note if maintainers want one.

nacos-community (commented):
Thanks for the detailed responses, @neoLsH — I verified against `14d2d079c` and accept all four explanations:

1. **Metric rename (`MetricsMonitor.java:61`)** — Understood: the rename is tied to the unit fix (the old series compared millisecond observations against second-based buckets), and migration is covered by the doc PR nacos-group/nacos-group.github.io#1147. Agreed that a release-notes entry at merge time makes sense.
2. **Timer lookup cost (`MetricsMonitor.java:199`)** — Deferral accepted. The registry returns the cached meter instance on repeated registration, so the per-call cost is bounded; revisiting via a tag-tuple cache can wait until profiling shows pressure.
3. **`catch (Throwable)` (`MetricsMonitor.java:203`)** — Confirmed this is the documented resilience choice (class javadoc, line 45), with the first failure logged as a warn including the stack trace and repeats suppressed via the `METRICS_FAILURE_LOGGED` guard. Whether to narrow to `Exception` is left to the maintainers.
4. **Removed Prometheus-typed accessors (`MetricsMonitor.java:49`)** — Acknowledged: the handle-based `ClientGauge` accessors are the agreed direction of this refactor. A short developer-facing note is at the maintainers' discretion.

No further objections from my side on these points. Remaining decision items (narrowing the catch to `Exception`, developer-facing migration note, release-notes entry) are deferred to the maintainers.

---

KomachiSion (approved):

PR #15798, opened 2026-09-01
[ISSUE #14107] Support batch clone and export for MCP servers

1lck:
Related issue: #14107

This PR adds batch export and batch clone support for MCP servers.

Highlights:
- Add batch export API and UI support.
- Add batch clone API and UI support.
- Support DIRECT, REF, and STDIO endpoints.
- Support ABORT, SKIP, and OVERWRITE conflict policies.
- Preserve the current version content, endpoint configuration, labels, and visibility.
- Clone through the MCP AI Resource lifecycle model without using the legacy import flow.
- Add backend, maintainer-client, and frontend tests.

Validation:
- MCP-related backend tests passed.
- Frontend type checking, linting, and MCP API tests passed.
- Full frontend test suite still contains unrelated baseline failures.


nacos-community (changes requested):
### Summary
This PR adds batch export and batch clone support for MCP servers (issue #14107): new v3 console APIs with ABORT/SKIP/OVERWRITE conflict policies, AI/maintainer-client plumbing for DIRECT/REF/STDIO endpoints, plus backend and frontend tests. The structure and test coverage are good and endpoints are properly authorized, but the OVERWRITE path has a data-loss correctness issue and the REF namespace routing has a behavioral regression, so requesting changes.



### Suggestions
- Reorder OVERWRITE so the replacement draft is created and published before deleting the old server (or keep a fallback copy), and explicitly reject/guard source == target self-overwrite.
- Fall back to the caller namespaceId when REF endpoint data has a blank namespaceId, and add validation/test coverage for that case.
- Add catch blocks in both batch handlers to surface errors to the user.

---

nacos-community:
Detailed findings for the review above (CHANGES_REQUESTED):

### Findings
- **[CRITICAL]** `console/src/main/java/com/alibaba/nacos/console/controller/v3/ai/ConsoleMcpController.java:536` — The OVERWRITE path deletes the target MCP server before the replacement draft is created and published. If createMcpServerDraft or forcePublishMcpServerVersion fails after the delete, the target is permanently lost. It also breaks self-overwrite (sourceName == targetName in the same namespace) because the source is deleted before requireSourceServer can fetch it again. Reorder the operation so the replacement is successfully created first, or keep a fallback copy, and explicitly guard against overwriting a source with itself.
- **[WARNING]** `ai/src/main/java/com/alibaba/nacos/ai/service/McpEndpointOperationService.java:107` — Switching from the caller namespace to the namespaceId stored in the REF endpoint data is correct when the data is present, but the preceding validation only checks groupName/serviceName, not namespaceId. Existing REF endpoints that omit namespaceId in their data will now be routed to a null/default namespace instead of the caller namespace, which is a behavioral regression. Fall back to the caller namespaceId when the endpoint data value is blank.
- **[WARNING]** `console-ui-next/src/pages/mcpServerManagement/index.tsx:129` — handleBatchExport only has a finally block; network or server errors thrown by mcpApi.exportServers are not caught, so users see no error feedback and the unhandled rejection may surface in the UI framework's error boundary. Add a catch block that toasts the error before resetting the loading state.
- **[WARNING]** `console-ui-next/src/pages/mcpServerManagement/index.tsx:150` — handleBatchClone also lacks a catch block for non-2xx failures or transport errors from mcpApi.cloneServers. Only the in-band failedCount/skippedCount returned by a successful HTTP response is handled. Add a catch block to notify the user and stop loading when the request itself fails.
- **[INFO]** `console/src/main/java/com/alibaba/nacos/console/controller/v3/ai/ConsoleMcpController.java:483` — Binding the policy parameter directly to the SameConfigPolicy enum relies on Spring's default enum converter, which is case-sensitive and throws MethodArgumentTypeMismatchException for invalid values. For consistency with other Nacos endpoints that return NacosApiException with PARAMETER_VALIDATE_ERROR, consider accepting a String and validating it manually.
- **[INFO]** `console/src/main/java/com/alibaba/nacos/console/controller/v3/ai/ConsoleMcpController.java:490` — The null check `policy == null ? SameConfigPolicy.ABORT : policy` is dead code because the @RequestParam already supplies defaultValue = "ABORT". Removing it would make the intent clearer.

---
*Automated comment by github-manager-bot*

PR #15842, opened 2026-09-12
feat(naming): add DNS server support for service discovery

### LINKED ISSUE #50: DNS protocol support for naming service
In order to integrate Kubernetes and support standard DNS resolution for registered services, We need build a DNS server in Nacos. Proposed features are listed below:
1. Standard A record query.
2. Support forwarding to external DNS servers.
3. Use Nacos as backend.
4. Use Kubernetes as backend and be qualified for Kubernetes DNS baseline.
5. Use Kubernetes labels as optional source for Load-Balance.

ABin-Huang:
## Description

This PR implements a basic DNS server for Nacos naming service, addressing issue #50.

## Implementation

This is an MVP (Minimum Viable Product) implementation that provides:

### Features
1. **Standard A record query** - Resolve service names to IP addresses
2. **UDP and TCP transport** - Support both DNS transport protocols
3. **Healthy instance filtering** - Only return healthy, enabled instances
4. **Configurable port** - Default 5353 to avoid conflict with system port 53
5. **Domain format**: `{serviceName}.{groupName}.nacos`

### Configuration
```properties
# Enable DNS server (disabled by default)
nacos.naming.dns.enabled=true
# DNS server port
nacos.naming.dns.port=5353
# Domain suffix
nacos.naming.dns.domain-suffix=nacos
# Default group name
nacos.naming.dns.default-group=DEFAULT_GROUP
# DNS TTL in seconds
nacos.naming.dns.ttl=60
```

### Usage
```bash
# Query service using dig
dig @localhost -p 5353 order-service.DEFAULT_GROUP.nacos A

# Query with default group
dig @localhost -p 5353 order-service.nacos A
```

## Future Enhancements

This MVP covers the core requirement of issue #50. The following features can be added in future PRs:

- [ ] Support forwarding to external DNS servers
- [ ] Use Kubernetes as backend
- [ ] Use Kubernetes labels as load balancing source
- [ ] Support SRV record queries
- [ ] Support AAAA record queries (IPv6)
- [ ] Metrics and monitoring

## References

- Fixes #50

## Does this pull request introduce a breaking change?

No, this is an optional feature that is disabled by default.


KomachiSion:
1. Your PR changed not only naming module But config and other module which I don't think this feature should be changed.
2. Dns feature should be a independence module like `address` module which can depend naming module.

ABin-Huang:
Hi @KomachiSion, I have refactored the DNS feature as requested:

1. **Independent module**: Moved all DNS code to a new independent `dns` module (nacos-dns), following the same pattern as the `address` module. It depends on `nacos-naming`.
2. **Clean changes**: The PR now only contains DNS-related code - no changes to config module or any other unrelated modules.

The new module structure:
- `dns/pom.xml`
- `dns/src/main/java/com/alibaba/nacos/dns/` (4 Java classes)
- `dns/src/test/java/com/alibaba/nacos/dns/` (unit tests)
- Root `pom.xml`: added `<module>dns</module>` and dependency management
- `server/pom.xml`: added nacos-dns dependency

All local builds and tests pass. Thanks for the review!

dns/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+    <url>https://nacos.io</url>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>nacos-naming</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>dnsjava</groupId>
+            <artifactId>dnsjava</artifactId>
```
KomachiSion, line 40: which license for this `dnsjava` ?

ABin-Huang:
@KomachiSion Regarding the license question on `dns/pom.xml`:

dnsjava 3.5.2 is licensed under **BSD 3-Clause "New" or "Revised" License**, which is fully compatible with Apache License 2.0 used by Nacos.

Reference: https://mvnrepository.com/artifact/dnsjava/dnsjava/3.5.2

dns/src/main/java/com/alibaba/nacos/dns/NacosDnsServer.java
```diff
@@ -0,0 +1,219 @@
...
+            }
+        } catch (IOException e) {
+            LOGGER.error("Failed to start TCP DNS server on port {}", properties.getPort(), e);
+        }
+    }
+    
+    /**
+     * Handle a TCP DNS connection.
+     */
+    private void handleTcpConnection(Socket clientSocket) throws IOException {
```
KomachiSion, line 341: TCP connections perform blocking reads without a timeout and share the same 10-thread executor with both listener loops and UDP requests. Since the listeners occupy two threads, only eight idle TCP connections are sufficient to block all further DNS query processing.
This was reproduced locally: eight connections sending no data caused an ordinary UDP query to time out. The executor also has an unbounded queue, allowing accepted sockets and requests to accumulate.
Please address this before merging: separate listener execution from request processing, enforce connection limits and a bounded queue, and add idle/read deadlines, including a deadline for receiving a complete frame. Rejected connections must be closed. Please add a regression test verifying that idle or incomplete TCP requests cannot starve UDP queries.

dns/src/main/java/com/alibaba/nacos/dns/NacosDnsServer.java
```diff
@@ -0,0 +1,219 @@
...
+    
+    /**
+     * Start UDP listener.
+     */
+    private void startUdpListener() {
+        try {
+            udpSocket = new DatagramSocket(properties.getPort());
+            LOGGER.info("DNS UDP server listening on port {}", properties.getPort());
+            
+            byte[] buf = new byte[MAX_DNS_PACKET_SIZE];
```
KomachiSion: Every DatagramPacket references the same receive buffer, but processing is asynchronous. A subsequent receive() can overwrite the data before an earlier task parses it, while the destination address still belongs to the earlier request.
This was reproduced locally: client A received client B's transaction ID and question name. Additionally, packet.getLength() is ignored when constructing Message; a two-byte datagram was parsed using residual bytes from the previous request and produced a valid response.
Please copy only the received bytes, respecting the packet offset and length, before submitting the task. Parse that independent copy and reject malformed or incomplete messages. Add regression tests for concurrent requests from different clients and short datagrams following a valid query.

dns/src/main/java/com/alibaba/nacos/dns/NacosDnsServer.java
```diff
@@ -0,0 +1,219 @@
...
+     * Handle a UDP DNS packet.
+     */
+    private void handleUdpPacket(DatagramPacket packet) throws IOException {
+        byte[] data = packet.getData();
+        int length = packet.getLength();
+        
+        Message query = new Message(data);
+        Message response = queryHandler.handleQuery(query);
+        
+        byte[] responseData = response.toWire();
```
KomachiSion: The handler includes every matching instance in the answer and sends response.toWire() without a UDP response-size limit. MAX_DNS_PACKET_SIZE only limits the receive buffer; it does not constrain the response.
With 100 instances, a local test produced a 1,635-byte response to a 35-byte query, approximately 46.7× amplification at the DNS payload level, with TC=false. If the endpoint is reachable and an attacker can spoof UDP source addresses, it can be used for reflection amplification.
Please enforce the applicable UDP response limit—512 bytes without EDNS, or the advertised EDNS size capped by a server limit—and set TC when truncation is required. Add response rate limiting and document the intended network exposure. Tests should cover large instance sets, truncation, and successful TCP retrieval.

dns/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+            <version>3.5.2</version>
```
KomachiSion: Please replace dnsjava:3.5.2 with a maintained, patched release before merging. Version 3.5.2 falls within the affected ranges of GHSA-cfxw-4h78-h7fw, GHSA-crjg-w57m-rqqf, and GHSA-mmwx-rj87-vfgr; the corresponding fixes were released in 3.6.0.
The current implementation does not appear to invoke the affected lookup/DNSSEC validation paths, so this is not a claim that those vulnerabilities are directly exploitable through this PR. However, a new dependency should start from a patched baseline. Please verify compatibility and run the DNS tests against the selected version.

KomachiSion (changes requested):

dns/src/main/java/com/alibaba/nacos/dns/NacosDnsServer.java
```diff
@@ -0,0 +1,343 @@
...
+            if (running) {
+                LOGGER.error("Failed to start TCP DNS server on port {}", properties.getPort(), e);
+            }
+        }
+    }
+
+    /**
+     * Handle a TCP DNS connection.
+     */
+    private void handleTcpConnection(Socket clientSocket) throws IOException {
```
KomachiSion, line 341: The connection limit, bounded queue, and read timeout improve the original implementation, but slow TCP requests can still exhaust the shared worker pool.
I reproduced this with 16 connections sending one byte per second for an incomplete DNS frame. After more than 6.5 seconds, all 16 workers remained occupied and a normal UDP query timed out. SO_TIMEOUT limits an individual blocking read; it does not impose a deadline for receiving the complete message.
A full non-blocking IO rewrite is not required for this PR. Please implement these two changes before merging:
1. Use separate bounded worker pools for TCP connections and UDP queries, so waiting for TCP data cannot consume all UDP processing capacity.
2. Enforce an absolute receive deadline covering both the two-byte length prefix and the complete message body. Each read must use the remaining time budget, and the socket must be closed when that budget expires. Receiving partial data must not reset the deadline.
Please retain the existing connection/queue limits and socket cleanup. Add regression tests proving that a slowly transmitted incomplete frame is closed within the configured deadline and that UDP queries remain responsive while TCP workers are occupied.
This provides a practical baseline for this PR; more comprehensive connection fairness and non-blocking IO can follow separately.
Reference: [RFC 7766 §6.2.3](https://www.rfc-editor.org/rfc/rfc7766.html#section-6.2.3) recommends resetting the idle timeout only after receiving a complete DNS message.

dns/src/main/java/com/alibaba/nacos/dns/NacosDnsQueryHandler.java
```diff
@@ -0,0 +1,208 @@
...
+        // Parse domain and look up service
+        List<InetAddress> addresses = resolveToAddresses(domain);
+
+        if (addresses == null || addresses.isEmpty()) {
+            response.getHeader().setRcode(Rcode.NXDOMAIN);
+            return response;
+        }
+
+        // Shuffle to provide basic round-robin, then limit to MAX_ANSWER_RECORDS
+        Collections.shuffle(addresses);
```
KomachiSion: The new answer limit is applied before filtering addresses by the requested record type. With mixed IPv4/IPv6 instances, the first 20 randomly selected addresses may contain no addresses matching the query, even though healthy matching instances exist.
With 100 IPv4 instances and one IPv6 instance, 81 out of 100 AAAA queries returned an empty answer in the updated implementation. The original implementation returned the IPv6 address in all 100 queries. A queries can fail in the same way when IPv4 instances are the minority.
Please filter addresses by the requested type first, then shuffle and apply any answer-count limit. An address from the other family must not consume the current query's record quota.
Please add regression tests for both cases: many IPv4 instances with one IPv6 instance queried via AAAA, and many IPv6 instances with one IPv4 instance queried via A. Each query should return the existing healthy address of the requested family.

KomachiSion (changes requested):
Thanks for your PR an update, There are two more review comment, please change or discussion more.

ABin-Huang:
@KomachiSion Thanks for the thorough review! All feedback has been addressed:

1. **dnsjava** upgraded to 3.6.0 (patched baseline, BSD 3-Clause).
2. **UDP buffer race** — each datagram is defensively copied before async processing.
3. **UDP amplification** — responses capped at 512 bytes with TC bit set; fallback minimal response if still over limit.
4. **TCP resource exhaustion** — listener threads separated from worker pool; 64 connection limit; bounded queues.
5. **Slow TCP frame / worker pool starvation** — separate UDP (16 threads) and TCP (8 threads) worker pools; absolute receive deadline (5s) covering both length prefix and message body, reset only after a complete message per RFC 7766 §6.2.3.
6. **Type filter ordering** — addresses are now filtered by A/AAAA type before shuffle and limit, so mixed IPv4/IPv6 deployments return matching records.

Regression tests added for: mixed IPv4/IPv6 queries, slow TCP frame closed within deadline, UDP responsiveness while TCP workers are occupied, large instance sets, truncation, and TCP retrieval. All 21 tests pass locally.

KomachiSion (approved):

KomachiSion:
Thanks for contribution, I'm not sure whether you have dingtalk. 

If you have, would you mind provide dingtalk number? we would like to invite you into nacos contributors groups.

ABin-Huang:
> Thanks for contribution, I'm not sure whether you have dingtalk.
> 
> If you have, would you mind provide dingtalk number? we would like to invite you into nacos contributors groups.

Thank you! My DingTalk ID:abin2ddh. Glad to join the contributors group.

PR #15862, opened 2026-09-17
[ISSUE #15345] Return cached frontmatter in Skill list responses

Zhengcy05:

Addresses the frontmatter response support requested in [[#15345](https://github.com/alibaba/nacos/issues/15345)](https://github.com/alibaba/nacos/issues/15345), allowing callers to display Skill aliases and custom metadata.

This follows up on the original contribution by @jay666mnj in [[#15508](https://github.com/alibaba/nacos/pull/15508)](https://github.com/alibaba/nacos/pull/15508). Following the maintainer's guidance, frontmatter is populated only when Skill content is created or updated, without historical backfill.

## Brief changelog

- Add nullable `frontMatter` to Skill list and metadata detail responses.
- Parse frontmatter when saving Skill content and store a display snapshot in existing resource metadata. List queries require no additional per-Skill version or package reads.
- Keep the snapshot consistent across lifecycle transitions, using version matching and CAS retries to handle concurrent updates.
- Preserve compatibility with historical content that lacks frontmatter metadata. Frontmatter search is outside this change.
- Update English and Chinese specs, unit tests, API/SDK integration scenarios, and coverage documentation.

## Verifying this change

Passed 138 focused unit tests, covering lifecycle transitions, historical data, pagination without extra reads, concurrency conflicts, and JSON compatibility:

```shell
mvn -B -pl api,ai -am "-Dtest=SkillSummaryTest,SkillOperationServiceImplTest,SkillFrontMatterTest,SkillListFilterTest" "-Dsurefire.failIfNoSpecifiedTests=false" -DskipITs test
```

Passed formatting and license checks for affected modules:

```shell
mvn -B -pl api,ai,test/openapi-test,test/maintainer-sdk-test spotless:apply spotless:check apache-rat:check
```

Passed compilation of OpenAPI and Maintainer SDK integration tests, including their dependencies:

```shell
mvn -B -pl test/openapi-test,test/maintainer-sdk-test -am -DskipTests test-compile
```

Checkstyle passed during the builds. Integration tests against a running Nacos server have not been executed. Full-repository validation, including SpotBugs, remains pending.

- [x] Link the related issue and keep changes scoped to it.
- [x] Follow the required PR title format.
- [x] Explain the purpose and implementation.
- [x] Add unit tests and integration test scenarios.
- [ ] Complete all required full-build, unit-test, and integration-test checks.


KomachiSion (changes requested):
Thanks for the update. The overall direction looks good:

- No new DAO, table, migration, or persistence model is introduced.
- The implementation reuses ai_resource.ext and ai_resource_version.storage.
- Historical Skills without cached frontmatter remain compatible and may return
  frontMatter = null.
- Skill list queries do not introduce per-Skill version or package reads, so there
  is no N+1 query problem.

Before merging, please address the following issues.

1. Do not let an optional cache refresh fail an already completed lifecycle operation

refreshFrontMatter is called after publish, force-publish, draft deletion, and
online/offline operations have already updated the version status, resource
metadata, and sometimes the manifest.

If the frontmatter CAS retries are exhausted, the API currently returns a conflict
even though the primary operation has already succeeded. This is especially
problematic for force-publish because the version is already online and the same
operation cannot be retried.

Since frontmatter is optional display metadata and a missing snapshot is already
represented as null, cache refresh failure should not override the result of the
primary lifecycle operation. Please make the cache refresh best-effort and log the
failure, or update it atomically as part of the primary metadata transition.

2. Store a bounded custom-field snapshot for list responses

Please keep the complete SKILL.md and the complete version-level frontmatter
metadata unchanged. Limits should apply only to the ai_resource.ext snapshot used
by list/detail responses.

The snapshot does not need to duplicate name, description, or version because
these values are already available while assembling SkillSummary. When
frontMatterVersion matches the resolved display version, frontMatter can be
assembled as follows:

- frontMatter.name = SkillSummary.name
- frontMatter.description = SkillSummary.description
- frontMatter.version = resolved display version
- merge the bounded cached custom fields afterwards

Reserved fields such as name, description, and version should be removed from the
cached custom map so they cannot override the generated values.

This assembly only performs a few in-memory Map operations and introduces no
additional DAO queries. It also guarantees that SkillSummary.description and
frontMatter.description never contain conflicting values.

For the cached custom fields, a suggested limit is:

- at most 64 entries;
- at most 128 UTF-8 bytes per key;
- at most 1024 characters per cached value, appending "..." when truncated;
- at most 16 KiB for the serialized custom-field snapshot.

Standard extension fields such as alias, license, compatibility, allowed-tools,
and metadata.* should be retained before other custom fields. Once the total
budget is exhausted, remaining lower-priority fields may be omitted. The
truncation behavior should be documented; exposing a frontMatterTruncated flag
would also make partial projections explicit.

Please keep the existing pageSize behavior for compatibility. Do not silently cap
or reduce the requested page size. Callers can reduce pageSize themselves when
bandwidth or timeout becomes a concern.

Historical Skills without a matching frontMatterVersion marker should continue
returning frontMatter = null rather than generating a new snapshot during list
queries.

3. Handle malformed historical metadata defensively

parseMetadata currently propagates JSON deserialization failures. A single
malformed historical ai_resource.ext value could therefore fail an entire list or
detail request.

Please follow the defensive behavior already used by parseStorageFiles and
parseStorageProvider: treat malformed metadata as unavailable, return an empty
map/null frontmatter, and avoid failing the whole request. Please also add a
regression test for this case.

At the reviewed head, the focused 138 unit tests pass. However, the PR still
conflicts with develop, and spotless:check fails for SkillSummary.java. Please
resolve the conflicts and run Spotless before the next review.

KomachiSion (approved):

Travis696 (approved):

Issue #15868, opened 2026-09-20
Does SnapshotSwitch support disabling via environment variables?


Kurok1:
**Is your feature request related to a problem? Please describe.**

I recently discovered using Codex that its AI capabilities have evolved to the point where it can autonomously read local Nacos snapshot caches to troubleshoot issues. This could lead to the risk of leaks in many critical configurations.

**Describe the solution you'd like**

I found that SnapshotSwitch can be manually turned off, but almost every application requires writing code to turn it off. I was wondering if it would be possible to enable/disable it via a global environment variable?

KomachiSion:
Which snapshot you want to close？

Kurok1:
A snapshot of the local configuration file. Nacos pulls the remote configuration file to the local machine to create a snapshot when the application starts.

Although I see there's a function to configure encryption, I feel that without restrictions, with current AI capabilities, it wouldn't be difficult to reproduce the decryption process on a user's machine and obtain the plaintext data.

I looked at the relevant source code, and currently it seems that snapshot production is disabled via the client-side using `com.alibaba.nacos.client.config.utils.SnapShotSwitch#setIsSnapShot`, but it doesn't seem to provide a configuration option or environment variable to disable it.

hutiefang76:
I opened #15871 for this. It adds the client property `nacos.config.snapshot.enabled` (and therefore supports `NACOS_CONFIG_SNAPSHOT_ENABLED`) with the default preserved as `true`. Setting it to `false` at startup disables local configuration snapshots and clears existing snapshots.

PR #15871, opened 2026-09-21
[Enhancement] Support disabling config snapshots by property

hutiefang76:
## What this changes

Adds the `nacos.config.snapshot.enabled` client property, defaulting to `true`.
It can be supplied as a JVM property or as the mapped environment variable
`NACOS_CONFIG_SNAPSHOT_ENABLED`. When disabled, the client does not read or
write local configuration snapshots and clears snapshots already present at startup.

## Why

Applications currently need to invoke `SnapShotSwitch#setIsSnapShot(false)` in
application code. This makes the same process-wide setting available at startup
for deployments that manage configuration through JVM options or environment
variables.

Related to #15868.

## Validation

- `mvn -q -pl client -am spotless:check`
- `mvn -q -pl client -am -DskipITs -Dtest=SnapShotSwitchTest -Dsurefire.failIfNoSpecifiedTests=false test`


api/src/main/java/com/alibaba/nacos/api/PropertyKeyConst.java
```diff
@@ -87,6 +87,11 @@ public class PropertyKeyConst {
     
     public static final String ENABLE_REMOTE_SYNC_CONFIG = "enableRemoteSyncConfig";
```
Travis696: Nice

api/src/main/java/com/alibaba/nacos/api/PropertyKeyConst.java
```diff
@@ -87,6 +87,11 @@ public class PropertyKeyConst {
...
     
+    /**
+     * Whether the client stores remote configuration snapshots on the local file system.
+     */
+    public static final String CONFIG_SNAPSHOT_ENABLED = "nacos.config.snapshot.enabled";
```
KomachiSion, line 93: I see others key are camel type.

KomachiSion (changes requested):

hutiefang76:
Addressed the property-naming feedback in `98cddf7c2`: the public setting now uses `configSnapshotEnabled`, with the regression test aligned to that name. The current CLA and commit-author checks pass. Ready for re-review when convenient.

